PR #40, opened 2019-05-31
Remove unused dependencies in prep for nitro-web integration

garettarrowood:
To verify that these libraries were safe to remove I got `playbook` setup on my system and have launched it locally. All styles look the same when compared to production playbook.

This effort is made in part to not add as many extra dependencies into `nitro-web`, as can be seen in https://github.com/powerhome/nitro-web/pull/11445/commits/5e5735a63981d46b8a04a8d4088e6e2d9eda2db0 .

Changes:

1. There are no plans to add tests to this repo. Thus `rspec-rails`, `simplecov`and rspec config files have been removed.
1. There is no plan to adhere to any Ruby linting conventions. Thus `rubocop` is removed.
1. Sign in/Sign up functionality from the repo has been removed for awhile, and there is no pursuit to add auth into `playbook`. Thus `devise` has been removed, along with a helper `devise.yml` file.
1. The rubygems bug in `rainbow` version 2.2.0 has long been resolved in subsequent patch releases. Thus `rainbow` can be safely raised to `2.2.2` and the code comment removed.
1. `sassc-rails` does a better job of compiling scss. Both `sass-rails` and `sassc-rails` were being installed and required. Swapping the `sass-rails` requirement to `sassc-rails` was all that was needed to be done to remove the `sass-rails` dependency.
1. Locked `webpacker` version.
1. With versions of yarn at or above `1.10.x`, integrity hashes have been added to `yarn.locks`. By default, `yarn install` lazily checks if a qualifying version of a dependency is available. But when passed the `--check-files` flag, it adheres to checking for the version specified in the integrity hash. To debug one of my issues getting playbook to load, I had to run `docker-compose run web bin/yarn install --check-files` followed by `docker-compose run webpacker rm -rf /app/public/packs; ./bin/webpack-dev-server`. Hopefully specifying `--check-files` in the install command in the Dockerfile helps avoid webpacker issues in the future (in addition to locking the webpacker version).
1. Migration files do not need to be included in gem build, so the `db` pattern is removed from the gemspec.


Dockerfile
```diff
@@ -24,7 +24,7 @@ ADD Gemfile* *.gemspec /home/app/src/
 RUN bundle install --frozen
 
 ADD package.json yarn.lock /home/app/src/
-RUN yarn install
+RUN yarn install --check-files
```
benlangfeld: This won't have any impact on running `yarn install` outside of the image build, since they are two separate installations. This bakes dependencies into the image, while executing via `docker-compose run` uses a volume.

garettarrowood: Ok - cool. Thank you! I will remove it.

PR #64, opened 2019-07-09
Add rubocop to repo with watered down Nitro Ruby linting conventions

garettarrowood:
This only adds `rubocop` as a development dependency and includes a sample configuration file. It is not tied into your CI pipeline. And running `rubocop` does not pass.

As mentioned in our meeting, the `Metrics` department of linting cops is disabled in the configuration file.

Running `rubocop` from the root of `playbook` generates 639 offenses. The large amount of offenses is mostly due to the `double_quote` Nitro style choice, and that `rubocop` likes all Ruby files to start with the magic string literal comment to prep the code for compatibility with Ruby 3.0.

```
37 files inspected, 639 offenses detected
```

Running `rubocop --autocorrect` would conform the codebase to follow the styles dictated in this config file.


jasoncypret:
@garettarrowood what would be invloved to get this part of our CI?...or is that a Ben L. question?

garettarrowood:
@jasoncypret - I think the next step would be you and I & anyone else you think should be part of the convo pairing for 30-60 mins to work this down to a passing run. It will be a combo of 1. autocorrecting offenses, 2. deciding to disable cops & 3. creating a todo file to make other fixes later.

I should be able to get it in your CI. But an alternative is to add `overcommit` to the repo and have it run there. `Nitro` uses overcommit, but it runs slowly because of its Bundle security checks. If overcommit only runs `rubocop` and everyone commits normally, that would be another way to loop it into you workflow (it's very likely ESLint could be added into overcommit as well).

jasoncypret:
@markiearnold Can you sync with @garettarrowood next week? I'd like to get it working using `overcommit` and autocorrect. I trust your judgment over what we should start with. If it like a weird cop lets skip it.

PR #72, opened 2019-07-18
Kits/stat

markiearnold:
**Kit**: Dashboard Value
**PR contains**: Rails version

## What is it 
Dashboard value is a reusable rails ui kit to be used to show stats within Nitro.

*IMPORTANT*: Dashboard Value is different than our more basic kits because we want to utilize existing kits to build it. Please look at my approach on how I did this in the PbDashboardValue and it’s child kits - such as PbDashboardValue::StatLabel classes and provide suggestions and feedback. Thank you!

## Usage
Dashboard Value is made up of the `stat_label`, `stat_value` and `stat_change` child kits. 
Dashboard Value kit is made so that you can use it as a single kit, or the child kits can be used individually so we can use them within other kits.

## Examples

---

### DIFF
--- a/app/pb_kits/playbook/pb_dashboard_value/dashboard_value.rb
+++ b/app/pb_kits/playbook/pb_dashboard_value/dashboard_value.rb
@@ -1,21 +1,21 @@
 module Playbook
   module PbDashboardValue
     class DashboardValue < Playbook::PbKit::Base
-      PROPS = [:configured_align,
-        :configured_classname,
-        :configured_data,
-        :configured_id,
-        :configured_stat_change,
-        :configured_stat_label,
-        :configured_stat_value].freeze
+      PROPS = %i[configured_align
+                 configured_classname
+                 configured_data
+                 configured_id
+                 configured_stat_change
+                 configured_stat_label
+                 configured_stat_value].freeze
 
       def initialize(align: default_configuration,
-        classname: default_configuration,
-        data: default_configuration,
-        id: default_configuration,
-        stat_change: default_configuration,
-        stat_label: default_configuration,
-        stat_value: default_configuration)
+                     classname: default_configuration,
+                     data: default_configuration,
+                     id: default_configuration,
+                     stat_change: default_configuration,
+                     stat_label: default_configuration,
+                     stat_value: default_configuration)
 
         self.configured_align = align
         self.configured_classname = classname
@@ -27,8 +27,8 @@ module Playbook
       end
 
       def align
-        align_options = %w(left center right)
-        one_of_value(configured_align, align_options, "left")
+        align_options = %w[left center right]
+        one_of_value(configured_align, align_options, 'left')
       end
 
       def stat_label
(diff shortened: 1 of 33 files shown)


app/pb_kits/playbook/pb_dashboard_value/dashboard_value.rb
```diff
@@ -0,0 +1,76 @@
+module Playbook
+  module PbDashboardValue
+    class DashboardValue < Playbook::PbKit::Base
+      PROPS = [:configured_align,
+          :configured_classname,
+					:configured_data,
+					:configured_id,
+          :configured_stat_change,
```
drborges: It seems you got a few tab characters in this file.

app/pb_kits/playbook/pb_dashboard_value/stat_label.rb
```diff
@@ -0,0 +1,53 @@
...
+          end
+          ApplicationController.renderer.render(partial: pb_label, as: :object)
+        end
+      end
+
+      def kit_class
+        stat_options = [
+          "pb_stat_label"
+        ]
+        stat_options.reject(&:nil?).join("_")
```
drborges: It seems this method could be reduced to simply:

```rb
def kit_class
  "pb_stat_label"
end
```

app/pb_kits/playbook/pb_dashboard_value/stat_value.rb
```diff
@@ -0,0 +1,63 @@
...
+          pb_unit = Playbook::PbTitle::Title.new({size: 3, text: self.unit})
+          ApplicationController.renderer.render(partial: pb_unit, as: :object)
+        end
+      end
+
+      def kit_class
+        stat_options = [
+          "pb_stat_value"
+        ]
+        stat_options.reject(&:nil?).join("_")
```
drborges: It looks like this entire method could be reduced to simply:

```rb
def kit_class
  "pb_stat_value"
end
```

app/pb_kits/playbook/pb_dashboard_value/dashboard_value.rb
```diff
@@ -0,0 +1,76 @@
...
+          ApplicationController.renderer.render(partial: pb_change, as: :object)
+        end
+      end
+
+      def kit_class
+        kit_options = [
+          "pb_dashboard_value",
+          align
+        ]
+        kit_options.reject(&:nil?).join("_")
```
drborges: It seems that `align` always return a value (since on its implementation `one_of_value` has a default return value), so `reject(&:nil?)` would be unnecessary?

app/pb_kits/playbook/pb_dashboard_value/stat_change.rb
```diff
@@ -0,0 +1,92 @@
...
+        end
+        ApplicationController.renderer.render(partial: pb_icon_element, as: :object)
+      end
+
+      def kit_class
+        stat_options = [
+          "pb_stat_change",
+          status
+        ]
+        stat_options.reject(&:nil?).join("_")
```
drborges: It seems that `status` always returns a non-nil value, so `reject(&:nil?)` would be unnecessary?

PR #73, opened 2019-07-18
Demo/js exports

jasperfurniss:
✅  Isolated All Styles from Business Intelligence
✅  Implemented our SCSS variables in all our BI JS files
✅  Added the BarGraph kit 

🎉 Woohoo!


app/pb_kits/playbook/pb_bar_graph/bar_graph.rb
```diff
@@ -0,0 +1,109 @@
...
+        self.configured_orientation = orientation
+      end
+
+      def orientation 
+        orientation_options = %w(vertical horizontal)
+        one_of_value(configured_orientation, orientation_options, "vertical")
+      end
+
+      def chart_type
+        case self.orientation
```
terryfinn: This can be simplified to:

```ruby
if orientation == "horizontal"
  "bar"
else
  "column"
end
```

app/pb_kits/playbook/pb_bar_graph/_bar_graph.html.erb
```diff
@@ -0,0 +1,13 @@
...
+    new PBChart('.selector', {
+      data: <%= object.chart_data %>,
+      type: "<%= object.chart_type %>",
+      title: "<%= object.title %>",
+      subtitle: "<%= object.subtitle %>",
+      axisTitle: "<%= object.axis_title %>",
+      pointStart: <%= object.point_start %>,
+    });
+  <% end %>
+<% end %>
```
terryfinn: This file is missing the final blank line.

app/pb_kits/playbook/pb_charts/business_intelligence/pbHighcharts.js
```diff
@@ -0,0 +1,84 @@
...
+
+const styleLegend = function(highchart) {
+  highchart.legend.itemStyle.fontFamily = typography.font_family_base;
+  highchart.legend.itemStyle.color = colors.text_lt_light;
+  highchart.legend.itemStyle.fontWeight = typography.regular;
+  highchart.legend.itemStyle.fontSize = typography.text_smaller;
+};
+
+// Exportable Global Styles Function
+export const pbHighcharts = function(highchart) {
```
indiebrain: Could you explain the decision to export `pbHighcharts` as a `named` export here in addition to exporting it via the `default` export on line 84?

https://github.com/powerhome/playbook/blob/8543f5c5aedb70564592f6f0bebfc3049e73e388/app/pb_kits/playbook/pb_charts/business_intelligence/pbHighcharts.js#L84

app/pb_kits/playbook/pb_bar_graph/bar_graph.rb
```diff
@@ -0,0 +1,109 @@
...
+              name: 'Installation',
+              data: [43934, 52503, 57177, 69658, 97031, 119931, 137133, 154175]
+          }]
+          return data.to_json.html_safe
+        else
+          configured_data.to_json.html_safe
+        end
+      end
+
+      def self.options
```
terryfinn: Is this method being used?

app/pb_kits/playbook/pb_bar_graph/bar_graph.rb
```diff
@@ -0,0 +1,109 @@
...
+        new_hash = PROPS.map { |e| e.to_s.remove("configured_") }
```
terryfinn: This will not return a hash, and this variable is not being used.
```suggestion
        PROPS.map { |e| e.to_s.remove("configured_") }
```

app/pb_kits/playbook/pb_bar_graph/bar_graph.rb
```diff
@@ -0,0 +1,109 @@
...
+      def point_start
+        if configured_point_start == default_configuration
+          "2012"
+        else
+          configured_point_start
+        end
+      end
+
+      def chart_data
+        if configured_chart_data == default_configuration
```
terryfinn: This can be simplified if we take advantage of the fact that everything in ruby is an expression.
```ruby
def chart_data
  data = if configured_chart_data == default_configuration
           [{ name: "Installation", data: [43934, 52503, 57177, 69658, 97031, 119931, 137133, 154175] }]
         else
           configured_data
         end
  data.to_json.html_safe
end
```

app/pb_kits/playbook/pb_bar_graph/bar_graph.rb
```diff
@@ -0,0 +1,109 @@
...
+          data = [{
+              name: 'Installation',
```
terryfinn: This kit has a very generic name of `Playbook::PbBarGraph::BarGraph`, why do we have such a specific value of `'Installation'` in here?

app/pb_kits/playbook/pb_bar_graph/_bar_graph.html.erb
```diff
@@ -0,0 +1,13 @@
+<div class="<%= object.classname %>" id="mychart"></div>
```
terryfinn: You could reduce interpolation here with the content_tag helper:
```suggestion
<%= content_tag(:div, "", class: object.classname, id: "mychart") %>
```
I personally find it more readable in this case.

app/pb_kits/playbook/pb_charts/business_intelligence/pbLineGraph.js
```diff
@@ -0,0 +1,15 @@
...
+const markerStyles = function(highchart){
+  highchart.plotOptions.line.marker.enabled = true;
+  highchart.plotOptions.series.marker.enabled = true;
+  highchart.plotOptions.series.marker.fillColor = "white";
+  highchart.plotOptions.series.marker.lineWidth = 2;
+}
+
+const pbLineGraph = function(highchart) {
+  markerStyles(highchart);
+  pbHighcharts(highchart);
```
markiearnold: Should pbHighcharts() be called before markerStyles() ? 
Since markerStyles is a specific override for pbLineGraph, in my mind you should apply the default highchart settings, then run all the specific chart overrides.

app/pb_kits/playbook/pb_bar_graph/bar_graph.rb
```diff
@@ -0,0 +1,109 @@
...
+        when "vertical"
+          "column"
+        when "horizontal"
+          "bar" 
+        else
+          "column"
+        end
+      end
+
+      def title
```
markiearnold: All of these could utilize the helpers to be more condensed if you want to look into the default_value helper
```
def title
  default_value(configured_title, "")
end
```

app/pb_kits/playbook/pb_bar_graph/bar_graph.rb
```diff
@@ -0,0 +1,109 @@
+module Playbook
+  module PbBarGraph 
+    class BarGraph < Playbook::PbKit::Base
+      PROPS = [
+          :configured_classname,
```
markiearnold: Every kit should have these props defined in the PbKit::Base class:
classname, id, data

Please reference other kits

app/pb_kits/playbook/pb_line_graph/line_graph.rb
```diff
@@ -1,11 +1,12 @@
 module Playbook
   module PbLineGraph
-    class LineGraph
+    class LineGraph < Playbook::PbKit::Base
       PROPS = [:configured_class,
```
markiearnold: Every kit should have these props defined in the PbKit::Base class:
classname, id, data

`configured_class` should be `configured_classname`
Please reference other kits

app/pb_kits/playbook/pb_line_graph/line_graph.rb
```diff
@@ -32,6 +35,22 @@ def class_name
         end
       end
 
+      def gradient
```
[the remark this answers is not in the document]
terryfinn: Why does this work on string versions of true and false, and not just TrueClass (`true`) and FalseClass (`false`)?

app/pb_kits/playbook/pb_bar_graph/bar_graph.rb
terryfinn: Can we come up with a better name for the value yield to this block? What is `e`, would `prop` or `property` be more descriptive?

app/pb_kits/playbook/pb_charts/business_intelligence/pbHighcharts.js
jasperfurniss: Oops. Good catch. No explanation - Just a mistake.
Testing this now.

app/pb_kits/playbook/pb_line_graph/line_graph.rb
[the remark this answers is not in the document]
markiearnold: The is_true? above should only work on true/false, not strings.  "true" would return false.

`configured_gradient` value should be passed to the kit as a boolean.  The [PbKit::Base](https://github.com/powerhome/playbook/blob/master/app/pb_kits/playbook/pb_kit/base.rb) class provides some validation methods (which could use more eyes on it to improve things!)

app/pb_kits/playbook/pb_bar_graph/bar_graph.rb
jasperfurniss: It was just using specific data to be an  example  on the page. It doesn't belong here  in this file though! You're right.

app/pb_kits/playbook/pb_line_graph/line_graph.rb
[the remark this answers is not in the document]
jasperfurniss: Gotcha.

app/pb_kits/playbook/pb_line_graph/_line_graph.html.erb
```diff
@@ -1,8 +1,10 @@
+<script type="text/javascript" src="http://10.1.1.242:8081/js/sisense.js"></script>
```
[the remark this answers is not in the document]
jasperfurniss: Yeah i removed this. We definitely don't need this in here anymore. The pages will be fully aware of the sisense scripts. This was here from early stages of testing.

app/pb_kits/playbook/pb_bar_graph/_bar_graph.html.erb
```diff
@@ -0,0 +1,17 @@
+<%= content_tag(:div, "",
+  id: object.id(object.id),
+  )
+%>
```
markiearnold: For all Playbook Kits, the user needs to be able to add their own custom id, data-attributes and classes. To do this, we have to that the wrapping div work a specific way.

This should look like this:
```
<%= content_tag(:div, "",
    id: object.id("bar-graph"),
    data: object.data,
    class: object.classname("pb_bar_graph") %>
```

More details on this: https://github.com/powerhome/playbook/wiki/Understanding-Rails-Kit-HTML-Wrapper

app/pb_kits/playbook/pb_line_graph/_line_graph.html.erb
```diff
@@ -1,8 +1,13 @@
-<div class="<%= object.class_name %>" id="mychart"></div>
+<%= content_tag(:div, "",
+  id: object.id(object.id),
+  )
+%>
```
markiearnold: Same as above for `_bar_graph.html.erb`

app/pb_kits/playbook/pb_line_graph/line_graph.rb
```diff
@@ -1,87 +1,69 @@
...
-          :configured_point_start].freeze
+    class LineGraph < Playbook::PbKit::Base
+      PROPS = [ :configured_id,
+                :configured_classname,
+                :configured_chart_data,
+                :configured_title,
+                :configured_subtitle,
+                :configured_axis_title,
+                :configured_gradient,
+                :configured_point_start].freeze
```
markiearnold: You will need to add `:configured_data` here to get the functionality talked about above for data attributes.

You will also need to add data prop to the initialize block below.

app/pb_kits/playbook/pb_bar_graph/_bar_graph.html.erb
markiearnold: If you want the kit to require an ID from the user it should be written like this:

```
<%= content_tag(:div, "",
    id: object.id,
    data: object.data,
    class: object.classname("pb_bar_graph") %>
```

.rubocop_todo.yml
```diff
@@ -1,15 +1,11 @@
...
 # versions of RuboCop, may require this file to be generated again.
 
-# Offense count: 1
-# Configuration parameters: MinNameLength, AllowNamesEndingInNumbers, AllowedNames, ForbiddenNames.
-# AllowedNames: io, id, to, by, on, in, at, ip, db
-Naming/UncommunicativeMethodParamName:
+Style/FrozenStringLiteralComment:
   Exclude:
-    - 'app/pb_kits/playbook/pb_list/list.rb'
+    - 'app/pb_kits/playbook/pb_bar_graph/bar_graph.rb'
```
markiearnold: To prevent rubocop errors for List, this needs to stay as:
```
- 'app/pb_kits/playbook/pb_list/list.rb'
```

If you have an offense in `'app/pb_kits/playbook/pb_bar_graph/bar_graph.rb'`, please fix

jasperfurniss: No idea how that got that way. But i fixed it just now. Good catch!

.rubocop_todo.yml
```diff
@@ -1,14 +1,11 @@
...
 # The point is for the user to remove these configuration records
 # one by one as the offenses are removed from the code base.
 # Note that changes in the inspected code, or installation of new
 # versions of RuboCop, may require this file to be generated again.
 
-# Offense count: 1
-# Configuration parameters: MinNameLength, AllowNamesEndingInNumbers, AllowedNames, ForbiddenNames.
-# AllowedNames: io, id, to, by, on, in, at, ip, db
-Naming/UncommunicativeMethodParamName:
+Style/FrozenStringLiteralComment:
```
markiearnold: One last change.
Not sure why this changed, but it still needs to be `Naming/UncommunicativeMethodParamName:` to catch the known List offense of the "lg" prop.

PR #79, opened 2019-07-21
KIT: User

markiearnold:
# PbUser Kit

##### Props:
* name
* title
* avatar
* align (one of: left*, center, right)
* orientation (one of: vertical, horizontal*)
* size (one of: lg, md, sm*)

##### What is it?
A consistent way to display a user


app/pb_kits/playbook/pb_user/user.rb
```diff
@@ -0,0 +1,117 @@
...
+        kit_options.join("_")
+      end
+
+      def to_partial_path
+        "pb_user/user"
+      end
+
+    private
+
+      DEFAULT = Object.new
```
drborges: Would it be enough to have `default_configuration` to just return `Object.new`?

terryfinn: @drborges This pattern works because `Object.new != Object.new`. The value of Object.new has to be captured so it can be compared to other values.

Since `initialize` is an instance method, we could rewrite this using a private instance method that captures the value of `Object.new` into a instance variable.

That would look something like:
```ruby
class Example
  def initialize(example_option: default)
    ...
  end

  private

  def default
    @default ||= default
  end
end
```

The only real downside to this, is that we will be creating more instances of `Object.new`, but that is really not a big deal, and I would not be concerned with it.

drborges: Yup, that's what I had in mind. I'm just wondering if the existing boilerplate is worth the trouble...

terryfinn: I think I am picking up what your putting down, and too am questing if the current version is worth the boiler plate.

PR #80, opened 2019-07-21
KIT: Button Group

markiearnold:
# PbButtonGroup Kit

##### Props:
* orientation (one of: horizontal*, vertical)
* connect (boolean)
* &block

##### What is it?
Button group can be wrapped around PbButton Kits to make a button group with consistent spacing and styles. Available option to make connected groups.

### DIFF
--- a/app/pb_kits/playbook/packs/examples.js
+++ b/app/pb_kits/playbook/packs/examples.js
@@ -47,3 +47,6 @@ WebpackerReact.setup (Icon);
 
 import * as DashboardValue from "pb_dashboard_value/docs";
 WebpackerReact.setup (DashboardValue);
+
+import * as ButtonGroup from "pb_button_group/docs";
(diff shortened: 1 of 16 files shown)


jasoncypret:
@markiearnold can you also update the button style to match the latest from sketch? The button text use the title 4 I believe.

markiearnold:
I had unintended behavior on :active/:hover/:focus from removing the double border, so I removed them all... I will add them back in and clean this up to match sketch!

Also, thinking about it now - the prop to make the buttons connected I believe should be `join: true` instead of `connect: true`.  Thoughts? Or suggestions on another prop name?

thestephenmarshall:
👍 @markiearnold You probably mean to suggest changing the classname convention to use `_join` instead of `_connect` as well right?

thestephenmarshall:
Closing this and opening issue https://github.com/powerhome/playbook/issues/389

PR #82, opened 2019-07-21
KIT: Title Count

markiearnold:
# PbTitleCount Kit

##### Props:
* Title
* Count (string or integer)
* Size (one of: lg, sm*)
* Align (one of: left*, center, right)

##### What is it?
A basic kit to display a count next to a title. 
Both an integer or string can be passed to `count`. If integer, the kit will format larger numbers to have commas.


jasoncypret:
@markiearnold for some of these I'm wondering if we would want to make money specific versions to handle the more common things we do with money. Round, float, smaller cents, ect

markiearnold:
Good point, I didn't think of the money use case.

I did go back and try to define those settings in the kit to handle money, but I ended up getting stuck. I think we need to define all the options we may need.

jasoncypret:
Let's kick that can down the road. I think this is fine...but we might want to extend it with more features for money

PR #85, opened 2019-07-21
KIT: Pill

markiearnold:
# PbPill Kit

##### Props:
* text
* variant (one of: success, warning, error, info, neutral*)

##### What is it?
A basic kit to show status pills


app/pb_kits/playbook/pb_pill/_pill.scss
```diff
@@ -0,0 +1,33 @@
...
+  display: inline-flex;
+  justify-content: center;
+  align-items: center;
+  padding: 0 $space-sm/1.8;
+  height: $pb_pill_height;
+  border-radius: $pb_pill_height/2;
+
+  [class^=pb_title] {
+    &[class*=_4] {
+      font-size: $text-small !important;
```
jasoncypret: Wondering how we treat these types of things. This size doesn't exist as a text pattern...but is part of our old variables. However, should this just use base? I'm conflicted because it looks good. In any case...we should clean up this file: https://github.com/powerhome/playbook/blob/master/app/pb_kits/playbook/tokens/_typography.scss

jasoncypret:
I just requested feedback to draw some dialog. I'm conflicted on this one @markiearnold

markiearnold:
I updated the text to use the standard body kit (Non-bold, base size).  Thoughts @jasoncypret ?

PR #88, opened 2019-07-21
KIT: Time

markiearnold:
# PbTime Kit

##### Props:
* timestamp (DateTime or String)
* timezone (String, default: "America/New York")
* size (one of: lg, sm*)

##### What is it?
A consistent way to show time


jasoncypret:
@markiearnold I feel like the timezone should be deemphasized since it's caps. Thoughts?

markiearnold:
@jasoncypret Yes, I agree!  I updated kit and the screenshot in PR description. Thanks!

app/pb_kits/playbook/pb_time/time.rb
```diff
@@ -78,7 +76,7 @@ def timezone_abbr
       end
 
       def timezone
-        "<span class='pb__time_timezone'>#{timezone_abbr}</span>".html_safe if is_set? configured_timestamp
+        "<span class='pb__time_timezone'>#{timezone_abbr}</span>".html_safe
```
terryfinn: Why not use a [time tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time)? It allows use to more sematically represent this value, and we can set a `datetime` attribute to specify a more machine friendly value.

app/pb_kits/playbook/pb_time/time.rb
```diff
@@ -92,7 +90,7 @@ def icon
       end
 
       def text
-        "<span>#{format_time_string}</span> #{timezone}".html_safe if is_set? configured_timestamp
+        "<span>#{format_time_string}</span> #{timezone}".html_safe
```
terryfinn: Same as above.

app/pb_kits/playbook/pb_time/time.rb
terryfinn: Also, I would consider using the content tag helper to generate this HTML fragment.
You can delegate to `ActionView::Helpers::TagHelper`. This has the benefit of ensuring that we generate properly escaped markup. It is important to remember what calling `html_safe` on a object does. It is similar to taint and trust. They simply signify if something should be trusted or not, they do not make things safe. Calling `html_safe` just tells Rails to please not escape this string, it is trusted. It does not make sure things are properly escaped and safe. Using the tag helpers will escape values and make sure they are safe.

markiearnold: Great feedback! That makes sense, thank you

terryfinn: I would add, the time tag would be a good consideration for things other than times. It is appropriate for both time and dates.

PR #89, opened 2019-07-22
KIT: Icon Circle

markiearnold:
# PbIconCircle Kit

##### Props: 
* icon
* size (sm, md*, lg)
* variant (one of: default royal blue purple teal red yellow green)

##### What is it?
A basic kit to display an icon in a circle element


jasoncypret:
@markiearnold Would we ever use this as a placeholder for an avatar? If so I wonder if that is part of this kit...or part of avatar

markiearnold:
I think this should be a stand-alone kit to be used in ui that does not involve user avatars. For example, if this pattern was used to display a dashboard stat, we won't have to call PbAvatar to display the icon.

I think we should add a task to extend Avatar to have an option to utilize this kit OR use the monogram pattern if no image.

markiearnold:
I did make a change for this kit to have the same size options as avatar
`size_options = %w[xs sm md base lg xl]`

PR #90, opened 2019-07-22
KIT: User Badge

markiearnold:
# PbUserBadge Kit

##### Props:
* badge (one of: million-dollar*, veteran)
* size (one of: sm, md*, lg)

##### What is it?
A consistent way to show a user badge. 
Will be used in more complex user kits.


jasoncypret:
@wadewinningham should these be hosted on a CDN, or included as an asset with Playbook? I'm thinking the way it's set up right now these wouldn't be cached since they are raw SVG's.

wadewinningham:
The way it's setup seems fine to me.

PR #99, opened 2019-07-28
KIT: Source

markiearnold:

### DIFF
--- a/app/pb_kits/playbook/pb_source/source.rb
+++ b/app/pb_kits/playbook/pb_source/source.rb
@@ -1,0 +1,110 @@
+# frozen_string_literal: true
+
+module Playbook
+  module PbSource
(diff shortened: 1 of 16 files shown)


app/pb_kits/playbook/pb_source/source.rb
```diff
@@ -0,0 +1,110 @@
...
+        self.configured_data = data
+        self.configured_hide_icon = hide_icon
+        self.configured_id = id
+        self.configured_source = source
+        self.configured_type = type
+        self.configured_user = user
+      end
+
+      def source
+        pb_source_title = Playbook::PbTitle::Title.new(size: 4, text: configured_source.upcase)
```
jasoncypret: We shouldn't be uppercasing anything except captions.

PR #100, opened 2019-07-29
Remove references to storybook in package.json scripts section

thestephenmarshall:
This shouldn't be here it is copypasta from nitro-storybook repo.


jasoncypret:
Good catch. We need to remove tons of stuff in this package

PR #101, opened 2019-07-30
Demo/sisense dashboard value kit js

jasperfurniss:
This PR updates the Sisense dashboard to reflect the latest styles from our kits. 
It also reflects the latest restructuring and name changes to files.

Ready for feedback and any adjustments that need to be made!


app/views/layouts/playbook/_nav.html.slim
```diff
@@ -1,3 +1,4 @@
 .pb--page--topNav
   h1
     a href="#{root_path}" class="text-power-gold" Playbook
+  h6 class="pb_body_lighter" v2.8.0
```
markiearnold: I think this should use our Version number we are already defining, so we don't have to update in more places.  @thestephenmarshall - do you think it is ok to use `Playbook::VERSION` directly here, or should we do something different here?

Also I think the number should be beside the "Playbook" logo text, so the header does not grow in height.

Here is the example of this:
```
.pb--page--topNav
  h1
    = link_to "Playbook", root_url
    | &nbsp;
    = pb_rails("caption", props: { tag: "span", text: Playbook::VERSION })
```

Until we figure out spacing in our kits, the example above requires us to use the `&nbsp;` so there is a single space between the logo and the version number.  I also removed the "V", because caption forces capitalization and it looks strange to me capitalized.

thestephenmarshall: @markiearnold I think this works. Another option would be to add this to a helper within `app/helpers/playbook` e.g. 

```ruby
def playbook_version
  Playbook::VERSION
end
```

So that, **if you wanted** to later do some string interpolation with the version number you could. This is optional though.

PR #102, opened 2019-07-31
Bumped Minor Version

jasperfurniss:
TLDR: I had to bump to a minor version because of an accidental gem push from my own branch, instead of Master. That resulted in me "gem yanking" the version from the ruby gems registry, which in turn meant needing to cut a new version.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    playbook_ui (2.8.0)
+    playbook_ui (2.8.1)
       method_source (~> 0.9.2)
       rails (>= 5.1.6, < 6.0)
       redcarpet (= 3.4.0)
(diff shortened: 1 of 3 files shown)


jasoncypret:
@thestephenmarshall shouldn't the package.lock file update as well?

jasoncypret:
Actually looking at it I guess not

jasperfurniss:
> Actually looking at it I guess not

Yeah from what I can tell, nothing in the yarn.lock or package.lock gets changed. I did run yarn though after bumping the version number in the package.json, just to verify. Nothing changed prior. Always good to check though.

jasperfurniss:
> The gems: playbook_ui-2.8.1.gem & playbook_ui-2.8.0.gem should be removed from the repo CC: @markiearnold @thestephenmarshall

These get created when I ran  `gem build playbook.gemspec` and that was the first of two commands that Garrett mentioned to run in his old PR for the  version bump. Seems like maybe we should remove the 2.8.0 one, and keep the current version one.

markiearnold:
@jasperfurniss Both should be removed. None of the .gem files should be checked into the repo.  (Suggestion from Garett on a previous release)

jasperfurniss:
@jasoncypret @markiearnold 👍  Removed and pushed up.

PR #103, opened 2019-08-05
Fix as many of the yarn.lock alerts as possible & remove unused libs

thestephenmarshall:
## Fix yarn.lock alerts

This was an attempt to fix as many of the yarn.lock alerts as possible. Right now, it is ultimately up to the authors of most of these offending packages to do the update. We will continue to see these alerts until that happens. 

Screen Shot 2019-08-05 at 10 44 51 AM

- [yarn-audit.txt](https://github.com/powerhome/playbook/files/3468590/yarn-audit.txt)
- [lodash-package-json.txt](https://github.com/powerhome/playbook/files/3468593/lodash-package-json.txt)

## Unused libraries copied from `nitro-storybook` have been removed


package.json
```diff
@@ -4,10 +4,11 @@
   "description": "Nitro's Design System",
   "main": "./app/pb_kits/playbook/packs/index.js",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
+    "build": "./bin/webpack --profile --progress",
```
thestephenmarshall: Pretty standard across our web apps to have this here. It is nice to have a shortcut when you want to test out webpack without running `make start` and waiting forever.

package.json
```diff
@@ -35,22 +36,25 @@
     "file-loader": "3.0.1",
     "flow-bin": "0.54.1",
     "lqip-loader": "^2.2.0",
-    "webpack-dev-server": "^3.1.14"
+    "webpack-dev-server": "^3.7.2"
```
thestephenmarshall: We don't use this yet, but likely will very soon. It should be kept up-to-date.

package.json
```diff
@@ -35,22 +36,25 @@
...
   },
   "dependencies": {
     "@babel/cli": "7.2.3",
     "@babel/core": "^7.3.4",
+    "@babel/plugin-proposal-class-properties": "^7.5.5",
+    "@babel/plugin-syntax-dynamic-import": "^7.2.0",
+    "@babel/plugin-transform-runtime": "^7.5.5",
```
thestephenmarshall: These plugins were needed for transpilation when running `yarn build`

package.json
```diff
@@ -35,22 +36,25 @@
...
     "@babel/preset-react": "^7.0.0",
-    "@rails/webpacker": "^4.0.2",
+    "@rails/webpacker": "^4.0.7",
```
thestephenmarshall: Again, let's keep this one up-to-date. In this case, two for one deal (one yarn alert fixed).

package.json
```diff
@@ -35,22 +36,25 @@
...
     "babel-loader": "^8.0.5",
     "babel-plugin-flow-runtime": "0.11.1",
     "babel-plugin-jsx-control-statements": "3.2.8",
     "babel-plugin-transform-decorators-legacy": "1.3.4",
     "babel-plugin-transform-flow-strip-types": "6.22.0",
     "babel-plugin-transform-react-remove-prop-types": "^0.4.23",
+    "@babel/preset-env": "^7.3.4",
```
thestephenmarshall: This preset is showing up as missing when running `make build`. Adding it makes sense, but I am still investigating exactly why I am now seeing this error. Likely it could be due to https://github.com/powerhome/playbook/pull/103/files#r310702140

package.json
```diff
@@ -59,7 +63,7 @@
     "flow-bin": "0.54.1",
     "flow-runtime": "0.14.0",
     "lazysizes": "^4.1.6",
-    "lodash": "4.17.13",
+    "lodash": "4.17.15",
```
thestephenmarshall: Another lib that we will want to keep update to date regularly. This one had the most violations because it is so common. Knowing this, we will need to keep our eyes on libs like this and ensure that we are beginning to see less alerts from GH and `yarn audit` as time goes by.

PR #106, opened 2019-08-09
rubocop fixes for list kit

kre8sions:


app/pb_kits/playbook/pb_list/list.rb
```diff
@@ -62,11 +62,12 @@ def layout
...
-        if configured_lg == default_configuration
+      def size
+        case configured_size
+        when default_configuration
           ""
-        elsif configured_lg == true
-          "_lg"
+        when "large"
+          "_large"
         end
```
markiearnold: this can be written like this in the future

```
size_options = %w[large]
one_of_value(configured_size, size_options, "")
```

Usage:
`one_of_value(prop_value, available_options, default_value)`

PR #121, opened 2019-08-16
Build out Phone kit

eabousaif:


app/pb_kits/playbook/pb_phone/docs/_phone_default.html.erb
```diff
@@ -0,0 +1,4 @@
+<%= pb_rails("phone", props: {
+  icon: "mobile-android",
+  number: "(349) 185-9988"
```
jasoncypret: Can you add a formatted in the kit class that will take a number, strip out any non-numeric characters, then use `number_to_phone(raw_numbers, area_code: true)` to make the number look good?

eabousaif: Add a number formatter to the number method in phone.rb

app/pb_kits/playbook/pb_phone/docs/_phone_default.html.erb
```diff
@@ -0,0 +1,4 @@
+<%= pb_rails("phone", props: {
+  icon: "mobile-android",
+  number: "3491859988"
```
jasoncypret: I tried passing a string that might come through nitro `349-185-9988` & it breaks.

We want to first strip any non-numeric characters. Then after the number is only numbers, use this to format the number `number_to_phone(your_formatted_number, area_code: true)`

jasoncypret: You could use:
`"(627)222-2777".delete('^0-9')`
or 
`"(627)222-2777".tr('^0-9', '')`

eabousaif: Tried with different formats and it worked fine

PR #123, opened 2019-08-16
Owner kit

KFad11:
##What does this PR do?

Creates a new playbook kit for an Owner. Also fixes a piece of the Title kit, where the React side of the kit was using a deprecated class name for the header. The React side of the Owner kit will be added down the line.

Screen Shot 2019-08-16 at 10 15 38 AM


app/pb_kits/playbook/pb_owner/_owner.html.erb
```diff
@@ -0,0 +1,11 @@
+<%= content_tag(:div,
+    id: object.id,
+    data: object.data,
+    class: object.classname("pb_owner")) do %>
+      <div class="owner_first">
```
jasoncypret: We can remove this div wrapper. It's not being used.

app/pb_kits/playbook/pb_owner/_owner.html.erb
```diff
@@ -0,0 +1,11 @@
...
+        <div> <%= object.first_name %> </div>
+      </div>
+      <div class="owner_last">
+        <div> <%= object.last_name %> </div>
```
jasoncypret: Can you add `pb_owner_last ` as a classname prop in the kit class? Also lets remove this wrapper.

app/pb_kits/playbook/pb_owner/_owner.html.erb
KFad11: Removed

app/pb_kits/playbook/pb_owner/_owner.html.erb
KFad11: Corrected!

PR #124, opened 2019-08-16
Add Kit: Label Pill

darrendc:
Mock:
Screen Shot 2019-08-16 at 10 31 39 AM

After:
Screen Shot 2019-08-16 at 10 37 02 AM

Utilizes Pill and Caption Kit to create Label Pill


app/pb_kits/playbook/pb_label_pill/_label_pill.scss
```diff
@@ -0,0 +1,8 @@
+.pb_label_pill {
+  display: flex;
+  justify-content: flex-start;
+  align-items: center;
+  .pb_caption {
```
jasoncypret: Can you add a unique class to pb_caption and reference it here? Add a classname in your kit class file.

app/pb_kits/playbook/pb_label_pill/_label_pill.scss
```diff
@@ -0,0 +1,8 @@
+.pb_label_pill {
```
jasoncypret: You need to import any sass files from the kits you are using at the top. This will make sure your kit works by itself.

`@import "../pb_caption/caption";`
`@import "../pb_pill/pill";`

app/pb_kits/playbook/pb_label_pill/_label_pill.scss
darrendc: Added "pb_label_pill_label" class name

PR #126, opened 2019-08-16
Add all files for Owner kit

KFad11:


KFad11:
A small amount of back story here... 

The first PR was missing a semicolon, so we reverted it. I had 3 of the 14 files up in text editor when adding in the missing semicolon, saved those 3 files, assuming the other 11 files were still there but they were not. 

So the new PR wasn't going to have most of the new kit. I recovered the files I needed, added a single line at the bottom of them and pushed them up here to mak sure we have them. 

The caveat here is that Leila had pulled down my version of the playbook WITH all 14 files just without the semicolon because she needed to import my kit. She added the semicolon and her kit and my kit are working fine on her branch. So we're keeping these files here, in case her merging the branch somehow reverts my kit. Hoping all the files she has will get pushed up and merged correctly though, in that case, we should be able to just close this PR.

jasoncypret:
Found out why this build wasn't compiling. It was missing the owner.jsx file. I fixed this in a new branch so I'm closing this PR

PR #145, opened 2019-09-03
Playbook Rails JS Imports

thestephenmarshall:
Exports `pbChart` plugin which contains `Highcharts` instance for use alongside chart kits where needed.


app/pb_kits/playbook/packs/index.js
```diff
@@ -53,5 +56,6 @@ export {
   commonSettings,
   lineGraphSettings,
   barGraphSettings,
-  dashboardValueSettings
+  dashboardValueSettings,
+  pbChart,
```
thestephenmarshall: Provide the JS for use in Nitro and PG

app/pb_kits/playbook/plugins/pb_chart_plugin.js
```diff
@@ -0,0 +1,94 @@
+if(typeof(Highcharts) !== 'object') {
+  const script = document.createElement('script')
+  script.src   = 'https://code.highcharts.com/highcharts.js'
+  document.body.appendChild(script)
+}
```
thestephenmarshall: This chart plugin is actually 100% reliant on this script. Require it once here only.

config/webpacker.yml
```diff
@@ -56,7 +56,7 @@ development:
   webpack_compile_output: true
 
   # Verifies that versions and hashed value of the package contents in the project's package.json
-  check_yarn_integrity: true
+  check_yarn_integrity: false
```
thestephenmarshall: Disabling this for now. Too much trouble.

Dockerfile
```diff
@@ -30,4 +30,4 @@ RUN yarn install
 ADD --chown=app:app . /home/app/src
 RUN mkdir /etc/service/puma && ln -s /home/app/src/services/puma.sh /etc/service/puma/run
 
-RUN cd spec/dummy; RAILS_ENV=production SECRET_KEY_BASE=does_not_matter_here bin/rails assets:precompile
+RUN cd spec/dummy; RAILS_ENV=development SECRET_KEY_BASE=does_not_matter_here bin/rails assets:precompile --trace
```
thestephenmarshall: This will be removed before merging, but will likely be configured at a higher level using `ARG` from docker-compose.yml

thestephenmarshall: WIP PR for that https://github.com/powerhome/playbook/pull/146

PR #149, opened 2019-09-04
Kit Currency Code Review

rruiz85:
Need a code review for this brand new shinny kit! $$$


app/pb_kits/playbook/pb_currency/currency.rb
```diff
@@ -0,0 +1,100 @@
...
+      def unit
+        pb_unit = Playbook::PbBody::Body.new(classname: "unit", color: "light") do
+          default_value(configured_unit, "")
+        end
+        ApplicationController.renderer.render(partial: pb_unit, as: :object)
+      end
+
+      def kit_class
+        kit_options = [
+          "pb_currency",
```
markiearnold: We had to address an issue with our kit classes, this should be `pb_currency_kit`

All top level kit elements, the class structure should be `"#{kit_name}_kit"`

app/pb_kits/playbook/pb_currency/_currency.scss
```diff
@@ -0,0 +1,50 @@
+
+  [class^=pb_currency]{
```
markiearnold: Please see other comments for class structure. This should be `pb_currency_kit`

markiearnold:
Really great job on this kit @rruiz85 . I have added an issue to resolve the RuboCop errors that are not associated to your kit.

The requested change prevents style bleed.

#### Explanation:

The below CSS is intended only for the top level kit element:
```scss
[class^=pb_currency]{ 
    ...
}
```

But because it is looking for any class that begins with "pb_currency" it will effect all of the additional elements within your kit:
* `.pb_currency_wrapper`
* `.pb_currency_value`
* `.pb_currency_center`
* `.pb_currency_right`

If we change `[class^=pb_currency]` to [class^=pb_currency_kit], it will only effect the top level element as intended.

app/pb_kits/playbook/pb_currency/_currency.scss
```diff
@@ -0,0 +1,50 @@
...
+      display: flex;
+      align-items: flex-end;
+    }
+    & > [class*=pb_currency_value] {
+      line-height: .9;
+      }
+
+  }
+
+  [class^=pb_currency_kit_center] {
```
markiearnold: The playbook kit class is generated from the props, so the class can easily change depending on which props are passed.  This is why we do the *if matches* type css syntax (Example `[class^=pb_currency_kit]` and `[class*=pb_currency_value]`).

To make it less rigid, we should not search for `pb_currency_kit_center`, but instead search for a class that contains `_center` within the element `.pb_currency_kit`.

Like: 
```scss
[class^=pb_currency_kit] {
  &[class*=_center]  {
    ...
  }
}
```

Full Example:
```scss
[class^=pb_currency_kit]{
  &[class*=_center]  {
    & > [class^=pb_caption],
    & > [class^=pb_title],
    & > [class^=pb_body] {
      text-align: center;
    }
    
    // Move alignment scss here
    & > [class*=pb_currency_wrapper] {
      justify-content: center;
    }
  }
  
  &[class*=_right]  {
    & > [class^=pb_caption],
    & > [class^=pb_title],
    & > [class^=pb_body] {
      text-align: right;
    }

    // Move alignment scss here
    & > [class*=pb_currency_wrapper] {
      justify-content: flex-end;
    }
  }

}

[class^=pb_currency_wrapper] {
  display: flex;

  & > [class*=dollar_sign] {
    display: flex;
    align-items: flex-start;
  }
  & > [class*=unit] {
    display: flex;
    align-items: flex-end;
  }
  & > [class*=pb_currency_value] {
    line-height: .9;
  }
}
```

app/pb_kits/playbook/pb_currency/_currency.scss
```diff
@@ -0,0 +1,50 @@
...
+    & > [class*=_wrapper] {
+      justify-content: center;
+    }
+
+  }
+
+  [class^=pb_currency_kit_right] {
```
markiearnold: See comment above with example scss changes.

app/pb_kits/playbook/pb_currency/_currency.scss
rruiz85: makes senses.. Done!

PR #179, opened 2019-09-10
Create Checkbox Kit

kre8sions:
checkbox kit 
props:
text,
dark

### DIFF
--- a/app/pb_kits/playbook/pb_checkbox/_checkbox.html.erb
+++ b/app/pb_kits/playbook/pb_checkbox/_checkbox.html.erb
@@ -1,0 +1,10 @@
+<%= content_tag(:label, id: object.id, data: object.data, class: object.classname(object.kit_class)) do %>
+<input type="checkbox" value="<%=object.value%>" name="<%=object.name%>" <%=object.checked%>>
+<span class="pb_checkbox_checkmark">
+  <%= object.icon %>
+</span>
+<span class="pb_checkbox_label">
+  <%= object.text %>
+</span>
+
+<% end %>
--- a/app/pb_kits/playbook/pb_checkbox/_checkbox.jsx
+++ b/app/pb_kits/playbook/pb_checkbox/_checkbox.jsx
@@ -1,0 +1,42 @@
+/* @flow */
+
+import React from 'react'
+
+type CheckboxProps = {
+  dark?: Boolean,
+  tag: String,
+  label: String,
+  name:String,
+  value:String,
+  checked?: Boolean,
+}
+
+const Checkbox = ({
+  dark=false,
+  tag='label',
+  label,
+  name='',
+  value='',
+  checked=false,
+}: CheckboxProps) => {
+
+  const Tag = `${tag}`
+
+  return (
+    <Tag
+        className={
+        `pb_checkbox_kit` +
+        (dark === true ? '_dark' : '')
+      }
+    > <input type="checkbox" name={name} value={value} defaultChecked={checked}/>
+  <span className="pb_checkbox_checkmark">
+      <i className="far fa-check check_icon"></i>
+    </span>
+    <span className="pb_checkbox_label">
+      {label}
+    </span>
+    </Tag>
+  )
+}
+
+export default Checkbox
--- a/app/pb_kits/playbook/pb_checkbox/_checkbox.scss
+++ b/app/pb_kits/playbook/pb_checkbox/_checkbox.scss
@@ -1,0 +1,66 @@
+@import "../tokens/colors";
+$transition: $transition_cubic;
+
+[class^=pb_checkbox_kit] {
+  .pb_checkbox_label {
+    display: block;
+    position: relative;
+    padding-left: $space_lg;
+    cursor: pointer;
+    font-size: $text_lt_default;
+    user-select: none;
+  }
+  .pb_checkbox_checkmark {
+    position: absolute;
+    height: 22px;
+    width: 22px;
+    border: solid $border_light 2px;
+    border-radius: $border_rad_light;
+    .check_icon{
+      opacity: 0;
+      position: relative;
+      top: -2px;
+      left: 1px;
+      height: 16px;
+      width:16px;
+      color: $white;
+
+    }
+  }
+
+  &:hover input ~ .pb_checkbox_checkmark {
+    border-color: $primary_action;
+  }
+
+  input {
+    position: absolute;
+    opacity: 0;
+    cursor: pointer;
+    height: 0;
+    width: 0;
+    display: none;
+    &:checked ~ .pb_checkbox_checkmark {
+      background-color: $primary_action;
+      border-color: $primary_action;
+    }
+    &:checked ~ .pb_checkbox_checkmark {
+      display: block;
+      .check_icon{
+        opacity: $opacity_10;
+      }
+    }
+  }
+
+
+
+  &[class*=_dark] {
+    input:checked ~ .pb_checkbox_checkmark {
+      background-color: $primary_action_dk;
+      border-color: $primary_action_dk;
+    }
+
+    &:hover .pb_checkbox_checkmark {
+      border-color: $primary_action_dk;
+    }
+  }
+}
--- a/app/pb_kits/playbook/packs/examples.js
+++ b/app/pb_kits/playbook/packs/examples.js
@@ -128,3 +128,6 @@ WebpackerReact.setup (SectionSeparator);
 
 import * as Currency from "pb_currency/docs";
 WebpackerReact.setup (Currency);
+
+import * as Checkbox from "pb_checkbox/docs";
+WebpackerReact.setup (Checkbox);
--- a/app/pb_kits/playbook/packs/kits.js
+++ b/app/pb_kits/playbook/packs/kits.js
@@ -38,3 +38,4 @@ import "./pb_stat_value.js";
 import "./pb_stat_change.js";
 import "./pb_section_separator.js";
 import "./pb_currency.js";
+import "./pb_checkbox.js";
--- a/app/pb_kits/playbook/packs/pb_checkbox.js
+++ b/app/pb_kits/playbook/packs/pb_checkbox.js
@@ -1,0 +1,4 @@
+import Checkbox from "pb_checkbox/_checkbox.jsx";
+import WebpackerReact from "webpacker-react";
+
+WebpackerReact.setup({ Checkbox });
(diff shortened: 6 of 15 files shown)


app/pb_kits/playbook/pb_checkbox/_checkbox.html.erb
```diff
@@ -0,0 +1,9 @@
+<%= content_tag(:label, id: object.id, data: object.data, value: object.value, name: object.name, class: object.classname(object.kit_class)) do %>
+
+<input type="checkbox" name="<%=object.name%>" <%=object.checked%> ></input>
+<span class="pb_checkbox_checkmark">
+<%= pb_rails("icon", props: { icon: "check",id:"check_icon"}) %></span>
```
markiearnold: I would suggest creating the icon through the kit class.  See example in [Icon Circle](https://github.com/powerhome/playbook/blob/master/app/pb_kits/playbook/pb_icon_circle/icon_circle.rb#L29)

app/pb_kits/playbook/pb_checkbox/_checkbox.scss
```diff
@@ -0,0 +1,66 @@
...
+    -moz-user-select: none;
+    -ms-user-select: none;
+    user-select: none;
+  }
+  .pb_checkbox_checkmark {
+    position: absolute;
+    height: 22px;
+    width: 22px;
+    border: solid $border_light 2px;
+    border-radius: 4px;
```
markiearnold: This should use the tokens - [Border Radius Tokens](https://github.com/powerhome/playbook/blob/master/app/pb_kits/playbook/tokens/_border_radius.scss#L6)

app/pb_kits/playbook/pb_checkbox/_checkbox.scss
```diff
@@ -0,0 +1,66 @@
+@import "../tokens/colors";
+$transition: 0.2s cubic-bezier(0.95, 0.05, 0.795, 0.035);
+
+[class^=pb_checkbox_kit] {
+  .pb_checkbox_label {
+    display: block;
+    position: relative;
+    padding-left: 30px;
```
markiearnold: Could this use a spacing token? - [Spacing Tokens](https://github.com/powerhome/playbook/blob/master/app/pb_kits/playbook/tokens/_spacing.scss#L7)

app/pb_kits/playbook/pb_checkbox/_checkbox.scss
```diff
@@ -0,0 +1,66 @@
+@import "../tokens/colors";
+$transition: 0.2s cubic-bezier(0.95, 0.05, 0.795, 0.035);
```
markiearnold: Should this be added to the transitions as a token? - [Transition Tokens](https://github.com/powerhome/playbook/blob/master/app/pb_kits/playbook/tokens/_transition.scss)

app/pb_kits/playbook/pb_checkbox/_checkbox.scss
```diff
@@ -0,0 +1,66 @@
...
+    -ms-user-select: none;
+    user-select: none;
+  }
+  .pb_checkbox_checkmark {
+    position: absolute;
+    height: 22px;
+    width: 22px;
+    border: solid $border_light 2px;
+    border-radius: 4px;
+    .pb_icon_kit{
```
markiearnold: I would suggest adding the prop `classname` to the icon kit, and adding a custom class to adjust the styles for the icon within this kit.

The reason for this suggestion, is because the kit class is dynamically created - if it is updated in the future this kit's scss will break.  If you add a custom class, it will not break.

Example:
```ruby
# In pb_checkbox.rb

pb_icon = Playbook::PbIcon::Icon.new(icon: "check", id: "check_icon", classname: "my_custom_icon", fixed_width: true)
ApplicationController.renderer.render(partial: pb_icon, as: :object)
```

```scss
/* In _checkbox.scss */

.my_custom_icon {
  ...
}
```

app/pb_kits/playbook/pb_checkbox/_checkbox.scss
```diff
@@ -0,0 +1,66 @@
...
+    width: 0;
+    display: none;
+    &:checked ~ .pb_checkbox_checkmark {
+      background-color: $primary_action;
+      border-color: $primary_action;
+    }
+    &:checked ~ .pb_checkbox_checkmark {
+      display: block;
+      .pb_icon_kit{
+        opacity:1;
```
markiearnold: This should use opacity tokens = [Opacity Tokens](https://github.com/powerhome/playbook/blob/master/app/pb_kits/playbook/tokens/_opacity.scss)

app/pb_kits/playbook/pb_checkbox/_checkbox.html.erb
```diff
@@ -0,0 +1,9 @@
+<%= content_tag(:label, id: object.id, data: object.data, value: object.value, name: object.name, class: object.classname(object.kit_class)) do %>
```
markiearnold: I do not think the `value` and `name` is needed on the label element. Is there a reason for this that I may have not thought about?

jasoncypret: Value and name should be on the checkbox

app/pb_kits/playbook/pb_checkbox/_checkbox.html.erb
jasoncypret: Does this give an advantage over this method @markiearnold  ?

app/pb_kits/playbook/pb_checkbox/_checkbox.html.erb
```diff
@@ -0,0 +1,9 @@
...
+
+<input type="checkbox" name="<%=object.name%>" <%=object.checked%> ></input>
+<span class="pb_checkbox_checkmark">
+<%= pb_rails("icon", props: { icon: "check",id:"check_icon"}) %></span>
+<span class="pb_checkbox_label">
+<%= object.text %>
+</span>
```
jasoncypret: I don't think this file is indented. Can you please indent this

app/pb_kits/playbook/pb_checkbox/_checkbox.scss
```diff
@@ -0,0 +1,66 @@
...
+  .pb_checkbox_label {
+    display: block;
+    position: relative;
+    padding-left: 30px;
+    cursor: pointer;
+    font-size: $text_lt_default;
+    -webkit-user-select: none;
+    -moz-user-select: none;
+    -ms-user-select: none;
+    user-select: none;
```
jasoncypret: I'm not sure we need these browser prefixes https://caniuse.com/#feat=user-select-none

app/pb_kits/playbook/pb_checkbox/_checkbox.scss
```diff
@@ -0,0 +1,66 @@
...
+  }
+  .pb_checkbox_checkmark {
+    position: absolute;
+    height: 22px;
+    width: 22px;
+    border: solid $border_light 2px;
+    border-radius: 4px;
+    .pb_icon_kit{
+      opacity:0;
```
jasoncypret: add a space between the colon

app/pb_kits/playbook/pb_checkbox/_checkbox.scss
```diff
@@ -0,0 +1,66 @@
...
+      position: relative;
+      top:-2px;
+      left:1px;
```
jasoncypret: add a space between the colons

app/pb_kits/playbook/pb_checkbox/_checkbox.scss
jasoncypret: add a space between the colon

app/pb_kits/playbook/pb_checkbox/_checkbox.html.erb
markiearnold: @jasoncypret I wasn't sure if it would effect performance since it had to go through the pb_rails helper logic rather than using the kit class directly. Not sure if that is an actual concern we should worry about. The only other reason for the suggestion was that all other kits are referencing other kits from within the Class, and it would be consistent.

app/pb_kits/playbook/pb_checkbox/_checkbox.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= content_tag(:label, id: object.id, data: object.data, class: object.classname(object.kit_class)) do %>
+<input type="checkbox" value="<%=object.value%>" name="<%=object.name%>" <%=object.checked%>></input>
```
markiearnold: No need for `</input>` here, input is a self closing element.

app/pb_kits/playbook/pb_checkbox/_checkbox.html.erb
```diff
@@ -0,0 +1,8 @@
...
+<span class="pb_checkbox_checkmark">
+  <%= object.icon %>
+  <span class="pb_checkbox_label">
+    <%= object.text %>
+  </span>
```
markiearnold: I think we are missing a closing `</span>` here

app/pb_kits/playbook/pb_checkbox/_checkbox.jsx
```diff
@@ -0,0 +1,33 @@
+/* @flow */
+
+import React from 'react'
+
+type CheckboxProps = {
+  children: Array<React.ReactNode> | React.ReactNode,
```
thestephenmarshall: Will this component be using it's `children` props as well? I noticed it isn't currently in the rendered output.

app/pb_kits/playbook/pb_checkbox/_checkbox.jsx
```diff
@@ -0,0 +1,33 @@
...
+
+  const Tag = `${tag}`
+
+  return (
+    <Tag
+        className={
+        `pb_checkbox_kit` +
+        (dark === true ? '_dark' : '')
+      }
+    > <input type="checkbox"/>
```
thestephenmarshall: Also, I think you might want to allow another prop in this kit called `name` so that devs can set it's value which will be used in any Rails form mappings. This prop is also used in a conventional nature in React apps within Nitro as well. e.g.

```jsx
<input name={name} type="checkbox"/>
```

This new prop would also open the doors to composing a toggle if need be in the future.

thestephenmarshall: See https://github.com/powerhome/playbook/pull/239/files for the `Toggle` example I mentioned above

PR #194, opened 2019-09-16
Update react kit gen

markiearnold:


markiearnold:
Updated changes:

In terminal, when creating a kit with the --props option:
```bash
rails g kit "Alert" --props message:string status:array active:boolean
```

Generates the following react kit:
```jsx
/* @flow */
/*eslint-disable react/no-multi-comp, flowtype/space-before-type-colon */

import React from 'react'

type AlertProps = {
  active?: Boolean,
  className?: String,
  data?: String,
  id?: String,
  message?: String,
  status?: 'option_1' | 'option_2',
  
}

const Alert = ({ active, className, data, id, message, status }: AlertProps) => (
  <div>
    {`kit content`}
  </div>
)

export default Alert
```

PR #244, opened 2019-09-19
weekday stacked kit created

christinaatai:


app/pb_kits/playbook/pb_weekday_stacked/_weekday_stacked.scss
```diff
@@ -0,0 +1,17 @@
+@import "../pb_caption/caption"; //used for weekday (caption)
+@import "../pb_title/title";   //used for date (title 1)
+
+[class^=pb_weekday_stacked_kit] {
+    &[class*=_center]  {
+      & > [class^=pb_caption],
+      & > [class^=pb_title] {
+        text-align: center;
```
markiearnold: Although `& > [class^=pb_caption]` works to target the nested kits, it is potentially dangerous to target the dynamic class.  A better way to target the nested kits is like below:

Add a custom class through the prop `classname`, and use the custom class within your scss.

```ruby
def weekday
  if is_set? timestamp_weekday
    pb_weekday = Playbook::PbCaption::Caption.new(text: timestamp_weekday, dark: dark, classname: my_custom_class)
    ApplicationController.renderer.render(partial: pb_weekday, as: :object)
  end
end
```

```scss 
[class^=pb_weekday_stacked_kit] {
    &[class*=_center]  {
        .my_custom_class {
              ...
         }
    }
}
```

app/pb_kits/playbook/pb_weekday_stacked/_weekday_stacked.scss
```diff
@@ -0,0 +1,17 @@
...
+      }
+  }
+    &[class*=_right]  {
+      & > [class^=pb_caption],
+      & > [class^=pb_title] {
+        text-align: right;
```
markiearnold: See comment above about using the `classname` prop

app/pb_kits/playbook/pb_weekday_stacked/weekday_stacked.rb
```diff
@@ -0,0 +1,112 @@
...
+        variant_options = %w[default basic minimal]
+        one_of_value(configured_variant, variant_options, "default")
+      end
+
+      def timestamp
+        convert_to_timestamp(configured_timestamp)
+      end
+
+      def convert_to_timestamp(time)
+        time.is_a?(String) ? DateTime.parse(time) : time
```
terryfinn: Why are we allowing string values to be passed in? Seems like it would be simpler to force the user to provide a date time like object. Are we taking into account time zone? When we parse a string value, what time zone are we using to convert the value into?

app/pb_kits/playbook/pb_weekday_stacked/weekday_stacked.rb
```diff
@@ -0,0 +1,112 @@
...
+      end
+
+      def timestamp_weekday
+        configured_variant == "minimal" ? timestamp.strftime("%^a").first(1) : timestamp.strftime("%^a")
```
terryfinn: If `Playbook::PbWeekdayStacked::WeekdayStacked#configured_variant` is `"minimal"`, then this will take the first letter from the abbreviated name. That would mean that for both Tuesday and Thursday I would see `"T"` returned?

app/pb_kits/playbook/pb_weekday_stacked/weekday_stacked.rb
```diff
@@ -0,0 +1,112 @@
...
+      end
+
+      def timestamp_date
+        configured_variant == "default" ? "#{month}/#{day}" : timestamp.strftime("%-d")
```
terryfinn:
```suggestion
        configured_variant == "default" ? "#{month}/#{day}" : day
```

app/pb_kits/playbook/pb_weekday_stacked/_weekday_stacked.html.erb
```diff
@@ -0,0 +1,12 @@
+<%= content_tag(:div,
```
terryfinn: Did you consider using a [Time tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time)? It would provide a more semantic way of presenting users with this value, and provides a canonical way for providing a machine readable value.

app/pb_kits/playbook/pb_weekday_stacked/weekday_stacked.rb
jasoncypret: Yup this is designed for smaller mobile interfaces and makes sense if you display a full week.

app/pb_kits/playbook/pb_weekday_stacked/_weekday_stacked.html.erb
jasoncypret: That's a great idea!

PR #247, opened 2019-09-20
Enable Dashboard Value react kit

markiearnold:
* Enables Dashboard Value react kit
* Enables Stat Value react kit
* Enables Stat Change react kit
* Enables Icon react kit
* Bug: Sets PbBody react kit default color to "default" instead of "light"

Also sets all documentation examples 1:1 between Rails and React for the above kits.

---

### DIFF
--- a/app/pb_kits/playbook/pb_dashboard_value/docs/_dashboard_value_align.jsx
+++ b/app/pb_kits/playbook/pb_dashboard_value/docs/_dashboard_value_align.jsx
@@ -1,0 +1,31 @@
+import React from "react"
+import DashboardValue from "../_dashboard_value.jsx"
+
+function DashboardValueAlign() {
+  return (
+    <div>
+      <DashboardValue
+          stat_label={{label: "Top Title Value"}}
+          stat_value={{value: "1,428", unit: "appts"}}
+          stat_change={{change: "decrease", value: "26.1%"}} />
+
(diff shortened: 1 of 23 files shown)


app/pb_kits/playbook/pb_dashboard_value/docs/_dashboard_value_align.jsx
```diff
@@ -0,0 +1,31 @@
+import React from "react"
+import DashboardValue from "../_dashboard_value.jsx"
+
+function DashboardValueAlign() {
```
jasperfurniss: `function DashboardValueAlign() {`   can also be written  `const DashboardValueAlign = () => {`  .... Wasn't sure if we had a standard for this yet. Carlos pushed me toward using the latter syntax. I'm actually going through all my hometour syntax now and fixing these to use the "const" syntax. What do ya'll think?

markiearnold: I am referencing the syntax of the other react kits. I will leave this decision to @thestephenmarshall and I will follow!

thestephenmarshall: It isn't super critical, but yeh I would generally defer to using the lambda version.

PR #249, opened 2019-09-20
Created Icon Title Kit

kre8sions:


package.json
```diff
@@ -64,6 +64,7 @@
     "jsx-control-statements": "^3.2.8",
     "lazysizes": "^4.1.6",
     "lodash": "4.17.15",
+    "node-sass": "^4.12.0",
```
thestephenmarshall: Why do we need to update this?

yarn.lock
```diff
@@ -5201,6 +5201,7 @@ node-releases@^1.1.8:
 node-sass@^4.12.0:
   version "4.12.0"
   resolved "https://registry.yarnpkg.com/node-sass/-/node-sass-4.12.0.tgz#0914f531932380114a30cc5fa4fa63233a25f017"
+  integrity sha512-A1Iv4oN+Iel6EPv77/HddXErL2a+gZ4uBeZUy+a8O35CFYTXhgA8MgLCWBtwpGZdCvTvQ9d+bQxX/QC36GDPpQ==
```
thestephenmarshall: This is happening bc the above `node-sass` change

app/pb_kits/playbook/pb_icon_title/_icon_title.jsx
```diff
@@ -0,0 +1,59 @@
...
+      `pb_icon_title_kit`,
+      className,
+      dark,
+      column
+
+    ])
+
+    return (
+      <div className={css}>
+        {value}
```
thestephenmarshall: @kre8sions can we also update the React kit as well?

package.json
kre8sions: @thestephenmarshall should i just delete that line?  I think that might have changed when I was going through the common errors page.

yarn.lock
kre8sions: Just yarn uninstall node-sass? @thestephenmarshall

app/pb_kits/playbook/pb_icon_title/_icon_title.html.erb
```diff
@@ -0,0 +1,12 @@
+<%= content_tag(:div, id: object.id,  data: object.data, class: object.classname(object.kit_class)) do %>
+
+<%= pb_rails("icon_circle", props: {
+  icon: "#{object.icon}",
+  size: object.column_sizing ,
+  variant:"#{object.icon_variant}"}) %>
+<div class="content">
+
+  <%= pb_rails("title", props: { text: "Value", size: object.size === "lg" ? 1 : 3 , dark: object.dark === "dark" ?  true : false }) %>
+  <%= pb_rails("caption", props: { text: "Description" , dark: object.dark === "dark" ?  true : false}) %>
```
markiearnold: Since we are doing logic on prop values for `Title` and `Caption`, this should be moved to icon_title.rb.  Please see [Label Value](https://github.com/powerhome/playbook/blob/master/app/pb_kits/playbook/pb_label_value/label_value.rb#L26) kit for reference.

markiearnold: @kre8sions To help on this, I thought I would help explain this a bit more!  

By bringing in the title and caption nested kits into the icon_title.rb file, you can do all the logic within your kit class:
```ruby
def value
  if is_set? configured_value
    title_value = default_value(configured_value, "Value")
    size_value = size === "lg" ? 1 : 3
    dark_value = dark === "dark" ?  true : false
    
    pb_value = Playbook::PbTitle::Title.new(text: title_value, size: size_value, dark: dark_value)
    ApplicationController.renderer.render(partial: pb_value, as: :object)
  end
end

def description 
  dark_value = dark === "dark" ?  true : false
  
  pb_description = Playbook::PbCaption::Caption.new(text: description, dark: dark_value)
  ApplicationController.renderer.render(partial: pb_description, as: :object)
end
```

And then your kit partial (_icon_title.html.erb) would look like the following:
```erb
<%= content_tag(:div, 
    id: object.id,  
    data: object.data, 
    class: object.classname(object.kit_class)) do %>

  <%= pb_rails("icon_circle", props: {
    icon: "#{object.icon}",
    size: object.column_sizing ,
    variant:"#{object.icon_variant}"}) %>
  <div class="content">
    <%= object.value %>
    <%= object.description %>
  </div>
  
<% end %>
```

app/pb_kits/playbook/pb_icon_title/_icon_title.html.erb
```diff
@@ -0,0 +1,12 @@
+<%= content_tag(:div, id: object.id,  data: object.data, class: object.classname(object.kit_class)) do %>
+
+<%= pb_rails("icon_circle", props: {
+  icon: "#{object.icon}",
+  size: object.column_sizing ,
+  variant:"#{object.icon_variant}"}) %>
+<div class="content">
+
+  <%= pb_rails("title", props: { text: object.value, size: object.size === "lg" ? 1 : 3 , dark: object.dark === "dark" ?  true : false }) %>
```
jasoncypret: Lets move this logic to the rb file

xjunior: Just the size logic. Also, the dark logic seem weird as it's declared as boolean everywhere and at some point the default value becomes a string. It would be cool if this component was moved to the new model already as this PR wasn't merged yet.

thestephenmarshall:
Closing and creating issue https://github.com/powerhome/playbook/issues/396

PR #276, opened 2019-09-24
WIP: Range slider

drborges:
# TODO

- [ ] Prevent CSS leaks
- [ ] Find workaround to Chrome not supporting `-fill-lower` range pseudo element.
- [ ] Implement React version


drborges:
@jasoncypret, @markiearnold I'm opening this WIP PR so you may take a look at the CSS, I've attempted a few different ideas (pure CSS solutions only) in order to work around the fact that Chrome does not provide a pseudo-element for styling the range progress as Mozilla and Edge do and this is as close I was able to get:



The current solution leverages `box-shadow` on the slider's `thumb` pseudo-element in order to simulate the "progress" part of the range (the blue section on the left-hand side), but unfortunately, I can't think of a way to "fix" the height of the progress section.

drborges:
I think I have a workaround :)

drborges:
Switching the direction of the `box-shadow` makes it slightly better:



I can explore a javascript based alternative if this solution isn't ideal.

app/pb_kits/playbook/pb_range_slider/_range_slider.scss
```diff
@@ -0,0 +1,102 @@
...
+    &:after {
+      content: '';
+      background: #808080;
+      display: block;
+      width: 100vw;
+      height: 4px;
+    }
+  }
+
+  @function make-progress($color) {
```
jasoncypret: This is the reason I think we should do this in JS. Looking at the output of the background it adds a significant amount of CSS to the page.

drborges: FYI, I'll get back to this at some point today.

thestephenmarshall:
Closed and issue opened https://github.com/powerhome/playbook/issues/397

Issue #206, opened 2019-09-17
Loading Inline


jasoncypret:
Use Pulse from icon: https://playbook.powerapp.cloud/kits/icon

Design details: https://tech.powerhrg.com/playgrounds/sketch_measure/new_kits/#artboard2

Use existing kits to build

![image](https://user-images.githubusercontent.com/863031/65063922-dff6ae80-d944-11e9-8edd-111b3d37c127.png)

PR #279, opened 2019-09-24
Loading Inline kit implementation for Rails

ghaydarov:
This PR is the work done during UX session on 2019.9 Create. It implements Loading Inline kit for Rails.


jasoncypret:
This resolves #206

Issue #182, opened 2019-09-12
Distribution Bar


jasoncypret:
Design Details: https://tech.powerhrg.com/playgrounds/sketch_measure/new_kits/#artboard21

This kit should have a large and small version. The user should be able to define how many parts are apart of the distribution. The colors should be automatically generated from the Status colors within the Playbook tokens. The user should be able to define the values for distribution.

![image](https://user-images.githubusercontent.com/863031/64826811-e0cbc100-d58f-11e9-8333-bdc41d63db51.png)

PR #285, opened 2019-09-24
[#182] Distribution Bar

smkopp92:
Screen Shot 2019-09-24 at 3 16 17 PM


app/pb_kits/playbook/pb_distribution_bar/_distribution_bar.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= content_tag(:div,
```
terryfinn: Did you consider using a [figure tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figure)? It would be a more semantic.

app/pb_kits/playbook/pb_distribution_bar/_distribution_bar.scss
```diff
@@ -0,0 +1,29 @@
...
+    &:first-child {
+      border-top-left-radius: 18px;
+      border-bottom-left-radius: 18px;
+    }
+    &:last-child {
+      border-top-right-radius: 18px;
+      border-bottom-right-radius: 18px;
+    }
+  }
+  &[class*=_sm] {
```
terryfinn: Does this class need to be this generic? It will match anything that that has a `_sm` in it. That means it would match `.pb_distribution_bar_smooth`. Seems like that is overly generic.

app/pb_kits/playbook/pb_distribution_bar/distribution_bar.rb
```diff
@@ -0,0 +1,57 @@
...
+
+      DEFAULT = Object.new
+      private_constant :DEFAULT
+      def default_configuration
+        DEFAULT
+      end
+      attr_accessor(*PROPS)
+
+      def normalize_characters(value)
+        return value.gsub!(/(\d\.\d)|[^a-zA-Z\d]/, '\\1').to_i if value.is_a? String
```
terryfinn: What are we doing with this regex here? First part matches any digit followed by a period followed by another digit, I am not sure what we are trying to match on the second part, and why are we replacing matches with `'\\1'`?

I would avoid checking the object type if we can, checking it is a smell.

app/pb_kits/playbook/pb_distribution_bar/_distribution_bar.html.erb
smkopp92: I think this is default from the playbook kit generator and thus the pattern for most kits in playbook.

app/pb_kits/playbook/pb_distribution_bar/distribution_bar.rb
```diff
@@ -0,0 +1,57 @@
...
+        value.to_i
+      end
+
+      def values_to_percents
+        total = configured_values.map {|v| normalize_characters(v) }.sum
```
terryfinn: `v` does not communicate much as a variable. I would consider being more verbose.

Alternatively, path the method to map instead of a block.
```suggestion
        total = configured_values.map(&method(:normalize_characters)).sum
```

app/pb_kits/playbook/pb_distribution_bar/_distribution_bar.scss
```diff
@@ -0,0 +1,29 @@
+@import "../tokens/colors";
+
+[class^=pb_distribution_bar] {
```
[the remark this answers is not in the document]
markiearnold: Currently this needs to be this complex. Playbook kits generate a single, dynamic class based on props.  For example the complete class for this kit could be `pb_distribution_bar_sm`.  `.pb_distribution_bar` would not apply the styles to the element in this case

app/pb_kits/playbook/pb_distribution_bar/distribution_bar.rb
jasoncypret: We are able to take any value and normalize it:

`values: [3242, "2425.60", 8292, "$5,662.98"]`

app/pb_kits/playbook/pb_distribution_bar/_distribution_bar.scss
```diff
@@ -0,0 +1,29 @@
...
+    &:last-child {
+      border-top-right-radius: 18px;
+      border-bottom-right-radius: 18px;
+    }
+  }
+  &[class*=_sm] {
+    height: 8px;
+  }
+  &[class*=_lg] {
+    height: 36px;
```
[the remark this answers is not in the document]
jasoncypret: Yes, we use relative units where we can.

app/pb_kits/playbook/pb_distribution_bar/_distribution_bar.scss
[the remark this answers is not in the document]
terryfinn: It took me awhile to work through this, but this is what I am understanding these rules to try to generate.

```
[class^=pb_distribution_bar] {
  // common rules for pb distribution bar

  &[class*=_sm] {
    // specific pb distribution bar small rules
  }
  &[class*=_lg] {
    // specific pb distribution bar large rules
  }
}
```
To avoid writing out

```
.pb_distribution_bar_sm {
   // specific pb distribution bar small rules
  // common rules for pb distribution bar
}

.pb_distribution_bar_lg {
  // specific pb distribution bar large rules
  // common rules for pb distribution bar
}
```

Have we considered a mixin to allow the two rule to share the common rules?

I find this code very difficult to read and understand as it is written. I also have concerns with the grep-ability of these kids of rules. Have we considered the performance of rules written like this, I would be suspicious of the rules performance.

app/pb_kits/playbook/pb_distribution_bar/_distribution_bar.scss
```diff
@@ -0,0 +1,31 @@
+@import "../tokens/colors";
+
+[class^=pb_distribution_bar] {
```
terryfinn: It took me awhile to work through this, but this is what I am understanding these rules to try to generate.

```
[class^=pb_distribution_bar] {
  // common rules for pb distribution bar

  &[class*=_sm] {
    // specific pb distribution bar small rules
  }
  &[class*=_lg] {
    // specific pb distribution bar large rules
  }
}
```
To avoid writing out

```
.pb_distribution_bar_sm {
   // specific pb distribution bar small rules
  // common rules for pb distribution bar
}

.pb_distribution_bar_lg {
  // specific pb distribution bar large rules
  // common rules for pb distribution bar
}
```

Have we considered a mixin to allow the two rule to share the common rules?

I find this code very difficult to read and understand as it is written. I also have concerns with the grep-ability of these kids of rules. Have we considered the performance of rules written like this, I would be suspicious of the rules performance.

jasoncypret: It depends on the kit, but this reduces repeating common styles over and over. This is like using a BEM/Atomic Design approach and combining with class chaining in order (in a protected way) to avoid repeating common things over and over and bloating our output CSS.

We have an example similar to what you are describing and this produces very tight classes but lots of repeated styles.

https://github.com/powerhome/playbook/blob/master/app/pb_kits/playbook/pb_layout/_layout.scss
https://github.com/powerhome/playbook/blob/master/app/pb_kits/playbook/pb_layout/_layout_mixin.scss

This is still in flux and will continue to adapt and change as we grow playbook. We may go back to class chaining but using very specific classes. I'll give you a look at the PR's as we develop other approaches! 👊

PR #294, opened 2019-09-25
Progress pills

gmfvpereira:


app/pb_kits/playbook/pb_progress_pills/_progress_pills.jsx
```diff
@@ -0,0 +1,27 @@
...
+  for(let step = 1; step <= steps; step++) {
+    items.push(<span className={step <= active ? "active" : "inactive"}></span>)
+  }
+
+  return items
+}
+
+const ProgressPills = ({ active = 0, steps = 0 }: ProgressPillsProps) => (
+  <div className="pb_progress_pills_kit">
+    {printSteps(steps, active)}
```
gmfvpereira: How can I do this loop in a better way? I don't like it, but the JSX does not accept the `{for(...)}` syntax.
I looked for some like `[1..steps].each ...` but I was not able to find it.

app/pb_kits/playbook/pb_progress_pills/_progress_pills.scss
```diff
@@ -0,0 +1,18 @@
...
+  width: 45px;
+  height: 4px;
+  display: inline-block;
+  border-radius: 4px;
+}
+
+[class^=pb_progress_pills_kit] {
+  .inactive {
+    @include pb_progress_pills_kit;
+    background-color: #E4E8F0;
```
gmfvpereira: Do we have vars for these color values?

app/pb_kits/playbook/pb_progress_pills/_progress_pills.jsx
gmfvpereira: @rafbgarcia do you have any ideas on how to make it looks better?

rafbgarcia: Does this work?

```js
const Step = ({ active, step }) => (
  <>
    <span className={classnames({ active: step <= active, inactive:  step > active }></span>)
    <span> </span>
  </>
)

{steps.map((step) => <Step step={step} active={active} />)}

Comment (rafbgarcia) [reply app/pb_kits/playbook/pb_progress_pills/_progress_pills.jsx]: Ah sorry I made a confusion on the `steps` variable type.

Comment (rafbgarcia) [reply app/pb_kits/playbook/pb_progress_pills/_progress_pills.jsx]: I found this `Array.apply(null, { length: 10 }).map(Number.call, Number)`, which returns an array from 0 to `length`

```js
Array.apply(null, { length: steps }).map(Number.call, Number).map((step) => (
  <Step 
      step={step} 
      active={active} 
  />
)
```

thestephenmarshall:
Closing PR and opening https://github.com/powerhome/playbook/issues/401

PR #301, opened 2019-09-26
PbKit simplification (plus custom errors & test suite)

xjunior:
1. Updates DSL used to define playbook kits.
1. Adds layer of validation for end users to see what data types to pass to pb_rails kit.
1. Adds test coverage to new functionality and newly ported kit.


garettarrowood:
@xjunior - I don't like my last commit, but it is the load order required to hit the byebug in the example without getting a `NameError: uninitialized constant` (and load the other kits). Feel free to overwrite it.

spec/spec_helper.rb
```diff
@@ -1,18 +1,11 @@
...
-# individual file that may not need all of that loaded. Instead, consider making
-# a separate helper file that requires the additional dependencies and performs
-# the additional setup, and require it from the spec files that actually need
-# it.
-#
-# See http://rubydoc.info/gems/rspec-core/RSpec/Core/Configuration
+# frozen_string_literal: true
+
+require_relative "../app/pb_kits/playbook/props.rb"
+Dir[File.expand_path("../app/pb_kits/playbook/props/*.rb", __dir__)].each { |f| require f }
```
xjunior: This could go right into `props.rb`, actually

garettarrowood: Moved it there in a followup commit.

---

As we add tests to kits that include the `Props` module, each spec file can require the one file they need at a time. Once all our converted, we can erase all the single line require statements from the tests, remove the `PbKit::Base` class, and add `Dir[File.expand_path("../app/pb_kits/playbook/**/*.rb", __dir__)].each { |f| require f }` into the spec helper.

garettarrowood:
```
garettarrowood: ~/Power/playbook (ruby-2.5.0) pb_rails_simplification
[Mon Sep 30 14:07:43]$ rspec

Playbook::PbCaption::Caption
  should define prop :dark of type Playbook::Props::Boolean with default false
  should define prop :large of type Playbook::Props::Boolean with default false
  should define prop :tag of type Playbook::Props::Enum with default "div"
  should define prop :text of type Playbook::Props::String with default "Caption"
  #classname
    returns namespaced class name

Playbook::Props::Base
  #validate!
    returns true when child class passes validation
    returns true when child class does not contain custom validation
    raises error when child class fails validation

Playbook::Props::Boolean
  #validate
    returns true given true
    returns true given false
    returns false given anything else

Playbook::Props::Enum
  #validate
    returns true given value supplied upon initialization
    returns false given anything else

Playbook::Props::Hash
  #validate
    returns true given a Ruby Hash
    returns false given anything else

Playbook::Props
  base props
    includes id with default of nil
    includes data with default of {}
    includes classname with default of ''
    includes aria with default of {}
    can be overwritten with custom values
      #id
      #data
      #classname
      #aria
  additional props
    can declare string prop with default
    can declare boolean string prop with default
    can declare hash string prop with default
    can declare enum string prop with default
    can be overwritten with custom values
      as string
      as boolean
      as hash
      as enum
  .props
    returns collection of available properties

Finished in 0.01314 seconds (files took 0.14728 seconds to load)
32 examples, 0 failures
```

xjunior:
This looks great, @garettarrowood ! Ready to ship!

spec/pb_kits/playbook/kits/caption_spec.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require_relative "../../../../app/pb_kits/playbook/pb_caption/caption"
+
+module Playbook
+  module PbCaption
+    describe Caption, type: :kit do
+      subject { Caption }
+
+      it { is_expected.to define_prop(:dark).of_type(Props::Boolean).with_default(false) }
```
xjunior: 👍

spec/support/playbook/rspec.rb
```diff
@@ -0,0 +1,46 @@
...
+    extend RSpec::Matchers::DSL
+
+    matcher :define_prop do |prop_key|
+      chain :of_type do |type_class|
+        @type_class = type_class
+      end
+
+      chain :with_default do |default|
+        @default = default
+      end
```
xjunior: We might want to also add a helper for testing valid enum values

xjunior:
This looks great, @garettarrowood ! Ready to ship!

spec/support/playbook/rspec.rb
garettarrowood: Yep - just wrote you the same thing in a Connect message haha.

garettarrowood: Would be pretty easy to create one for each type.

`define_string_prop(:prop)`
`define_hash_prop(:prop)`
`define_boolean_prop(:prop)`
`define_enum_prop(:prop).with_values("foo", "bar").with_default("foo")`

spec/pb_kits/playbook/props/base_spec.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module Playbook
+  module Props
+    describe Base do
+      describe "#validate!" do
+        it "returns true when child class passes validation" do
```
terryfinn: You don't want to define classes like this. A better option would have been:
```ruby
truthy_validation_class = Class.new(Playbook::Props::Base) do
  def validate(_value); true; end
end
```
This option avoid defining a constant `TruthyValidationClass`.

spec/pb_kits/playbook/props/base_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+          class TruthyValidationClass < Playbook::Props::Base
+            def validate(_value); true; end
+          end
+
+          expect(TruthyValidationClass.new.validate!(nil)).to eq true
+        end
+
+        it "returns true when child class does not contain custom validation" do
+          class NoValidationClass < Playbook::Props::Base; end
```
terryfinn: Same as above. I would also consider dropping the inheritance all together, and be more direct that the value return true.

spec/pb_kits/playbook/props/base_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+          expect(NoValidationClass.new.validate!(nil)).to eq true
+        end
+
+        it "raises error when child class fails validation" do
+          class FalsyValidationClass < Playbook::Props::Base
```
terryfinn: Same as above, don't define a new constant like this.

spec/spec_helper.rb
```diff
@@ -0,0 +1,92 @@
...
+  # Print the 10 slowest examples and example groups at the
+  # end of the spec run, to help surface which specs are running
+  # particularly slow.
+  config.profile_examples = 10
+
+  # Run specs in random order to surface order dependencies. If you find an
+  # order dependency and want to debug it, you can fix the order by providing
+  # the seed, which is printed after each run.
+  #     --seed 1234
+  config.order = :random
```
terryfinn: We should turn this on.

spec/spec_helper.rb
```diff
@@ -0,0 +1,92 @@
...
+  # the `--only-failures` and `--next-failure` CLI options. We recommend
+  # you configure your source control system to ignore this file.
+  config.example_status_persistence_file_path = "spec/examples.txt"
+
+  # Limits the available syntax to the non-monkey patched syntax that is
+  # recommended. For more details, see:
+  #   - http://rspec.info/blog/2012/06/rspecs-new-expectation-syntax/
+  #   - http://www.teaisaweso.me/blog/2013/05/27/rspecs-new-message-expectation-syntax/
+  #   - http://rspec.info/blog/2014/05/notable-changes-in-rspec-3/#zero-monkey-patching-mode
+  config.disable_monkey_patching!
```
terryfinn: We should turn this on.

spec/spec_helper.rb
```diff
@@ -0,0 +1,92 @@
...
+  # you care about by tagging them with `:focus` metadata. When nothing
+  # is tagged with `:focus`, all examples get run. RSpec also provides
+  # aliases for `it`, `describe`, and `context` that include `:focus`
+  # metadata: `fit`, `fdescribe` and `fcontext`, respectively.
+  config.filter_run_when_matching :focus
+
+  # Allows RSpec to persist some state between runs in order to support
+  # the `--only-failures` and `--next-failure` CLI options. We recommend
+  # you configure your source control system to ignore this file.
+  config.example_status_persistence_file_path = "spec/examples.txt"
```
terryfinn: This is very nice to have.

spec/spec_helper.rb
```diff
@@ -0,0 +1,92 @@
...
+  # order dependency and want to debug it, you can fix the order by providing
+  # the seed, which is printed after each run.
+  #     --seed 1234
+  config.order = :random
+
+  # Seed global randomization in this process using the `--seed` CLI option.
+  # Setting this allows you to use `--seed` to deterministically reproduce
+  # test failures related to randomization by passing the same `--seed` value
+  # as the one that triggered the failure.
+  Kernel.srand config.seed
```
terryfinn: We should turn this on with random option above.

spec/pb_kits/playbook/kits/caption_spec.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require_relative "../../../../app/pb_kits/playbook/pb_caption/caption"
+
+module Playbook
+  module PbCaption
+    describe Caption do
+      subject { Caption }
```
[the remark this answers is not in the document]
garettarrowood: It's needed for the custom rspec matchers to infer the class instead of an instance of the class.

spec/pb_kits/playbook/kits/caption_spec.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require_relative "../../../../app/pb_kits/playbook/pb_caption/caption"
+
+module Playbook
+  module PbCaption
```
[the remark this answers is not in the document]
garettarrowood: This is a rubocop choice that allows non-namespaced class name usage in the body of the spec file.

spec/pb_kits/playbook/kits/caption_spec.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require_relative "../../../../app/pb_kits/playbook/pb_caption/caption"
```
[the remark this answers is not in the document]
garettarrowood: When loading all the files, there were various load order issues because of the non-converted kits.

I think we'd have to do something like this commit otherwise. https://github.com/powerhome/playbook/pull/301/files/14d4702e4557b516652b5dd6fc25a051f4eaec76..95001d9548e56eb019c5107d2051537efaffcd6e#diff-93830fa29d616f7c87903d08b5b1b29a

I'm open to moving it into sorted directory load in the `spec_helper`. It'd be much nicer.

spec/pb_kits/playbook/props/base_spec.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module Playbook
+  module Props
```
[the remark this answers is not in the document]
garettarrowood: Same as above. This allows non-namespaced class references in the body of the file. And is the convention enforced by rubocop.

spec/pb_kits/playbook/props/base_spec.rb
garettarrowood: Cool!

app/pb_kits/playbook/props/hash.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Playbook
+  module Props
+    class Hash < Playbook::Props::Base
+      def validate(value)
+        value.is_a?(::Hash)
```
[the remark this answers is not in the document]
garettarrowood: Perhaps a `responds_to?` assertion then?

xjunior:
Hello, @terryfinn . Thank you for taking the time to comment on this PR. I have some questions:

> I have concerns about being overly focused on the type and being overly specific of a class inheritance.
Can you elaborate?

> Avoid explicit subject, it is a good way of authoring obscure tests.
What do you mean?

> Avoid wrapping specs in modules. It provide very very minimal ergonomic upfront, and can lead to odd errors and can hurt the TDD workflow.

I agree. Will push a PR later fixing this.

app/pb_kits/playbook/props.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+require "active_support/concern"
+Dir[File.expand_path("./props/*.rb", __dir__)].each { |f| require f }
```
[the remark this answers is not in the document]
xjunior: I understand the concern, but as in this case it doesn't matter, I think it's fine adding this extra calculation.

app/pb_kits/playbook/props/hash.rb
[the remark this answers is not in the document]
xjunior: For this case I rather keep it as a hash. I get the concern of being overly aggressive, but knowing it is a hash is important as we know everything a hash does can be used, such as splatting and enumerable operations. They can be important tools for the developer expecting a hash prop. Also, an object that wants to integrate into playbook can implement its own `to_h`.

spec/pb_kits/playbook/kits/caption_spec.rb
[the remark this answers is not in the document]
garettarrowood: Looks like rubocop doesn't have an issue in this repo, so I'm going with your suggestion and opening up a PR to swap all the existing test files to that format.

PR #312, opened 2019-09-27
Text Input kit

gregblake:


gregblake:
I'm creating this PR as WIP. So far, it adds a kit to accept block content. 

There is cross over between this kit and the `Input` kit. Do we want to combine these two kits, or remove one of them?

PR #313, opened 2019-09-27
Created Selectable Icon

christinaatai:


app/pb_kits/playbook/pb_selectable_icon/selectable_icon.rb
```diff
@@ -0,0 +1,102 @@
+# frozen_string_literal: true
+
+module Playbook
```
xjunior: Shouldn't we move to the new model since this ins't merged yet?

app/pb_kits/playbook/pb_selectable_icon/_selectable_icon.html.erb
```diff
@@ -0,0 +1,10 @@
+<%= content_tag(:div,
+    id: object.id,
+    data: object.data,
+    class: object.classname(object.kit_class)) do %>
```
xjunior: `classname` could be overriden in the view model

app/pb_kits/playbook/pb_selectable_icon/selectable_icon.rb
jasoncypret: I believe Brave and The Bold is handling those migrations independently.

PR #318, opened 2019-10-01
Migrate Checkbox Kit

garettarrowood:
1. Migrate Checkbox Kit to new format
1. Add a checked example to the docs.
1. Adjust base classname default (see change in test expectation in `caption_spec.rb`).
1. Add more custom matchers.
1. Add `partial` helper method.
1. Add `generate_classname` helper method.

### DIFF
--- a/app/pb_kits/playbook/pb_checkbox/checkbox.rb
+++ b/app/pb_kits/playbook/pb_checkbox/checkbox.rb
@@ -10,9 +10,9 @@ module Playbook
       prop :dark, type: Playbook::Props::Boolean, default: false
       prop :checked, type: Playbook::Props::Boolean, default: false
       prop :icon, type: Playbook::Props::Boolean, default: false
-      prop :text, default: ""
-      prop :value, default: ""
-      prop :name, default: ""
+      prop :text
+      prop :value
+      prop :name
 
       def checked_html
         checked ? "checked='true'" : nil
--- a/app/pb_kits/playbook/pb_checkbox/checkbox.rb
+++ b/app/pb_kits/playbook/pb_checkbox/checkbox.rb
@@ -5,16 +5,14 @@ module Playbook
     class Checkbox
       include Playbook::Props
 
+      partial "pb_checkbox/checkbox"
+
       prop :dark, type: Playbook::Props::Boolean, default: false
       prop :checked, type: Playbook::Props::Boolean, default: false
       prop :icon, type: Playbook::Props::Boolean, default: false
       prop :text, default: ""
       prop :value, default: ""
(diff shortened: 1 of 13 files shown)


app/pb_kits/playbook/pb_checkbox/checkbox.rb
```diff
@@ -2,93 +2,51 @@
...
+      prop :checked, default: nil
+      prop :icon, default: nil
 
       def to_partial_path
         "pb_checkbox/checkbox"
       end
 
-      def icon
-        if is_set? configured_icon
+      def icon_html
```
xjunior: I don't like this much. `ApplicationController` is a constant coming from the outer world, that might not even be called just that (in our componentized world it would always have some namespace, and although this would work we're basically relying on anything this constant matches).

Maybe this method could just be `icon_props` and in the view you could have a `pb_rails("icon", props: object.icon_props)` call.

app/pb_kits/playbook/props.rb
```diff
@@ -21,7 +21,7 @@ def prop(name)
     included do
       prop :id, default: nil
       prop :data, type: Playbook::Props::Hash, default: {}
-      prop :classname, default: ""
+      prop :classname, default: nil
```
xjunior: do we need to default to nil? Maybe this should be set in the props DSL

app/pb_kits/playbook/pb_checkbox/checkbox.rb
garettarrowood: That sounds good! I'll give it a shot.

app/pb_kits/playbook/props.rb
garettarrowood: Makes sense. I'll update the DSL, add a test assertion in the appropriate spot, and remove unneeded default declarations.

app/pb_kits/playbook/pb_checkbox/checkbox.rb
```diff
@@ -2,93 +2,42 @@
...
-
-      def text
-        default_value(configured_text, "")
-      end
+      prop :dark, type: Playbook::Props::Boolean, default: false
+      prop :checked, type: Playbook::Props::Boolean, default: false
+      prop :icon, type: Playbook::Props::Boolean, default: false
+      prop :text, default: ""
+      prop :value, default: ""
+      prop :name, default: ""
```
xjunior: Are these `default` needed?

garettarrowood: Looking at the template, I think no. I will remove them.

PR #322, opened 2019-10-06
Migrate card kit

garettarrowood:
1. Migrate `card` kit and `card_body` subkit.
1. Extend base initialization of kit to capture a `block` if yielded one.
1. Add `#with_values` matcher for `Enum` prop assertions.

### DIFF
--- a/app/pb_kits/playbook/pb_card/_card.html.erb
+++ b/app/pb_kits/playbook/pb_card/_card.html.erb
@@ -1,7 +1,7 @@
 <%= content_tag(:div,
     id: object.id,
     data: object.data,
-    class: object.classname(object.kit_class),
+    class: object.classname,
     aria: object.aria) do %>
   <%= object.yield(context: self) %>
 <% end %>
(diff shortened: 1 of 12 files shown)


app/pb_kits/playbook/pb_card/_card.html.erb
```diff
@@ -1,7 +1,9 @@
 <%= content_tag(:div,
     id: object.id,
     data: object.data,
-    class: object.classname(object.kit_class),
+    class: object.classname,
     aria: object.aria) do %>
-  <%= object.yield(context: self) %>
+  <% if object.block %>
+    <%= pb_rails("card/card_body", props: { padding: object.padding}) { capture(&object.block) } %>
```
xjunior: I think we could make `Block` a proptype so it can be passed either as a prop `block: object.block` or as a block, so we avoid reimplementing the capture. Have you tried?

garettarrowood: No. I'll give it shot.

garettarrowood: @xjunior - Conceptually, I get this. But implementation wise, I'm stabbing in the dark. Care to add this change yourself, or perhaps we could pair on Monday?

PR #325, opened 2019-10-07
Modernize badge kit

garettarrowood:
Live code demo on how to convert a kit into new format.


app/pb_kits/playbook/pb_badge/docs/_badge_colors.html.erb
```diff
@@ -91,7 +91,6 @@
 <div>
   <%= pb_rails("badge", props: {
     text: "+1",
-    variant: "neutral",
```
garettarrowood: Why delete this? To expose that it is the default and thus does not need to be declared.

PR #326, opened 2019-10-07
Children prop

xjunior:


app/pb_kits/playbook/pb_card/_card.html.erb
```diff
@@ -3,7 +3,7 @@
     data: object.data,
     class: object.classname,
     aria: object.aria) do %>
-  <% if object.block %>
-    <%= pb_rails("card/card_body", props: { padding: object.padding}) { capture(&object.block) } %>
+  <% if object.children %>
+    <%= pb_rails("card/card_body", props: { padding: object.padding, children: object.children }) %>
```
garettarrowood: I'm not sure about calling this `children`. I understand this is what React may call it. But children represents something that is plural, and a block may represent something that is plural or singular. It makes it a little confusing. And since it is a block, and we are writing Ruby, I'd prefer to leave it named as such. We wouldn't write `block` in javascript code, because it would be confusing in that context.

spec/pb_kits/playbook/props_spec.rb
```diff
@@ -35,6 +35,22 @@ class BasePropsClass
...
+
+          expect(kit.children.call).to eql 42
+        end
+
+        it "allows to be passed as a block" do
+          kit = BasePropsClass.new({}) { 42 }
+
+          expect(kit.children.call).to eql 42
+        end
+      end
```
garettarrowood: 👍

app/helpers/playbook/pb_kit_helper.rb
```diff
@@ -7,47 +7,25 @@
...
-
-    def get_class_name(kit)
-      folder = is_subkit?(kit) ? pb_camelize(kit.split("/")[0]) : pb_camelize(kit)
-      item = is_subkit?(kit) ? pb_camelize(kit.split("/")[-1]) : pb_camelize(kit)
-      "Playbook::Pb#{folder}::#{item}".safe_constantize
+    def build_view_model(kit, props, &block)
+      folder = is_subkit?(kit) ? kit.split("/")[0] : kit
+      item = is_subkit?(kit) ? kit.split("/")[-1] : kit
+      class_name = "Playbook::Pb#{folder.camelize}::#{item.camelize}"
+      class_name.safe_constantize.new(props, &block)
```
garettarrowood: Good refactoring!

app/helpers/playbook/pb_kit_helper.rb
```diff
@@ -7,47 +7,25 @@
...
 
   private
 
-    def render_rails(kit, props, &block)
-      props = defined?(props) && !props.nil? ? props : {}
-      kit_class_obj = get_class_name(kit)
-      original_value = ActionView::Base.prefix_partial_path_with_controller_namespace
-      ActionView::Base.prefix_partial_path_with_controller_namespace = true
-      render(partial: kit_class_obj.new(**props, &block), as: :object)
-    rescue ActionView::MissingTemplate
```
garettarrowood: Would be great to see this all go. You'll just have to absolutely verify both parent and child partials render properly in `nitro-web`.

Issue #331, opened 2019-10-09
Currency Kit should format number


jasoncypret:
![image](https://user-images.githubusercontent.com/863031/66500571-622b4c00-ea87-11e9-8426-d9ca4e4994d6.png)

1. Dollar sign should probably be renamed to currency indicator (we might go international). Also, it should be the default.
2. The unit and period should be reworked.
    - Either pass a value with decimals and it splits it into the unit
    - Or when the unit is present add the period automatically
3. The comma should be added/formatted automatically
4. The React kit is not using the correct color for the unit.

jasperfurniss:
My thought here is to utilize the power of an ActionView helper. 
https://api.rubyonrails.org/classes/ActionView/Helpers/NumberHelper.html#method-i-number_to_currency

I noticed we started doing that in some of our rails kits. 
Ex. phone.rb, we use the number_to_phone helper and it works perfect!
```
      def number
        if is_set? configured_number
          return number_to_phone(formatted_number, area_code: true) if formatted_number
        end
      end
```

This is the direction i'm headed down right now.
Wanted to check with you all before I go that route to see if there were any tips/thoughts.
@jasoncypret @thestephenmarshall @markiearnold @garettarrowood

jasperfurniss:
It uses locale for setting the currency type, and by default, it uses the current locale (which in our case would mean $$$ dollar symbols).

PR #335, opened 2019-10-09
Currency formatting

jasperfurniss:
### 4 Fixes for issue #331

Implemented a new ActionView helper called number_to_currency for the rails kit. 
Also fixed the colors on the react kit. 
Markup is simpler for calling the rails kit now. 
Screen Shot 2019-10-09 at 3 31 13 PM
Screen Shot 2019-10-09 at 3 44 16 PM


app/pb_kits/playbook/pb_currency/currency.rb
```diff
@@ -61,20 +55,32 @@ def value_size
         size == "lg" ? 1 : 2
       end
 
+      def formatted_value
```
garettarrowood: This method, and perhaps more, could be made private if they are not getting used in the html.erb file.

app/pb_kits/playbook/pb_currency/currency.rb
```diff
@@ -61,20 +55,32 @@ def value_size
...
+        number_to_currency(configured_value)
+      end
+
       def value
         if is_set? configured_value
-          pb_title = Playbook::PbTitle::Title.new(size: value_size, text: configured_value, classname: "pb_currency_value")
+          display_value = formatted_value[1...-2]
+          pb_title = Playbook::PbTitle::Title.new(size: value_size, text: display_value, classname: "pb_currency_value")
           ApplicationController.renderer.render(partial: pb_title, as: :object)
```
garettarrowood: As we've been moving over kits to the new format, we've been removing all `ApplicationController.renderer.render` invocations in preference to simply using `pb_rails` in the view partial. You're fine to keep this here, but wanted to give you a heads up. -- I guess as your team enhances more kits, we should try to move them into the view file for consistency.

app/pb_kits/playbook/pb_currency/currency.rb
```diff
@@ -61,20 +55,32 @@ def value_size
         size == "lg" ? 1 : 2
       end
 
+      def formatted_value
+        number_to_currency(configured_value)
```
garettarrowood: Can `value` have a default?  Does this rely on `value` being set?

app/pb_kits/playbook/pb_currency/currency.rb
```diff
@@ -3,14 +3,15 @@
 module Playbook
   module PbCurrency
     class Currency < Playbook::PbKit::Base
+      include ActionView::Helpers::NumberHelper
       PROPS = %i[configured_align
                  configured_classname
                  configured_data
                  configured_id
                  configured_label
                  configured_value
```
garettarrowood: Instead of calling this `value`, can we call this `amount`? Value seems pretty generic.

app/pb_kits/playbook/pb_currency/currency.rb
jasperfurniss: Currently, this relies on value being set. Value cannot have a default because we wouldn't want to display something that wasn't explicitly set right? At least for this kit, I think that makes sense. What do you think?

garettarrowood: I agree. --- I will shortly be working on a PR that adds some functionality to give us the ability to make a prop required (in the new kit definition format though). When we loop back to `currency`, we can make this a required prop.

app/pb_kits/playbook/pb_currency/currency.rb
```diff
@@ -2,99 +2,57 @@
...
+                  default: "sm"
+      prop :unit, default: "$"
 
-      def dollar_sign
-        pb_dollar_sign = Playbook::PbBody::Body.new(classname: "dollar_sign", color: "light") do
-          default_value(configured_dollar_sign, "")
-        end
-        ApplicationController.renderer.render(partial: pb_dollar_sign, as: :object)
+      def classname
+        generate_classname("pb_currency_kit")
```
garettarrowood: looks like this needs to be passed `align` as well to match the last implementation.

app/pb_kits/playbook/pb_currency/_currency.html.erb
```diff
@@ -1,13 +1,14 @@
...
-
+    class: object.classname,
+    aria: object.aria) do %>
+  
   <%= object.label %>
   <div class="pb_currency_wrapper">
-    <%= object.dollar_sign %>
-    <%= object.value %>
-    <%= object.unit %>
+    <%= pb_rails("body", props: { color: "light", text: object.unit }) %>
```
garettarrowood: Looks like the `body` before was passed a block. You could do that here in the view. OR, you could use the `children` prop that @xjunior added in #326 and pass it inline. --- If this already looks how it needs too, disregard this comment :).

app/pb_kits/playbook/pb_currency/currency.rb
jasperfurniss: @garettarrowood  For some reason, my git ammend wasn't working correctly. You should be seeing the align_class private method I added now.

app/pb_kits/playbook/pb_currency/currency.rb
```diff
@@ -2,99 +2,53 @@
...
-          "pb_currency_kit",
-          align,
-        ]
-        kit_options.join("_")
-      end
+    private
 
-      def to_partial_path
-        "pb_currency/currency"
+      def align_class
```
garettarrowood: Since `align` can already only be these 3 strings, you are good to drop this method and just use `align` in `classname` directly.

app/pb_kits/playbook/pb_currency/_currency.scss
```diff
@@ -26,7 +26,7 @@
   [class^=pb_currency_wrapper] {
     display: flex;
 
-    & > [class*=dollar_sign] {
+    & > [class*=currency_indicator] {
```
garettarrowood: At first inspection, it doesn't look like this classname is getting included in the rails kit anymore.

jasperfurniss: Good catch. Updated.

app/pb_kits/playbook/pb_currency/currency.rb
jasperfurniss: Awesome. Get to delete more code. Woohoo. 🎉

app/pb_kits/playbook/pb_currency/_currency.jsx
```diff
@@ -1,72 +1,75 @@
...
+  let css = "pb_currency_kit";
+  css += `_${align}`;
+  return css;
+};
 
-const bodyCSS = ({}: CurrencyProps) => {
-  let css = ''
-  return css
-}
+const bodyCSS = ({  }: CurrencyProps) => {
```
markiearnold: This isn't being used, it can be removed (`bodyCSS`)

app/pb_kits/playbook/pb_currency/_currency.jsx
```diff
@@ -1,72 +1,75 @@
...
-  Body,
-  Caption,
-  Title,
-} from '../'
+import { Body, Caption, Title } from "../";
 
 type CurrencyProps = {
-  align?: 'left' | 'center' | 'right',
+  align?: "left" | "center" | "right",
   className: String,
```
markiearnold: className should be optional (`className?: String`)

app/pb_kits/playbook/pb_currency/_currency.jsx
```diff
@@ -1,72 +1,75 @@
...
-  currencySymbol: '$',
+  currencySymbol: "$",
```
markiearnold: I believe this should be updated to type String.
And should it be changed to optional? (`currentSymbol?: String`)
You are setting the default value to "$" below.

app/pb_kits/playbook/pb_currency/_currency.jsx
```diff
@@ -1,72 +1,75 @@
...
-    separator='.',
-    size='sm',
-    units='00',
-    value,
-  } = props
+    currencySymbol = "$",
+    label = "",
+    separator = ".",
+    size = "sm",
+    units = "00",
```
markiearnold: The prop name is `unit`, this should be consistent with prop name. (units)

app/pb_kits/playbook/pb_currency/_currency.jsx
```diff
@@ -1,72 +1,75 @@
...
+        <Title className="pb_currency_value"
             size={sizes[size]}
         >
           {`${value}${separator}`}
         </Title>
-        <Body className='unit'>{units}</Body>
+        <Body className="cents"
+            color={"light"}
+        >
+          {units}
```
markiearnold: See comment above. Prop name is `unit`, this should be consistent with prop name. (units)

app/pb_kits/playbook/pb_currency/_currency.jsx
```diff
@@ -1,72 +1,75 @@
...
+  currencySymbol: "$",
   label?: String,
-  separator?: '.' | ',',
-  size?: 'sm' | 'lg',
+  separator?: "." | ",",
+  size?: "sm" | "lg",
   value?: String,
-  unit?: '00',
-}
+  unit?: "00"
```
markiearnold: Same as above for `currentSymbol`. I think the type here should be String (`unit?: String`)

app/pb_kits/playbook/pb_currency/currency.rb
```diff
@@ -1,100 +1,46 @@
...
-      end
+      prop :align, type: Playbook::Props::Enum,
+                   values: %w[left center right],
+                   default: "left"
+      prop :amount, required: true
+      prop :label
+      prop :size, type: Playbook::Props::Enum,
+                  values: %w[lg sm],
+                  default: "sm"
+      prop :unit, default: "$"
```
markiearnold: The rails and react kits are working differently.

The react kit has 2 props: 
* `currencySymbol` which determines beginning currency symbol ($)
* `unit` which determines the smaller end amount (00)

---

The rails kit has 1 prop of these props
* `unit` which determines the beginning currency symbol ($)
And the smaller end amount is auto-generated without an option to customize.

The unit prop is doing completely different things between the kits. I think these kits should behave the same way.

app/pb_kits/playbook/pb_currency/_currency.jsx
```diff
@@ -1,72 +1,75 @@
...
+};
 
-const kitCSS = ({align='left'}: CurrencyProps) => {
-  let css = 'pb_currency_kit'
-  css += `_${align}`
-  return css
-}
+const kitCSS = ({ align = "left" }: CurrencyProps) => {
```
drborges: This [props utility file](https://github.com/powerhome/playbook/blob/master/app/pb_kits/playbook/utilities/props.js#L48) was introduced a few weeks ago with a handful of useful helpers that you may be able to leverage here to dry up a bit the code.

The [pb_toggle kit](https://github.com/powerhome/playbook/blob/89d83abe394e9924787bf079dfef4eeff0fb2987/app/pb_kits/playbook/pb_toggle/_toggle.jsx#L45) makes use of these helpers and should be a good reference.

PR #336, opened 2019-10-09
Modernize Input

KFad11:
Brings Input pb_kit up to our new standard convention with test spec

BAB 155

### DIFF
--- a/app/pb_kits/playbook/pb_input/input.rb
+++ b/app/pb_kits/playbook/pb_input/input.rb
@@ -7,56 +7,11 @@ module Playbook
 
       partial "pb_input/input"
 
-        PROPS = %i[
-                   configured_name
-                   configured_label
-                   configured_placeholder
-                   configured_type
-                   configured_value].freeze
-
-        def label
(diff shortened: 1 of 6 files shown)


garettarrowood:
Is this still WIP?

app/pb_kits/playbook/pb_input/input.rb
```diff
@@ -2,89 +2,20 @@
...
+      prop :placeholder
+      prop :value
+      prop :type, default: "text"
 
-      DEFAULT = Object.new
-      private_constant :DEFAULT
-      def default_configuration
-        DEFAULT
+      def classname
+        generate_classname("pb_input_field_kit")
```
garettarrowood: Before, it looks like this was just `input_field` and not `"pb_input_field_kit"`. Are you sure that this change will not break any other kits that depend on it? Is there any custom css for this kit?

KFad11: Got it updated in the latest commit

PR #337, opened 2019-10-09
Revert "Children prop"

garettarrowood:
Reverts powerhome/playbook#326

http://playbook.powerhrg.com/ is not loading since this PR merged. Testing locally, opening up this revert in the meantime.

PR #338, opened 2019-10-09
Revert "Revert "Children prop""

xjunior:
Reverts powerhome/playbook#337


app/helpers/playbook/pb_doc_helper.rb
```diff
@@ -60,9 +60,9 @@ def get_kit_examples(kit, type)
     end
```
garettarrowood: While you are editing this file, could you get rid of the unneeded `return`s as well?

xjunior:
Both tested locally and working. Anyway, we should cut a branch in nitro and test the update of playbook before get it to nitro#master

PR #350, opened 2019-10-11
Date Range Stacked Kit

christinaatai:
Created rails and react versions


app/pb_kits/playbook/pb_date_range_stacked/date_range_stacked.rb
```diff
@@ -0,0 +1,88 @@
...
+      end
+
+      def start_date
+        Playbook::PbKit::PbDateTime.new(configured_start_date)
+      end
+
+      def start_day_month
+        day_month = "#{start_date.to_day} #{start_date.to_month}"
+        pb_title = Playbook::PbTitle::Title.new(text: day_month, size: 4, dark: dark)
+        ApplicationController.renderer.render(partial: pb_title, as: :object)
```
garettarrowood: Besides moving rails kits to the "modernized" format, we are also removing all direct invocations of `ActionController`. There's no need to update this kit to the more recent convention, but it would be good if you could move this into the `html.erb` template itself. So instead of:

```html
<div class="pb_date_range_stacked_start_date_wrapper">
  <%= object.start_day_month %>
  <%= object.start_year %>
</div>
```

it'd be 

```html
<div class="pb_date_range_stacked_start_date_wrapper">
  <%= pb_rails("title", props: { text: object.day_month, size: 4, dark: object.dark }) %>
  <%= pb_rails("body", props: { text: object.start_year, color: light, dark: object.dark }) %>
</div>
```
and in your object file, these methods would be removed and replaced with:

```ruby
def day_month
  "#{start_date.to_day} #{start_date.to_month}"
end

def start_year
  start_date.to_year
end
```

garettarrowood: It also makes what's happening in the template more obvious.

app/pb_kits/playbook/pb_date_range_stacked/date_range_stacked.rb
```diff
@@ -0,0 +1,88 @@
...
+      def end_day_month
+        day_month = "#{end_date.to_day} #{end_date.to_month}"
+        pb_title = Playbook::PbTitle::Title.new(text: day_month, size: 4, dark: dark)
+        ApplicationController.renderer.render(partial: pb_title, as: :object)
+      end
+
+      def end_year
+        year = end_date.to_year
+        pb_body = Playbook::PbBody::Body.new(text: year, color: "light", dark: dark)
+        ApplicationController.renderer.render(partial: pb_body, as: :object)
```
garettarrowood: Same goes for the ones down here.

xjunior:
Can you post a screenshot of this new component?

thestephenmarshall:
Agree - @christinaatai let's start a new branch off master and re-run the kit generator to get the modernized format. You can reference this PR for your specific changes and simply transfer those over. The most significant changes that shouldn't be directly transferred are in the `.rb` files.

PR #356, opened 2019-10-14
Bab/modernize bar graph kit

leilaweiss:
Modernize Bar Graph Kit and add Spec file
Also add a Array type


garettarrowood:
Don't forget to squash and merge

PR #363, opened 2019-10-15
Modernize Title Playbook Kit

darrendc:
BAB-171
Modernize Title Playbook kit.

### DIFF
--- a/app/pb_kits/playbook/pb_title/title.rb
+++ b/app/pb_kits/playbook/pb_title/title.rb
@@ -10,8 +10,8 @@ module Playbook
       prop :dark, type: Playbook::Props::Boolean,
                   default: false
       prop :size, type: Playbook::Props::Enum,
-                  values: %w[1 2 3 4],
-                  default: "3"
+                  values: [1, 2, 3, 4],
+                  default: 3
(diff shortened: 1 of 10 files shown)


app/pb_kits/playbook/pb_title/title.rb
```diff
@@ -2,77 +2,30 @@
...
-        adjusted_value(size, size, nil)
-      end
-
-      def dark_class
-        true_value(configured_dark, "dark", nil)
-      end
+      prop :dark, type: Playbook::Props::Boolean,
+                  default: false
+      prop :size, type: Playbook::Props::Enum,
+                  values: %w[1 2 3 4],
```
garettarrowood: We definitely want to get this up in a local server before merging. `size` looks to accept integers 1-4, but we are declaring that it be strings "1"-"4". I'm wondering if the examples in the docs blow up...

xjunior: Just remove `%w` here and separate the values with `,` and it should be fine.

darrendc: Resolved. The fix required changing several .slim files and a helper method use of pb_title's size value from a string to an integer.

PR #364, opened 2019-10-15
Modernize ProgressSimple Playbook kit

darrendc:
BAB-162
Modernize ProgressSimple Playbook kit

### DIFF
--- a/app/pb_kits/playbook/pb_progress_simple/progress_simple.rb
+++ b/app/pb_kits/playbook/pb_progress_simple/progress_simple.rb
@@ -49,7 +49,7 @@ module Playbook
       end
 
       def wrapper_classname
-        "pb__progress_simple_wrapper_#{align}"
+        "pb_progress_simple_wrapper_#{align}"
       end
 
       def classname
--- a/app/pb_kits/playbook/pb_progress_simple/progress_simple.rb
+++ b/app/pb_kits/playbook/pb_progress_simple/progress_simple.rb
@@ -14,8 +14,11 @@ module Playbook
       prop :max, type: Playbook::Props::Number
       prop :muted, type: Playbook::Props::Boolean,
            default: false
-      prop :percent, type: Playbook::Props::Percentage # Float type
-      prop :width, default: "100%" # Create type that validates only valid css measurements
+      prop :percent, type: Playbook::Props::Percentage
+      # :width prop should not probably be a string type
+      # Should we be allowing the user to pass value at all?
+      # could this possibly be [sm, md, lg]?
+      prop :width, default: "100%"
 
       def number_value
         if percent
--- a/app/pb_kits/playbook/pb_progress_simple/progress_simple.rb
+++ b/app/pb_kits/playbook/pb_progress_simple/progress_simple.rb
@@ -2,99 +2,59 @@
 
 module Playbook
   module PbProgressSimple
-    class ProgressSimple < Playbook::PbKit::Base
-      PROPS = %i[configured_align
-                 configured_classname
-                 configured_data
-                 configured_id
-                 configured_max
-                 configured_muted
-                 configured_percent
-                 configured_value
-                 configured_width].freeze
+    class ProgressSimple
+      include Playbook::Props
 
-      def initialize(align: default_configuration,
-                     classname: default_configuration,
-                     data: default_configuration,
-                     id: default_configuration,
-                     max: default_configuration,
-                     muted: default_configuration,
-                     percent: default_configuration,
-                     value: default_configuration,
-                     width: default_configuration)
-        self.configured_align = align
-        self.configured_classname = classname
-        self.configured_data = data
-        self.configured_id = id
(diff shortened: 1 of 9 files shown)


app/pb_kits/playbook/pb_progress_simple/progress_simple.rb
```diff
@@ -2,99 +2,97 @@
...
+           default: "left"
+      prop :value, type: Playbook::Props::Number
+      prop :max, type: Playbook::Props::Number
+      prop :muted, type: Playbook::Props::Boolean,
+           default: false
+      prop :percent, type: Playbook::Props::Percentage
+      # :width prop should not probably be a string type
+      # Should we be allowing the user to pass this value at all?
+      # could this possibly be [sm, md, lg]?
+      prop :width, default: "100%"
```
thestephenmarshall: Width should be an integer from 0-100 if the plan is for it to be this configurable.. but my question is why do we even need this prop at all? Shouldn't `percent` prop take care of this?

app/pb_kits/playbook/pb_progress_simple/progress_simple.rb
```diff
@@ -2,99 +2,97 @@
...
-        ]
-        kit_options.compact.join("_")
+      def value_style
+        "width:#{number_value}%;"
       end
 
-      def to_partial_path
-        "pb_progress_simple/progress_simple"
+      def wrapper_classname
+        "pb__progress_simple_wrapper_#{align}"
```
thestephenmarshall: Weird, but it appears you have **two** underscores `_` between `pb` and `progress`. Was this intentional?

app/pb_kits/playbook/pb_progress_simple/progress_simple.rb
```diff
@@ -2,99 +2,97 @@
...
+      def classname
+        generate_classname("pb_progress_simple_kit", muted_class, align)
       end
 
-      def format_percent(num)
-        "#{num}%"
+    private
+
+      def muted_class
+        muted ? "muted" : nil
```
thestephenmarshall: Shouldn't we just be able to rely on `muted` without the ternary?

app/pb_kits/playbook/props/percentage.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Playbook
+  module Props
+    class Percentage < Playbook::Props::Base
+      def validate(value)
+        value.nil? ||
+          (value.is_a?(Numeric) &&
+          value <= 100 &&
+          value >= 0)
```
thestephenmarshall: Would it make sense to refactor this to:

```rb
...
value.nil? || numeric_in_range?(value)
...
def numeric_in_range?(value)
  (value.is_a?(Numeric) &&
          value <= 100 &&
          value >= 0)
end
```

for readability up-top?

app/pb_kits/playbook/pb_progress_simple/progress_simple.rb
garettarrowood: This is the way we found it in the old implementation and just copied it over. We will take a closer look to see if this was intended or if we can adjust to use only one underscore.

app/pb_kits/playbook/pb_progress_simple/progress_simple.rb
garettarrowood: Since `muted` is a boolean, it would add `"true"` or `"false"` into the ` classname` if we didn't have this method to interpret a true as `"_muted"` or to just leave it out of the `classname`.

app/pb_kits/playbook/props/percentage.rb
garettarrowood: Yep! That looks good!

app/pb_kits/playbook/pb_progress_simple/progress_simple.rb
garettarrowood: The `width` prop is used here - https://github.com/powerhome/playbook/pull/364/files#diff-546b67958995ee7824144b42d48a0e79R44 - and seems to provide stylistic configuration for the size of the full progress bar container. `percent` OR the combo of `max` & `value` sets how much actual progress (blue) is in the bar.

We talked about this prop A LOT, and finally decided to leave a note instead of make an architectural change. The reason for that is we don't yet have any use cases in Nitro. One though of mine is that the developer using this kit could change the "width" of how long the progress bar is by simply putting in any sort of element with a fixed width. - Which makes my vote to deal with this lean towards removing this prop entirely.

Thoughts are welcome on this, but we don't want to get to bogged down. If we can come to a consensus, we are happy to open up another PR and address `width` specifically.

app/pb_kits/playbook/pb_progress_simple/progress_simple.rb
darrendc: It looks to be a typo from the scss file. I changed it to one underscore

PR #371, opened 2019-10-18
Modernize table kit

eabousaif:
Modernize the table playbook kit

### DIFF
--- a/app/pb_kits/playbook/pb_table/docs/_table_sm.html.erb
+++ b/app/pb_kits/playbook/pb_table/docs/_table_sm.html.erb
@@ -1,4 +1,4 @@
-<%= pb_rails("table", props: {size: "sm"}) do %>
+<%= pb_rails("table", props: { size: "sm", container: false }) do %>
   <thead>
     <tr>
       <th>Column 1</th>
--- a/app/pb_kits/playbook/pb_table/table.rb
+++ b/app/pb_kits/playbook/pb_table/table.rb
@@ -17,7 +17,7 @@ module Playbook
       prop :disable_hover, type: Playbook::Props::Boolean,
                            default: false
       prop :container, type: Playbook::Props::Boolean,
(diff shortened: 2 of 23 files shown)


app/pb_kits/playbook/pb_table/table.rb
```diff
@@ -2,94 +2,48 @@
...
+                  values: %w[sm md lg],
+                  default: "md"
+      prop :single_line, type: Playbook::Props::Boolean,
+                         default: false
+      prop :dark, type: Playbook::Props::Boolean,
+                  default: false
+      prop :disable_hover, type: Playbook::Props::Boolean,
+                           default: false
+      prop :container, type: Playbook::Props::Boolean,
+                       default: false
```
garettarrowood: Every story that uses the table kit that we bring in front of Kraig comes back to us with him wanting `container: true`. Can we set this as the default option, @jasoncypret ?

garettarrowood:
Nice work on this conversion!

app/pb_kits/playbook/pb_table/table.rb
eabousaif: As per the last conversation with Jason, the default for container has been set to true.

app/pb_kits/playbook/pb_table/docs/_table_sm.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= pb_rails("table", props: {size: "sm"}) do %>
+<%= pb_rails("table", props: { size: "sm", container: false }) do %>
```
darrendc: It's picky, but you used a space after opening and before closing the curly brackets here, but did not do that in the other .erb files.

eabousaif: Fixed the spacing in all html files, thanks!

PR #372, opened 2019-10-18
Modernize the Person Playbook Kit

darrendc:
Modernize the Person Playbook kit
BAB-192

### DIFF
--- a/app/pb_kits/playbook/pb_person/person.rb
+++ b/app/pb_kits/playbook/pb_person/person.rb
@@ -10,20 +10,6 @@ module Playbook
       prop :first_name
       prop :last_name
 
-      def first_name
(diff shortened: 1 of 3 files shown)


app/pb_kits/playbook/pb_person/person.rb
```diff
@@ -2,55 +2,17 @@
...
-    private
+      prop :first_name
+      prop :last_name
 
-      DEFAULT = Object.new
-      private_constant :DEFAULT
-      def default_configuration
-        DEFAULT
+      def classname
+        generate_classname("pb_person_kit", first_name, last_name)
```
garettarrowood: This should not be taking `first_name` and `last_name`. It should only be `generate_classname("pb_person_kit")`

PR #373, opened 2019-10-21
Show Zipcode and Add Territory to Home Address Street Playbook

kskilt:
Originally, the zipcode was an accepted prop that was not displayed anywhere in the kit. Also the territory was not an accepted prop, and the position the territory would take was filled in by the state. In Nitro there is a difference between state and territory (example: Delaware state is PHL territory) so prop added and also displayed. 

This was the original pbkit without Zip or proper territory:
Screen Shot 2019-10-21 at 9 49 34 AM

Compared to old Nitro with correct fields: 
Screen Shot 2019-10-21 at 10 00 19 AM

This is the new kit with proper values and displays:

Screen Shot 2019-10-21 at 10 24 03 AM

### DIFF
--- a/app/pb_kits/playbook/pb_home_address_street/_home_address_street.jsx
+++ b/app/pb_kits/playbook/pb_home_address_street/_home_address_street.jsx
@@ -47,7 +47,7 @@ const HomeAddressStreet = ({
       {address} {`\u00b7`} {houseStyle}
     </Title>
     <Body color="light">
-      {city}, {state}
+      {city}, {state} {zipcode}
     </Body>
     <Body
         className="home-hashtag"
@@ -59,7 +59,7 @@ const HomeAddressStreet = ({
         color="light"
         tag="span"
     >
-      <small>{state}</small>
+      <small>{territory}</small>
     </Body>
   </div>
 )
--- a/app/pb_kits/playbook/pb_home_address_street/_home_address_street.jsx
+++ b/app/pb_kits/playbook/pb_home_address_street/_home_address_street.jsx
@@ -18,6 +18,7 @@ type HomeAddressStreetProps = {
   houseStyle: String,
   state: String,
   zipcode: String,
+  territory: String,
 }
 
 const classes = (className, dark) => (
@@ -36,6 +37,7 @@ const HomeAddressStreet = ({
   houseStyle,
   state,
   zipcode,
+  territory,
 }: HomeAddressStreetProps) => (
   <div className={classes(className, dark)}>
     <Title
--- a/app/pb_kits/playbook/pb_home_address_street/_home_address_street.jsx
+++ b/app/pb_kits/playbook/pb_home_address_street/_home_address_street.jsx
@@ -9,8 +9,18 @@ import {
   Title,
 } from '../'
 
+
+  const selector = (houseStyle) =>
+    { if (houseStyle == "") {
+      return ""
+    } else {
+      "\u00b7"
+    }
+  }
+
 type HomeAddressStreetProps = {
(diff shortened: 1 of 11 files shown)


app/pb_kits/playbook/pb_home_address_street/_home_address_street.jsx
```diff
@@ -36,6 +37,7 @@ const HomeAddressStreet = ({
   houseStyle,
   state,
   zipcode,
+  territory,
```
garettarrowood: Lets add `zipcode` and `territory` to the jsx template as well to keep these consistent.

app/pb_kits/playbook/pb_home_address_street/docs/_home_address_street_dark.html.erb
```diff
@@ -1,10 +1,11 @@
 <%= pb_rails("home_address_street", props: {
   address: "70 Prospect Ave",
   address2: "Apt M18",
```
jasoncypret: Looks like this got merged last week as wasn't picked up. Can you rename this? Address2 is not descriptive enough. Let's go for address_cont or address_continued

app/pb_kits/playbook/pb_home_address_street/docs/_home_address_street_default.html.erb
```diff
@@ -1,9 +1,10 @@
 <%= pb_rails("home_address_street", props: {
   address: "70 Prospect Ave",
   address2: nil,
```
jasoncypret: Are you required to pass `address2`? It should be optional

jasoncypret:
I'm thinking we will want to make changes to the view of this kit. WIth `address_continued` added it doesn't look right. I'm looping in @markiearnold here

app/pb_kits/playbook/pb_home_address_street/docs/_home_address_street_dark.html.erb
```diff
@@ -1,10 +1,11 @@
...
-  city: "North Arlington",
+  city: "West Chester",
   home_id: 8250263,
   house_style: nil,
```
jasoncypret: You shouldn't have to pass `nil` to any of the kits

app/pb_kits/playbook/pb_home_address_street/docs/_home_address_street_dark.html.erb
kskilt: We can definitely rename it. I went with address and address2 as this is how the addresses exist in Nitro database

app/pb_kits/playbook/pb_home_address_street/docs/_home_address_street_default.html.erb
kskilt: So it is optional. I hardcoded nil, but leaving it blank would result in the same outcome -- no visible change to the kit

app/pb_kits/playbook/pb_home_address_street/docs/_home_address_street_dark.html.erb
KFad11: I like address continued, I see cont apply to "contains" as well in Nitro.  Changing that now.

app/pb_kits/playbook/pb_home_address_street/docs/_home_address_street_dark.html.erb
markiearnold: I agree with @jasoncypret above. In the documented examples, if we do not want to display home_style, it should be optional and we should not pass it in as a prop.

PR #375, opened 2019-10-21
Modernize Body Playbook kit

darrendc:
BAB-178
Modernize Body Playbook kit.


darrendc:
This PR also updates the `children` prop utilizing `pb_body` in `pb_person`

PR #376, opened 2019-10-23
Playbook rails form kit

terryfinn:
Add a form kit for rails that provides a thin wrapper around the conventional rails `form_with`. This over rides the default form builder to use one the builds the form by composing kits.

Screen Shot 2019-11-19 at 10 59 58 AM
Screen Shot 2019-11-19 at 11 00 47 AM


xjunior:
Bringing the conversation that happened in `Connect` to this PR:

@xjunior  10:26am
> @terryfinn for whole forms, wouldn't be better if we do something that integrates more with simple form and customize the fields using its plugin system and it's appearance using its options?
> I say that because we already have some simple form inputs in nitro

@terryfinn  11:44am
> @xjunior I did consider it, but I decided against a direct integration with simple form at this point. First, form_with, form_for, and simple_form_for can call take a custom form builder as an argument. This means, that this work can be applied to multiple systems. Second, I found that form_with provides maximal flexibility. It felt like the most useful integration to build at this point. I see this work at this point as being a stepping stone into a larger body of work, and this seemed like the best foundation to lay and would be the least fraught with contention.

@xjunior 11:49am
> The problems I see are: 1) it doesn't integrate well with Simple Form even with a custom form builder because the layout for each field wrapper is hardcoded in each helper (instead of a single reusable wrapper html – see nitro_search/form_builder.rb:273); 2) we already use simple form as a standard to produce reusable fields (see nitro_theme/app/inputs) which will be lost if we transition into a new system
> I feel like this kit is useful, but would do better if we integrate with simple form
> If we go the simple form way we can create input plugins that will integrate not just within this form component, but also in search forms. Also, it makes this component a form only component, giving it a single responsibility instead of the responsibility for the form and all the input types we wish to implement

wadewinningham:
Would definitely need to incorporate `simple_form`. As far as I know, it still wraps some fields in extra elements to enforce styling at at least maintains a single location for making global style changes even if the end results look the same.

@terryfinn I'm not overly familiar with `form_with` -- do you think we could remove `simple_form` and be able to switch over to that? I'd be more interested if we could reduce dependencies. What would we lose with such a switch?

terryfinn:
My  thought is this. We are not going to be able to bang out a form system  tomorrow. In fact we don't even yet have that many form input type kits  built. The once we do have need some extension to support work with build forms. Simple form will need a revised form builder; one that uses kits for foundation inputs. Also, we  would need to change other parts of simple form to support kits.

This PR is still a work in progress providing a starting point; giving forms a place to live in playbook. It starts the process of  building a new form builder and updating form inputs kits to work with it.  The ramp to rolling this out, or updating simple form seems long, and this is progress towards that.

> @terryfinn I'm not overly familiar with form_with -- do you think we could remove simple_form  and be able to switch over to that? I'd be more interested if we could  reduce dependencies. What would we lose with such a switch?

@wadewinningham  I don't know if form_with would be a replacement for us. simple_form definitely has conveniences that don't exist out of the box in form_with: like inferring a input type form the field  type on the model. I don't know how vital those conveniences are.  form_with also has some conveniences over simple_form, like not needing  to be tied to an active record to active model object, but again, I  don't know how important those conveniences are.

I can tell you I like form_with. It was part of Rails 5.1 and started off as a propose form DHH. It provides unity around form_tag and form_for. You can read by documentation on form with here.

xjunior:
@terryfinn my point is that we could have a form kit, but based on Simple Form, already taking into consideration kits for building forms, but continue to support any simple form input.

xjunior:
Form with does not replace SimpleForm AFAIK as SimpleForm uses the rails foundation to build forms and provides the abstractions that we're looking for (custom inputs, unified wrapper)

bmwest:
> @wadewinningham I don't know if form_with would be a replacement for us. simple_form definitely has conveniences that don't exist out of the box in form_with: like inferring a input type form the field type on the model. I don't know how vital those conveniences are. form_with also has some conveniences over simple_form, like not needing to be tied to an active record to active model object, but again, I don't know how important those conveniences are.

This might not be something we even need to think about right now, but do we know of any features in Nitro that may rely on some of the conveniences of `simple_form`? For example, does NitroSearch? 

I like the idea of continuing with `form_with`, I'm mostly just curious about how the conveniences we might loose if we choose to switch to `form_with` would need to be accounted for. Might `form_with` enhance something like NitroSearch in any way?

wadewinningham:
I'm interested in @thestephenmarshall 's take once he's back from vacation.

thestephenmarshall:
Great conversation here! I share the same concerns as @wadewinningham and @bmwest here. I like the general direction that this PR is headed however, but agree with @terryfinn that this is really just the starting WIP. It would be worth well worth our time to open up a branch in Nitro and start playing with this particular branch there within a more complex built form to see where the pitfalls really are with regarding to `form_with` vs. `simple_form`. 

As you might expect, the other nut to crack here is creating this on the React side of the house. Questions arise as to how we might map form fields to props in a maintainable way from Rails as well as state management. Some sideline discussion has occurred on this very topic, but it has been scarce so perhaps this is the right PR for those discussions to live.

It is worth noting that PB Input is looking to be a bit too generically named as well, but know that this is something that @powerhome/ux-team is looking into making that much more meaningful.

wadewinningham:
@terryfinn By building a custom form builder, which of these do you mean? 

- Writing a complete *replacement* for SimpleForm
- A new [wrapper](https://github.com/plataformatec/simple_form#the-wrappers-api) in SimpleForm
- A custom class based on [`SimpleForm::FormBuilder`](https://rubydoc.info/github/plataformatec/simple_form/master/SimpleForm/FormBuilder)

The first seems like a heavy lift. I was originally thinking you were talking about the 2nd one above (a new wrapper) but since SimpleForm does have a FormBuilder class, it's a little confusing, and I want to be sure I understand.

I'm for whatever the best path forward for Playbook is, so I'd lean on @thestephenmarshall 's thoughts as I'm not as familiar as either one of you on the internals of Playbook (something I need to rectify).

terryfinn:
@wadewinningham a custom [`FormBuilder`](https://api.rubyonrails.org/classes/ActionView/Helpers/FormBuilder.html). `SimpleForm::FormBuilder` is a `ActionView::Helpers::FormBuilder` instance.

xjunior:
My concern is that we'll have to extend this form with inputs very specific to nitro, and therefore that shouldn't be in this repo. Examples are a territory select, or a user title select. Simple Form is nothing but an extension of rails' forms with an interface for input plugins. I think that's desireable.

wadewinningham:
We plan to open source Playbook, so it would be important to not include Nitro-specific things into it.

terryfinn:
So, what are people thoughts on the best path forward?

It seems that the group has a strong opinion to start off targeting simple form. It also seem to group has concerns about Nitro specific code staying out of this project. To address the later concern, we could create a playbook-nitro component to be the place where Nitro specific kits live.

We could modify the existing input kit to be more usable in a form building system, target simple form and ship that change. When it comes time to the future work of building a user input, we could either: build that in playbook or build it in nitro, but I don't think we need to decide that to start making progress again.

xjunior:
In summary, @terryfinn:

* Target `simple_form` on this kit: Agreed
* Move nitro specific kits out of this repo and into nitro: Agreed

wadewinningham:
I think the biggest thing we want to avoid is business-specific logic within this repo.

As for Simple Form, I'm okay with it here but it doesn't seem to be a current dependency so I'm wondering how it could fit in in a way so that if we *do* include it, it could be an optional form system.

terryfinn:
> As for Simple Form, I'm okay with it here but it doesn't seem to be a current dependency so I'm wondering how it could fit in in a way so that if we do include it, it could be an optional form system.

@wadewinningham Is your thought we would not in this project target simple form, but provide code that if simple form is present, that it is configured to use kit code?

wadewinningham:
I was thinking that we target SimpleForm for now but plan for the option of alternates. But I like what you said.

terryfinn:
Still a work in progress, but this now works with both Simple Form and Form With.

playbook_ui.gemspec
```diff
@@ -35,6 +35,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency "rainbow", "2.2.2"
   s.add_development_dependency "rspec", "~> 3.8.0"
   s.add_development_dependency "rubocop", "0.72.0"
+  s.add_development_dependency "simple_form"
```
wadewinningham: Tie to a version?

terryfinn: @wadewinningham I can specify a more specific version.

xjunior: I wouldn't tie to a single version, but something more flexible like a trusted range. I say that because playbook is a library, and tying to a single version can be too hard.

app/pb_kits/playbook/pb_form/docs/example.yml
```diff
@@ -0,0 +1,5 @@
+examples:
+
+  rails:
+  - form_form_with: Default (Form With)
+  - form_simple_form: Default (Simple Form)
```
wadewinningham: Maybe my ignorance, but what's this for?

app/pb_kits/playbook/pb_text_input/_text_input.scss
```diff
@@ -8,51 +8,50 @@
...
-      opacity: 1;
-    }
-
-    &:-ms-input-placeholder {
-      color: $text_lt_lighter;
-    }
-
-    &::-ms-input-placeholder {
-      color: $text_lt_lighter;
+    input {
```
wadewinningham: I like everything else in this PR. This `input` is the one thing I put in the hands of UX to approve.

jasoncypret: I get it since it’s a block kit. Is there anything we can do to make this more specific? Wish we could add a class to whatever input they pass

terryfinn: I could change it to this to it would match the first immediate child that is a input.

```suggestion
    & > input:first-child {
```

jasoncypret: I like the update you made. I agree with the second as well. Might be something worth looking at in the future

app/pb_kits/playbook/pb_form/docs/example.yml
terryfinn: The example / documentation that playbook shows in the web ui for the project uses this yaml file to find and render all the examples and tells it what title to use for each example.

terryfinn:
@jasoncypret Thoughts on merging this?

thestephenmarshall:
@terryfinn could we possibly create a test page that uses this kit in Nitro before merging this? I think this would help.

terryfinn:
@thestephenmarshall Have you pulled a Playbook branch into Nitro before? I am trying right now to do it, but am getting errors coming from the playbook development configuration. Does our package task change the file structure when a release is cut?

I could use direction on pulling in a branch version into Nitro.

terryfinn:
Making progress getting a version of this integrated into Nitro: https://github.com/powerhome/nitro-web/pull/13362

PR #377, opened 2019-10-23
Chart kit Additions

jasperfurniss:
This PR adds a few additional to props to both the Y and X axis of the bar graph, enabling Jon to move forward with solar and style the bar graphs with our kit. It allows us to be able to define the tick interval, a min and max value for the labels on the axis, as well as custom categories. Looking for feedback!

### DIFF
--- a/app/pb_kits/playbook/pb_bar_graph/_bar_graph.html.erb
+++ b/app/pb_kits/playbook/pb_bar_graph/_bar_graph.html.erb
@@ -8,17 +8,17 @@
     window.addEventListener('DOMContentLoaded', function() {
       new pbChart('.selector', {
         id: "<%= object.id %>",
-        chartData: <%= object.sanitized_data(chart_data) %>,
+        chartData: <%= object.sanitized_chart_data %>,
         type: "<%= object.chart_type %>",
         title: "<%= object.title %>",
         subtitle: "<%= object.subtitle %>",
         axisTitle: "<%= object.axis_title %>",
         pointStart: <%= object.point_start %>,
-        xCategories: <%= object.x_categories.any? ? object.sanitized_data(x_categories) : "undefined" %>,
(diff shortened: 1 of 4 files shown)


app/pb_kits/playbook/pb_bar_graph/_bar_graph.html.erb
```diff
@@ -14,6 +14,14 @@
         subtitle: "<%= object.subtitle %>",
         axisTitle: "<%= object.axis_title %>",
         pointStart: <%= object.point_start %>,
+        xCategories: <%= object.x_categories.any? ? object.sanitized_x_categories : "undefined" %>,
```
terryfinn: We just push this logic up into the view object?

app/pb_kits/playbook/pb_bar_graph/_bar_graph.html.erb
```diff
@@ -14,6 +14,14 @@
...
+        xMin: <%= object.x_min %>,
+        xMax:<%= object.x_max %>,
+        xTickInterval: <%= object.x_tick_interval %>,
+        yCategories: <%= object.y_categories.any? ? object.sanitized_y_categories : "undefined"  %>,
```
terryfinn: We just push this logic up into the view object?

jasoncypret:
@jasperfurniss are these required now? Seems like we should make that optional

app/pb_kits/playbook/pb_bar_graph/_bar_graph.html.erb
xjunior: I think that all of this chart object could go into the view model.

jasoncypret:
@kre8sions Are you picking up this work, or starting a new PR?

thestephenmarshall:
@jasperfurniss any updates here?

PR #380, opened 2019-10-24
Bab/modernize source kit

gillianwenhold:
BAB 164 - Modernize Source Kit

### DIFF
--- a/app/pb_kits/playbook/pb_source/_source.html.erb
+++ b/app/pb_kits/playbook/pb_source/_source.html.erb
@@ -3,7 +3,7 @@
     data: object.data,
     class: object.classname) do %>
   <div class="pb__source_layout">
-    <% if object.hide_icon == false %>
+    <% if !object.hide_icon %>
       <% if object.show_icon? %>
         <%= pb_rails("icon_circle", props: { icon: object.type_icon_name, size: "sm" }) %>
       <% else %>
@@ -11,10 +11,12 @@
       <% end %>
     <% end %>
     <div class="pb__source_content">
-      <%= pb_rails("title", props: { text: object.source, tag: "h4", size: 4}) %>
+      <%= pb_rails("title", props: { text: object.source, tag: "h4", size: 4 }) %>
       <div class="pb__source_value">
         <%= pb_rails("body", props: { text: object.type_text, color: "light" }) %>
-        <%= object.user_id %>
+        <% if object.user_id %>
+          <%= pb_rails("caption", props: { text: object.user_id }) %>
+        <% end %>
       </div>
     </div>
   </div>
--- a/app/pb_kits/playbook/pb_source/_source.html.erb
+++ b/app/pb_kits/playbook/pb_source/_source.html.erb
@@ -14,9 +14,7 @@
       <%= pb_rails("title", props: { text: object.source, tag: "h4", size: 4 }) %>
       <div class="pb__source_value">
         <%= pb_rails("body", props: { text: object.type_text, color: "light" }) %>
-        <% if object.user_id %>
-          <%= pb_rails("caption", props: { text: object.user_id }) %>
-        <% end %>
+        <%= pb_rails("caption", props: { text: object.user_id }) if object.user_id %>
       </div>
     </div>
   </div>
--- a/app/pb_kits/playbook/pb_source/_source.html.erb
+++ b/app/pb_kits/playbook/pb_source/_source.html.erb
@@ -1,16 +1,19 @@
 <%= content_tag(:div,
     id: object.id,
     data: object.data,
-    class: object.classname("pb_source")) do %>
+    class: object.classname) do %>
   <div class="pb__source_layout">
     <% if object.hide_icon == false %>
-      <%= object.type_icon %>
-      <%= object.avatar %>
+      <% if object.show_icon? %>
+        <%= pb_rails("icon_circle", props: { icon: object.type_icon_name, size: "sm" }) %>
+      <% else %>
+        <%= pb_rails("avatar", props: object.avatar) %>
+      <% end %>
     <% end %>
     <div class="pb__source_content">
-      <%= object.source %>
+      <%= pb_rails("title", props: { text: object.source, tag: "h4", size: 4}) %>
       <div class="pb__source_value">
-        <%= object.type_text %>
+        <%= pb_rails("body", props: { text: object.type_text, color: "light" }) %>
         <%= object.user_id %>
       </div>
     </div>
--- a/app/pb_kits/playbook/pb_source/source.rb
+++ b/app/pb_kits/playbook/pb_source/source.rb
@@ -37,9 +37,7 @@ module Playbook
       end
 
       def user_id
-        if user.dig(:user_id)
-          user[:user_id]
-        end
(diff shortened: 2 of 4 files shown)


app/pb_kits/playbook/pb_source/_source.html.erb
```diff
@@ -1,17 +1,22 @@
...
     <div class="pb__source_content">
-      <%= object.source %>
+      <%= pb_rails("title", props: { text: object.source, tag: "h4", size: 4 }) %>
       <div class="pb__source_value">
-        <%= object.type_text %>
-        <%= object.user_id %>
+        <%= pb_rails("body", props: { text: object.type_text, color: "light" }) %>
+        <% if object.user_id %>
+          <%= pb_rails("caption", props: { text: object.user_id }) %>
+        <% end %>
```
garettarrowood: Nit:

You can make this one line:

```ruby
<%= pb_rails("caption", props: { text: object.user_id }) if object.user_id %>
```

app/pb_kits/playbook/pb_source/_source.html.erb
```diff
@@ -1,17 +1,22 @@
...
+      <% if object.show_icon? %>
+        <%= pb_rails("icon_circle", props: { icon: object.type_icon_name, size: "sm" }) %>
+      <% else %>
+        <%= pb_rails("avatar", props: object.avatar) %>
+      <% end %>
     <% end %>
     <div class="pb__source_content">
-      <%= object.source %>
+      <%= pb_rails("title", props: { text: object.source, tag: "h4", size: 4 }) %>
       <div class="pb__source_value">
```
garettarrowood: @markiearnold - All the `"pb__source..."` classes have 2 underscores. Is this on purpose? Does it communicate something about these styles?

app/pb_kits/playbook/pb_source/_source.html.erb
gillianwenhold: On it!

app/pb_kits/playbook/pb_source/_source.html.erb
gillianwenhold: It's written that way in the scss file...I think it's like that on purpose since they're inherited styles from the main pb_source classname

app/pb_kits/playbook/pb_source/source.rb
```diff
@@ -2,81 +2,46 @@
...
+      prop :user, type: Playbook::Props::Hash, default: {}
 
       def type_text
-        pb_type = Playbook::PbBody::Body.new(color: "light") do
-          if type == "user" || (type == "referral" && is_set?(configured_user))
-            configured_user[:name]
-          else
-            type.titleize
-          end
+        if type == "user" || (type == "referral" && user)
```
garettarrowood: `user`'s default is {}.  And `!!{}` is true. Should this check `!user.empty?`

```ruby
2.5.0 :002 > {}.empty?
 => true
2.5.0 :003 > {a: 1}.empty?
 => false
```

app/pb_kits/playbook/pb_source/source.rb
```diff
@@ -2,81 +2,46 @@
...
+          user[:name]
```
garettarrowood: Or possibly - this would be better:

```ruby
user.dig(:user)
```

app/pb_kits/playbook/pb_source/_source.html.erb
garettarrowood: Ah. Cool. Thanks!

app/pb_kits/playbook/pb_source/source.rb
gillianwenhold: Yes!

app/pb_kits/playbook/pb_source/source.rb
```diff
@@ -2,81 +2,46 @@
...
-        ApplicationController.renderer.render(partial: pb_type, as: :object)
       end
 
-      def type_icon
-        if !type_icon_name.nil? && avatar.nil?
-          pb_type_icon = Playbook::PbIconCircle::IconCircle.new(icon: type_icon_name, size: "sm")
-          ApplicationController.renderer.render(partial: pb_type_icon, as: :object)
-        end
+      def show_icon?
+        !type_icon_name.nil? && avatar.nil? ? true : false
```
garettarrowood: `type_icon_name.nil? && avatar.nil?` is already a conditional. It can only return `true` or `false`. So you can drop the ternary here.

app/pb_kits/playbook/pb_source/source.rb
```diff
@@ -2,81 +2,46 @@
...
       end
 
       def avatar
-        if is_set?(configured_user) && (type == "user" || type == "referral")
-          avatar_props = configured_user.clone
+        if user != {} && (type == "user" || type == "referral")
```
garettarrowood: Instead of `user != {}`, `!user.empty?` is a little more direct. If we add access to `ActiveSupport` in here, you could use `user.present?`

garettarrowood: One good article on these possibilities (there are many articles on this) - https://blog.appsignal.com/2018/09/11/differences-between-nil-empty-blank-and-present.html

app/pb_kits/playbook/pb_source/source.rb
```diff
@@ -2,81 +2,46 @@
...
-          ApplicationController.renderer.render(partial: pb_avatar, as: :object)
+          avatar_props
         end
       end
 
       def user_id
-        if is_set?(configured_user) && configured_user[:user_id].present?
-          pb_user_id = Playbook::PbCaption::Caption.new(text: configured_user[:user_id])
-          ApplicationController.renderer.render(partial: pb_user_id, as: :object)
+        if user && user[:user_id].present?
```
garettarrowood:
```ruby
user && user[:user_id].present?
```

should be replaced with

```ruby
user.dig(:user_id)
```

garettarrowood: Looks like we have `present?` available to us already :)

app/pb_kits/playbook/pb_source/source.rb
```diff
@@ -2,81 +2,46 @@
...
+          user[:user_id]
```
garettarrowood: Actually - this whole method should just be `user.dig(:user_id)`

spec/pb_kits/playbook/kits/source_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+  describe "#user_id" do
+    it "returns the user_id if a user is present with a user id, or returns nil", :aggregate_failures do
+      expect(subject.new(type: "prospecting", user: {}).user_id).to eq nil
+      expect(subject.new(type: "referral", user: { name: "User" }).user_id).to eq nil
+      expect(subject.new(type: "user", user: { name: "User", user_id: 435 }).user_id).to eq 435
+
+    end
+  end
+
+  describe "#show_icon?" do
```
garettarrowood: I love all the extra tests you put in here! Nice work!

gillianwenhold: Thanks!

PR #384, opened 2019-10-24
Modernize Icon Playbook Kit

kskilt:
Modernize Icon kit to match new formatting

Additional Information:
Commit 5d69785 adjusts the Enum specs to coerce array into strings for comparison. This allows the default to be nil.

### DIFF
--- a/app/pb_kits/playbook/pb_icon/_icon.html.erb
+++ b/app/pb_kits/playbook/pb_icon/_icon.html.erb
@@ -1,7 +1,4 @@
-<% if !object.icon_class.nil? %>
-  <%= content_tag(:i, nil,
-      id: object.id,
-      data: object.data,
-      class: object.classname(object.kit_class),
-      aria: object.aria) %>
-<% end %>
+<%= content_tag(:i, nil,
+    id: object.id,
+    data: object.data,
+    class: object.classname) %>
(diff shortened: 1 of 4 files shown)


app/pb_kits/playbook/pb_icon/_icon.html.erb
```diff
@@ -1,7 +1,4 @@
-<% if !object.icon_class.nil? %>
-  <%= content_tag(:i, nil,
-      id: object.id,
-      data: object.data,
-      class: object.classname(object.kit_class),
-      aria: object.aria) %>
```
garettarrowood: We want to keep `aria` in there. It's a base prop.

kskilt: Added back in

spec/support/playbook/rspec.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "pry"
```
garettarrowood: You will have to remove this too. It will likely fail your CI (or I hope it would).

kskilt: It did fail, and I thought we ran `git co` on this file. Good catch!

kskilt:
> LGTM. Did you smoke test this?

Yup, threw it up on localhost:8080 and compared to prod. Everything looked the same

PR #386, opened 2019-10-24
Modernize Stat Value Playbook Kit

kskilt:
Common modernization to the Stat Value kit to match our current formatting.

Additional Information:
This kit originally accepted strings or integers where it was displaying values. The values were always an integer, but it was formatted this way in the examples if the end user typed in a string. To better stream line, we made the prop always a number, however this may effect previous uses that contained strings. With that in mind, we should watch any previous Stat Value pbkit or Dashboard Value kit (which inherits Stat Value in it) for backwards compatibility issues. It will be an easy fix of adjusting the value to an integer.

### DIFF
--- a/app/pb_kits/playbook/pb_stat_value/stat_value.rb
+++ b/app/pb_kits/playbook/pb_stat_value/stat_value.rb
@@ -2,61 +2,17 @@
 
 module Playbook
   module PbStatValue
-    class StatValue < Playbook::PbKit::Base
-      PROPS = %i[configured_classname
-                 configured_data
-                 configured_id
-                 configured_unit
-                 configured_value].freeze
+    class StatValue
+      include Playbook::Props
 
-      def initialize(classname: default_configuration,
-                     data: default_configuration,
-                     id: default_configuration,
-                     unit: default_configuration,
-                     value: default_configuration)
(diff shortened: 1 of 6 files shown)


app/pb_kits/playbook/pb_stat_value/stat_value.rb
```diff
@@ -2,61 +2,17 @@
...
-        "pb_stat_value_kit"
-      end
-
-      def to_partial_path
-        "pb_stat_value/stat_value"
-      end
+      partial "pb_stat_value/stat_value"
 
-    private
+      prop :unit
```
gillianwenhold: Does this one need a `Playbook::Props::String` and `default: nil`?

garettarrowood: He could have specified these. But since they are the default, it's okay to leave them off.

kskilt: So for a string prop, it is the default and automatically assumed by the code. No need to specify that or a default as nil.

kskilt:
> No added spec file???
@KFad11 
End of the week vibes 😄 Added spec file now.

PR #391, opened 2019-10-28
Modernize label value kit

gillianwenhold:
BAB-157 Modernized label value playbook kit

Note: Label and value are both required props, may cause backwards compatibility issues

### DIFF
--- a/app/pb_kits/playbook/pb_label_value/_label_value.html.erb
+++ b/app/pb_kits/playbook/pb_label_value/_label_value.html.erb
@@ -2,8 +2,6 @@
     id: object.id,
     data: object.data,
     class: object.classname) do %>
-  <% if object.label? %>
-    <%= pb_rails("caption", props: { text: "#{object.label}" }) %>
-  <% end %>
+  <%= pb_rails("caption", props: { text: "#{object.label}" }) if object.label? %>
   <%= pb_rails("body", props: { text: "#{object.value}" }) %>
 <% end %>
--- a/app/pb_kits/playbook/pb_label_value/_label_value.html.erb
+++ b/app/pb_kits/playbook/pb_label_value/_label_value.html.erb
@@ -2,6 +2,6 @@
     id: object.id,
     data: object.data,
     class: object.classname) do %>
-  <%= pb_rails("caption", props: { text: object.label }) if object.label? %>
+  <%= pb_rails("caption", props: { text: object.label }) if object.label %>
   <%= pb_rails("body", props: { text: object.value }) %>
 <% end %>
--- a/app/pb_kits/playbook/pb_label_value/_label_value.html.erb
+++ b/app/pb_kits/playbook/pb_label_value/_label_value.html.erb
@@ -1,7 +1,9 @@
 <%= content_tag(:div,
     id: object.id,
     data: object.data,
-    class: object.classname(object.kit_class)) do %>
-  <%= object.label %>
-  <%= object.value %>
+    class: object.classname) do %>
+  <% if object.label? %>
+    <%= pb_rails("caption", props: { text: "#{object.label}" }) %>
+  <% end %>
+  <%= pb_rails("body", props: { text: "#{object.value}" }) %>
 <% end %>
--- a/app/pb_kits/playbook/pb_label_value/_label_value.html.erb
+++ b/app/pb_kits/playbook/pb_label_value/_label_value.html.erb
@@ -2,6 +2,6 @@
     id: object.id,
     data: object.data,
     class: object.classname) do %>
-  <%= pb_rails("caption", props: { text: "#{object.label}" }) if object.label? %>
-  <%= pb_rails("body", props: { text: "#{object.value}" }) %>
+  <%= pb_rails("caption", props: { text: object.label }) if object.label? %>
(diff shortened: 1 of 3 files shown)


app/pb_kits/playbook/pb_label_value/_label_value.html.erb
```diff
@@ -1,7 +1,7 @@
 <%= content_tag(:div,
     id: object.id,
     data: object.data,
-    class: object.classname(object.kit_class)) do %>
-  <%= object.label %>
-  <%= object.value %>
+    class: object.classname) do %>
+  <%= pb_rails("caption", props: { text: "#{object.label}" }) if object.label? %>
+  <%= pb_rails("body", props: { text: "#{object.value}" }) %>
```
garettarrowood: No string interpolation is needed because `label` and `value` are already strings.

app/pb_kits/playbook/pb_label_value/label_value.rb
```diff
@@ -2,55 +2,22 @@
...
-        ApplicationController.renderer.render(partial: pb_body, as: :object)
-      end
+      prop :label, type: Playbook::Props::String
+      prop :value, type: Playbook::Props::String,
+                   default: ""
 
-      def kit_class
-        "pb_label_value_kit"
+      def label?
+        label.present?
```
garettarrowood: Since `label` defaults to `nil`, which is `falsy`, there is not a need to make this presence checking method. `if object.label` would be fine to use in your template.

app/pb_kits/playbook/pb_label_value/_label_value.html.erb
```diff
@@ -1,7 +1,7 @@
 <%= content_tag(:div,
     id: object.id,
     data: object.data,
-    class: object.classname(object.kit_class)) do %>
-  <%= object.label %>
-  <%= object.value %>
+    class: object.classname) do %>
+  <%= pb_rails("caption", props: { text: object.label }) if object.label? %>
```
garettarrowood:
```suggestion
  <%= pb_rails("caption", props: { text: object.label }) if object.label %>
```

app/pb_kits/playbook/pb_label_value/label_value.rb
```diff
@@ -2,55 +2,18 @@
...
-
-      def to_partial_path
-        "pb_label_value/label_value"
-      end
+      partial "pb_label_value/label_value"
 
-    private
+      prop :label, type: Playbook::Props::String
+      prop :value, type: Playbook::Props::String,
+                   default: ""
```
KFad11: So for the props, the default type is `string` & the default value in `nil`, so the `, type: Playbook::Props::String` & `, defaults: ""` wouldn't be necessary here.

gillianwenhold: Thank you! Changing that here / in the other PRs too

KFad11: I was about to add the same comment to the other PR... I shall refrain. lol

gillianwenhold: Haha sounds good - thanks!

app/pb_kits/playbook/pb_label_value/_label_value.html.erb
```diff
@@ -1,7 +1,7 @@
 <%= content_tag(:div,
     id: object.id,
     data: object.data,
-    class: object.classname(object.kit_class)) do %>
-  <%= object.label %>
-  <%= object.value %>
+    class: object.classname) do %>
+  <%= pb_rails("caption", props: { text: object.label }) %>
+  <%= pb_rails("body", props: { text: object.value }) %>
```
garettarrowood: It doesn't look like the Body.new invocation was receiving the `text` prop. It looks like it was receiving the object's value via a block.

app/pb_kits/playbook/pb_label_value/_label_value.html.erb
```diff
@@ -1,7 +1,7 @@
 <%= content_tag(:div,
     id: object.id,
     data: object.data,
-    class: object.classname(object.kit_class)) do %>
-  <%= object.label %>
-  <%= object.value %>
+    class: object.classname) do %>
+  <%= pb_rails("caption", props: { text: object.label }) %>
```
garettarrowood: Doesn't this still need the conditional? Only render a label if a label is passed?

If this prop is truly required (and the conditional is inappropriate), let's make it `required: true`

gillianwenhold: The label isn't currently required, but I think it might make sense for it to be since without a label a body kit alone would be more appropriate?

garettarrowood: I agree. It makes sense that a kit with the word "label" in it means that the prop is required.

app/pb_kits/playbook/pb_label_value/_label_value.html.erb
garettarrowood: We discussed this via a screen share and found out that this will work the same way for the body kit.

PR #392, opened 2019-10-28
Modernize title-detail kit

gillianwenhold:
BAB-173: Modernize title detail kit

Note: Title and detail are both required props, may cause backwards compatibility issues

### DIFF
--- a/app/pb_kits/playbook/pb_title_detail/_title_detail.html.erb
+++ b/app/pb_kits/playbook/pb_title_detail/_title_detail.html.erb
@@ -2,6 +2,6 @@
     id: object.id,
     data: object.data,
     class: object.classname) do %>
-  <%= pb_rails("title", props: { text: object.title, size: 4 }) if object.title %>
-  <%= pb_rails("body", props: { text: object.detail, color: "light" }) if object.detail %>
+  <%= pb_rails("title", props: { text: object.title, size: 4 }) %>
+  <%= pb_rails("body", props: { text: object.detail, color: "light" }) %>
 <% end %>
--- a/app/pb_kits/playbook/pb_title_detail/_title_detail.html.erb
+++ b/app/pb_kits/playbook/pb_title_detail/_title_detail.html.erb
@@ -2,6 +2,6 @@
     id: object.id,
     data: object.data,
     class: object.classname) do %>
-  <%= pb_rails("title", props: { text: object.title, size: 4 }) if object.title? %>
-  <%= pb_rails("body", props: { text: object.detail, color: "light" }) if object.detail? %>
+  <%= pb_rails("title", props: { text: object.title, size: 4 }) if object.title %>
+  <%= pb_rails("body", props: { text: object.detail, color: "light" }) if object.detail %>
 <% end %>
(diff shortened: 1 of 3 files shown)


app/pb_kits/playbook/pb_title_detail/_title_detail.html.erb
```diff
@@ -1,7 +1,7 @@
 <%= content_tag(:div,
     id: object.id,
     data: object.data,
-    class: object.classname(object.kit_class)) do %>
-  <%= object.title %>
-  <%= object.detail %>
+    class: object.classname) do %>
+  <%= pb_rails("title", props: { text: "#{object.title}", size: 4 }) if object.title? %>
+  <%= pb_rails("body", props: { text: "#{object.detail}", color: "light" }) if object.detail? %>
```
garettarrowood: No need for string interpolation in these two. `title` and `detail` are already `String` props.

app/pb_kits/playbook/pb_title_detail/title_detail.rb
```diff
@@ -2,69 +2,29 @@
...
-      end
 
-      def kit_class
-        kit_options = [
-          "pb_title_detail_kit",
-          align,
-        ]
-        kit_options.join("_")
+      def title?
+        title.present?
```
garettarrowood: Since title defaults to `nil`, these presence checking methods are unneeded.

app/pb_kits/playbook/pb_title_detail/_title_detail.html.erb
```diff
@@ -1,7 +1,7 @@
 <%= content_tag(:div,
     id: object.id,
     data: object.data,
-    class: object.classname(object.kit_class)) do %>
-  <%= object.title %>
-  <%= object.detail %>
+    class: object.classname) do %>
+  <%= pb_rails("title", props: { text: object.title, size: 4 }) if object.title %>
+  <%= pb_rails("body", props: { text: object.detail, color: "light" }) if object.detail %>
```
garettarrowood: It looks like the previous use of the `body` kit was passing the `detail` as a block. Does this work the same way for this use case?

gillianwenhold: No - this one passes the title and detail text to kits that will render if passed a string (defaulting to nil, and not rendering without one)

app/pb_kits/playbook/pb_title_detail/_title_detail.html.erb
```diff
@@ -1,7 +1,7 @@
 <%= content_tag(:div,
     id: object.id,
     data: object.data,
-    class: object.classname(object.kit_class)) do %>
-  <%= object.title %>
-  <%= object.detail %>
+    class: object.classname) do %>
+  <%= pb_rails("title", props: { text: object.title, size: 4 }) %>
```
garettarrowood: Without the conditional, does this make `title` a required prop?

app/pb_kits/playbook/pb_title_detail/_title_detail.html.erb
```diff
@@ -1,7 +1,7 @@
...
+  <%= pb_rails("body", props: { text: object.detail, color: "light" }) %>
```
garettarrowood: Same here. This was wrapped in a conditional in the view object before. If we remove it here, are we saying that the `detail` is a required prop?

PR #393, opened 2019-10-28
Modernize line graph kit

gillianwenhold:
BAB-158: Modernize line graph kit

Note: Updated prop `point_start` for bar graph and line graph kits to a Numeric

### DIFF
--- a/app/pb_kits/playbook/pb_line_graph/_line_graph.html.erb
+++ b/app/pb_kits/playbook/pb_line_graph/_line_graph.html.erb
@@ -2,14 +2,14 @@
     aria: object.aria,
     id: object.id,
     data: object.data,
-    class: object.classname("pb_line_graph") 
+    class: object.classname
     ) %>
 <% content_for :pb_js do %>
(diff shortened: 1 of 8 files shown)


app/pb_kits/playbook/pb_line_graph/_line_graph.html.erb
```diff
@@ -2,14 +2,14 @@
...
     ) %>
 <% content_for :pb_js do %>
   <%= javascript_tag do %>
     window.addEventListener('DOMContentLoaded', function() {
       new pbChart('.selector', {
         id: "<%= object.id %>",
-        chartData: <%= object.chart_data %>,
+        chartData: <%= object.chart_data.to_json.html_safe %>,
         type: "<%= object.chart_type %>",
         title: "<%= object.title %>",
```
garettarrowood: All of these properties can be moved into the class definition. Something like:

```ruby
def chart_options
  {
    id: id,
    chartData: chart_data.to_json.html_safe,
    etc...
  }
end
```

PR #408, opened 2019-10-31
Add contact_details prop to contact kit

KFad11:
Added a new prop to contact kit so we can show who's contact info the specific entries belong to. ex: John's Cell, Home, Pauline's Work.

### DIFF
--- a/app/pb_kits/playbook/pb_contact/_contact.jsx
+++ b/app/pb_kits/playbook/pb_contact/_contact.jsx
@@ -27,14 +27,14 @@ const Contact = ({
   className,
   dark=false,
   contactValue,
-  contactDetail,
+  contactDetail="",
 }: ContactProps) => {
 
   const formatDetail = (contactDetail) => {
     if (contactDetail !== undefined) {
       return `\u00b7 ${contactDetail}`
     } else {
-      return ""
+      return contactDetail
     }
   }
 
@@ -44,11 +44,11 @@ const Contact = ({
     if (contactType == "email") {
       return contactString
     } else {
-      let cleaned = ('' + contactString).replace(/\D/g, '')
-      let match = cleaned.match(/^(1|)?(\d{3})(\d{3})(\d{4})$/)
-      if (match) {
-        let intlCode = (match[1] ? '+1 ' : '')
-        return [intlCode, '(', match[2], ') ', match[3], '-', match[4]].join('')
+      const cleaned = contactString.replace(/\D/g, '')
+      const phoneNumber = cleaned.match(/^(1|)?(\d{3})(\d{3})(\d{4})$/)
+      if (phoneNumber) {
+        let intlCode = (phoneNumber[1] ? '+1 ' : '')
+        return [intlCode, '(', phoneNumber[2], ') ', phoneNumber[3], '-', phoneNumber[4]].join('')
       }
       return null
     }
--- a/app/pb_kits/playbook/pb_contact/_contact.jsx
+++ b/app/pb_kits/playbook/pb_contact/_contact.jsx
@@ -14,6 +14,7 @@ type ContactProps = {
   className?: String | Array<String>,
   dark?: Boolean,
   contactValue: String,
+  contactDetail?: String,
 }
 
 const kitClasses = ({}: ContactProps) => {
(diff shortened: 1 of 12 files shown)


markiearnold:
I don't think we want to make this into a reusable pattern yet. To display a person with contact information, we have the [Person Contact kit](https://playbook.powerapp.cloud/kits/person_contact). 

For this PR, the current Nitro pattern for displaying contact information for two people (ie Jane's cell) is on our upcoming list to redesign.  If you can code this solution into your page, but not within the kit - I think that will be a better option for now.  /cc @jasoncypret

xjunior:
@markiearnold I feel like playbook is missing one key feature which is extensibility. Some of the kits we're adding to this library are really very nitro specific UI's that could use the playbook framework (with view objects and all) but live in nitro.

app/pb_kits/playbook/pb_contact/_contact.jsx
```diff
@@ -26,15 +27,24 @@ const Contact = ({
...
 }: ContactProps) => {
 
+  const formatDetail = (contactDetail) => {
+    if (contactDetail !== undefined) {
+      return `\u00b7 ${contactDetail}`
+    } else {
+      return ""
+    }
+  }
+
```
thestephenmarshall:
```suggestion
const formatDetail = (contactDetail) => {
    if (contactDetail !== undefined) {
      return `\u00b7 ${contactDetail}`
    } else {
      return contactDetail
    }
}
```

app/pb_kits/playbook/pb_contact/_contact.jsx
```diff
@@ -26,15 +27,24 @@ const Contact = ({
...
   dark=false,
   contactValue,
+  contactDetail,
 }: ContactProps) => {
 
+  const formatDetail = (contactDetail) => {
+    if (contactDetail !== undefined) {
+      return `\u00b7 ${contactDetail}`
+    } else {
+      return ""
```
thestephenmarshall: Consider returning the default set from the props

app/pb_kits/playbook/pb_contact/_contact.jsx
```diff
@@ -26,15 +27,24 @@ const Contact = ({
   className,
   dark=false,
   contactValue,
+  contactDetail,
```
thestephenmarshall:
```suggestion
  contactDetail="",
```

app/pb_kits/playbook/pb_contact/_contact.jsx
```diff
@@ -26,15 +27,24 @@ const Contact = ({
...
   const css = classnames(kitClasses({contactType}), className)
 
-  const formatPhoneNumber = (phoneNumberString, contactType) => {
+  const formatContact = (contactString, contactType) => {
     if (contactType == "email") {
-      return phoneNumberString
+      return contactString
     } else {
-      let cleaned = ('' + phoneNumberString).replace(/\D/g, '')
+      let cleaned = ('' + contactString).replace(/\D/g, '')
```
thestephenmarshall: Since we aren't mutating `cleaned`, use `const`

```suggestion
      const cleaned = (contactString).replace(/\D/g, '')
```

also, can `contactString` possibly come through as another type? Is that why we are forcibly stringifying here? `' ' + contactString...`

app/pb_kits/playbook/pb_contact/_contact.jsx
```diff
@@ -26,15 +27,24 @@ const Contact = ({
...
       let match = cleaned.match(/^(1|)?(\d{3})(\d{3})(\d{4})$/)
```
thestephenmarshall: Also not mutating `match` here so use `const`

```suggestion
      const match = cleaned.match(/^(1|)?(\d{3})(\d{3})(\d{4})$/)
```

also consider this variable name as the regex is only sort-of clear to readers. Maybe a more descriptive name that tells us `what we are matching` would be better here.

thestephenmarshall:
I defer to what @markiearnold and @jasoncypret think for this particular kits future, ultimately. @xjunior you are right and we are starting to hone in on that and try to genericize as we go!

garettarrowood:
This seems like a great spot to add this, so here goes.

The need for `contact_detail` on views in Nitro can be avoided because we can list the specific `person`'s contacts directly under them. However, there are still places in the system where one person has more than one cell phone. Here is an example:

Screen Shot 2019-11-01 at 7 24 50 AM

I can not attest to how or why this shows up this way at the business level. But at our database level, that `1st` and `2nd` cell is stored in a serialized field on an `owner_referral` object, which is affiliated with an `owner`, which embodies both of those names. Meaning - we'd only be making a guess at whose is whose.

Last thought --

Let's NOT use the detail on the homes page, as we have discussed. But having an extra piece of info that comes after a number seems like an extensible addition. I vote we keep this.

app/pb_kits/playbook/pb_contact/_contact.jsx
KFad11: No it looks like that's going to be passed as the `contactValue` which will be a String, so we should be OK to remove it. Or

app/pb_kits/playbook/pb_contact/_contact.jsx
KFad11: Got it changed! Thanks Stephen.

app/pb_kits/playbook/pb_contact/_contact.jsx
KFad11: Hey @thestephenmarshall, is this just a suggested indentation change?  This `const` is the first in the `ContactProps` hash, so they're all indented. Let me know if something still needs adjusting here. Thanks!

wadewinningham:
@KFad11 What does this look like? Can you post a screenshot of an example?

wadewinningham:
We definitely need a way to provide a *label* for a phone number. I think I prefer that to *detail*.

While the existing phone number display does incorporate an icon, some folks have multiple numbers (I honestly don't know how many mobile numbers Tim has) and we need some way to distinguish between them.

KFad11:
Screen Shot 2019-11-05 at 9 09 05 AM

How the details would look.

PR #411, opened 2019-11-01
Adjust HomeAddressStreet kit to have optional H#

kskilt:
Currently if a Home Address Street kit was not passed an home_id, a bubble would still appear with H#. This wraps that prop in a conditional that will only display if home_id is not nil. If no home_id is passed, territory will simply fill the space. 

Additional information:
The H# is currently not using the Hashtag kit, but instead a badge kit. I made an initial attempt to migrate this to the Hashtag kit so we dont have to hardcode "H#" before the home_id, however the Hashtag kit requires a url prop taking this out of the scope of this card. This is something I will bring up to Garett as a possibly follow up card.

### DIFF
--- a/app/pb_kits/playbook/pb_home_address_street/home_address_street.rb
+++ b/app/pb_kits/playbook/pb_home_address_street/home_address_street.rb
@@ -15,7 +15,6 @@ module Playbook
       prop :state
       prop :zipcode
       prop :territory
-      prop :url
       prop :dark, type: Playbook::Props::Boolean, default: false
 
       def classname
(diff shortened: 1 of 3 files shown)


app/pb_kits/playbook/pb_home_address_street/home_address_street.rb
```diff
@@ -15,6 +15,7 @@ class HomeAddressStreet
       prop :state
       prop :zipcode
       prop :territory
+      prop :url
```
gillianwenhold: Is this url prop still needed?

kskilt:
Good catch! I got rid of the definition it was used in but not the prop above.

PR #412, opened 2019-11-01
Titleize Home Address Street Playbook Kit

kskilt:
This will titleize the Address and City props.

Additional Information:
I had only placed a safety navigator on the address_cont since this is the only field that needs to be titleized but also optional. Street and City will always exist.

### DIFF
--- a/app/pb_kits/playbook/pb_home_address_street/_home_address_street.jsx
+++ b/app/pb_kits/playbook/pb_home_address_street/_home_address_street.jsx
@@ -15,6 +15,24 @@ const dot = (houseStyle) =>
   }
 }
 
+function titleize(sentence) {
+    if(!sentence.split) return sentence;
+    var _titleizeWord = function(string) {
+            return string.charAt(0).toUpperCase() + string.slice(1).toLowerCase();
+        },
+        result = [];
+    sentence.split(" ").forEach(function(w) {
+        result.push(_titleizeWord(w));
+    });
+    return result.join(" ");
+}
+
+function titleize_address_cont(address_cont) {
+  if (!!address_cont) {
+   return titleize(address_cont);
+  }
+}
+
 type HomeAddressStreetProps = {
   address: String,
   address_cont: String,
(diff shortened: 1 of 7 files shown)


kskilt:
@jasoncypret  Took a little finagling (had to google that spelling), but I got it working on the React side with same functionality (Safety navigator on Address_cont only as this is the only optional prop that would be titleized)

app/pb_kits/playbook/pb_home_address_street/_home_address_street.jsx
```diff
@@ -15,6 +15,24 @@ const dot = (houseStyle) =>
...
+    if(!sentence.split) return sentence;
+    var _titleizeWord = function(string) {
+            return string.charAt(0).toUpperCase() + string.slice(1).toLowerCase();
+        },
+        result = [];
+    sentence.split(" ").forEach(function(w) {
+        result.push(_titleizeWord(w));
+    });
+    return result.join(" ");
+}
```
thestephenmarshall: Utilities like this could be very useful across many future kits. What do you think about maybe moving this to a new file under `/app/pb_kits/playbook/utilities/text.js` or similar?

app/pb_kits/playbook/pb_home_address_street/docs/_home_address_street_dark.html.erb
```diff
@@ -1,5 +1,5 @@
 <%= pb_rails("home_address_street", props: {
-  address: "70 Prospect Ave",
+  address: "70 prospect AVE",
```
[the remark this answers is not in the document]
kskilt: I changed this back to proper text since it was visible in the code example.

app/pb_kits/playbook/pb_home_address_street/_home_address_street.jsx
kskilt: Moved to a new file 'utilities/text.js'. Thanks for the help on that!

app/pb_kits/playbook/pb_home_address_street/_home_address_street.jsx
```diff
@@ -15,6 +17,12 @@ const dot = (houseStyle) =>
   }
 }
 
+function titleize_address_cont(address_cont) {
+  if (!!address_cont) {
+   return titleize(address_cont);
+  }
+}
```
xjunior: Please, follow the javascript naming convention (lower-first camel case). Also, define this function as a constant:

```suggestion
const titleizeAddessCont = (addressCont) => addressCont ? titleize(addressCont) : null
```

app/pb_kits/playbook/utilities/text.js
```diff
@@ -0,0 +1,16 @@
+function titleize(sentence) {
```
xjunior: Have a look at lodash's `startCase`: https://lodash.com/docs/4.17.15#startCase

app/pb_kits/playbook/pb_home_address_street/_home_address_street.jsx
kskilt: These changes have been added

kskilt:
@xjunior @jasoncypret  Hey guys, Just reaching out to see if there is any more necessary changes to this.

PR #414, opened 2019-11-01
Modernize layout kit

gillianwenhold:
BAB-235: Modernize Playbook Layout/LayoutBody/LayoutSidebar kit(s)

### DIFF
--- a/app/pb_kits/playbook/pb_layout/body.rb
+++ b/app/pb_kits/playbook/pb_layout/body.rb
@@ -2,38 +2,18 @@
 
 module Playbook
   module PbLayout
-    class Body < Playbook::PbKit::Base
-      PROPS = %i[configured_classname
-                 configured_data
-                 configured_id
-                 block].freeze
+    class Body
+      include Playbook::Props
 
-      def initialize(classname: default_configuration,
-                     data: default_configuration,
-                     id: default_configuration,
-                     &block)
-        self.configured_classname = classname
-        self.configured_data = data
-        self.configured_id = id
-        self.block = block_given? ? block : nil
+      partial "pb_layout/body"
+
+      def classname
+        generate_classname("layout_body")
       end
 
       def yield(context:)
-        context.capture(&block)
+        context.capture(&children)
       end
-
-      def to_partial_path
-        "pb_layout/body"
-      end
-
-    private
-
-      DEFAULT = Object.new
-      private_constant :DEFAULT
-      def default_configuration
-        DEFAULT
-      end
-      attr_accessor(*PROPS)
     end
   end
 end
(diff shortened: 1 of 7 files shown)


app/pb_kits/playbook/pb_layout/layout.rb
```diff
@@ -2,92 +2,47 @@
...
-        else
-          " layout#{position}_collapse_#{configured_collapse}"
-        end
+      def transparent_class
+        transparent ? "transparent" : nil
       end
 
-      def full
-        true_value(configured_full, " full", "")
+      def collapse_class
```
garettarrowood: Instead of putting whitespace in this and the `full` classnames, just join the three on `" "` in your classname method.

app/pb_kits/playbook/pb_layout/body.rb
```diff
@@ -2,38 +2,18 @@
...
       end
 
-    private
-
-      DEFAULT = Object.new
-      private_constant :DEFAULT
-      def default_configuration
-        DEFAULT
+      def yield(context:)
+        context.capture(&children)
```
garettarrowood: We've removed the custom `yield` methods in the new format. It's better to just `<%= capture(&object.children) %>` inside the template itself, instead of passing the template into the view object to invoke a method on it.

app/pb_kits/playbook/pb_layout/layout.rb
```diff
@@ -2,92 +2,44 @@
...
-    private
-
-      DEFAULT = Object.new
-      private_constant :DEFAULT
-      def default_configuration
-        DEFAULT
+      def classname
+        first_class = generate_classname("pb_layout", size, position, dark_class, transparent_class)
+        classname = first_class + full_class + collapse_class
+        [first_class, full_class, collapse_class].reject(&:empty?).join(" ")
```
garettarrowood: Since all of the `_class` methods are now only used by the `classname` method, they can all be moved into `private` access.

app/pb_kits/playbook/pb_layout/layout.rb
```diff
@@ -2,92 +2,44 @@
...
+                      default: "xs"
+      prop :dark, type: Playbook::Props::Boolean, default: false
+      prop :full, type: Playbook::Props::Boolean, default: false
+      prop :position, type: Playbook::Props::Enum,
+                      values: %w[left right],
+                      default: "left"
+      prop :transparent, type: Playbook::Props::Boolean, default: false
+      prop :size, type: Playbook::Props::Enum,
+                      values: %w[xs sm md lg xl],
+                      default: "md"
```
garettarrowood: tiny nit: indentation alignment of `values` and `default` with `type`

PR #419, opened 2019-11-04
Modernize User Kit

eabousaif:
Modernize the user kit, add specs, and modify the avatar image URL to be encapsulated in one prop as opposed to it being a nested hash. Updated the Avatar kit as well to simplify that hash.

### DIFF
--- a/app/pb_kits/playbook/pb_user/_user.html.erb
+++ b/app/pb_kits/playbook/pb_user/_user.html.erb
@@ -1,11 +1,22 @@
 <%= content_tag(:div,
     id: object.id,
     data: object.data,
-    class: object.classname(object.kit_class)) do %>
-  <%= object.avatar %>
+    class: object.classname) do %>
+  <% if object.avatar_url.present? || object.avatar %>
+    <%= pb_rails("avatar", props: {
+      name: object.name,
+      size: object.avatar_size,
+      image: {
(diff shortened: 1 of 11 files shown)


app/pb_kits/playbook/pb_user/_user.html.erb
```diff
@@ -1,11 +1,22 @@
...
     id: object.id,
     data: object.data,
-    class: object.classname(object.kit_class)) do %>
-  <%= object.avatar %>
+    class: object.classname) do %>
+  <% if object.avatar_url.present? || object.avatar %>
+    <%= pb_rails("avatar", props: {
+      name: object.name,
+      size: object.avatar_size,
+      image: {
```
gillianwenhold: Instead of this hash, should it be `image_url: object.avatar_url` ?

eabousaif: We're being explicit with this one to reach into the hash and get the image url for it to be then used as image_url on the front-end.

jasoncypret:
This change to the props will likely break any user kit in Nitro. Are you planning on updating any kit in Nitro after this launch?

eabousaif:
Yes. There is only one place that uses it and I have it marked to make it compatible once the new PB version is released

PR #423, opened 2019-11-05
Modernize dashboard value

gillianwenhold:
BAB-181: Modernize DashboardValue Playbook Kit

Note: This modernization changes the prop type of `:stat_label` in rails from a hash `stat_label: { label: "This is the label" }` to a string `stat_label: "This is the label"`. 

Backwards breaking changes: 
components/dev_docs/app/views/dev_docs/playbook/index.html.erb
components/solar_roofing/app/views/solar_roofing/solar_proposals/_overview.html.erb


gillianwenhold:
@jasoncypret Yes! We'll do that when raising the version in Nitro. I'll update the React version now

PR #424, opened 2019-11-05
Home tour react kits

megantrimble:


app/pb_kits/playbook/pb_circle_icon_button/_circle_icon_button.scss
```diff
@@ -0,0 +1,41 @@
...
+  position: relative !important;
+  width: $pb_circle_icon_button_size !important;
+  height: $pb_circle_icon_button_size !important;
+  border-radius: $border_rad_mega !important;
+  line-height: $pb_circle_icon_button_size !important;
+  flex-shrink: 0 !important;
+  flex-grow: 0 !important;
+  flex-basis: $pb_circle_icon_button_size !important;
+  min-width: $pb_circle_icon_button_size !important;
+  padding: 0 !important;
```
thestephenmarshall: This is super important... why is this necessary here?

app/pb_kits/playbook/pb_text_input/_text_input.scss
```diff
@@ -3,6 +3,59 @@
...
+  display: block;
+  width: 100%;
+  color: $color;
+  font-weight: bold;
+  background: $bg;
+  padding: $space_xs $space_sm;
+  transition: 0.2s all;
+  outline: none;
+  font-size: $font_small;
+  margin: .1rem 0 0;
```
thestephenmarshall: We are using `rem` and `px` here.. is there a reason for using `rem` where `px` wouldn't quite work out?

jasperfurniss: This should be using our spacing variable with a math calculation of some sort too probably.

PR #426, opened 2019-11-05
Modernize Message Kit

eabousaif:
Modernize the message kit and write spec file

### DIFF
--- a/app/pb_kits/playbook/pb_message/_message.html.erb
+++ b/app/pb_kits/playbook/pb_message/_message.html.erb
@@ -2,16 +2,14 @@
     id: object.id,
     data: object.data,
     class: object.classname) do %>
-    <% if object.avatar_url.present? || object.avatar %>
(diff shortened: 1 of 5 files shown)


app/pb_kits/playbook/pb_message/_message.html.erb
```diff
@@ -1,11 +1,24 @@
 <%= content_tag(:div,
     id: object.id,
     data: object.data,
-    class: object.classname(object.kit_class)) do %>
-    <%= object.avatar %>
+    class: object.classname) do %>
+    <% if object.avatar_url.present? || object.avatar_name %>
```
thestephenmarshall: Could we possibly abstract this logic away from the ERB into `app/pb_kits/playbook/pb_message/message.rb` ? Wondering if we couldn't do something across all kits like `object.valid?`

eabousaif: Abstracted the logic into the rb file and just called `object.valid?`

PR #427, opened 2019-11-05
Add "wrong" tag to contacts

KFad11:
Adds a wrong tag to contacts so that we can display them differently than the other contacts.

### DIFF
--- a/app/pb_kits/playbook/pb_person_contact/_person_contact.html.erb
+++ b/app/pb_kits/playbook/pb_person_contact/_person_contact.html.erb
@@ -15,4 +15,13 @@
           contact_detail: contact[:contact_detail],
       }) %>
     <% end %>
+    <% unless object.wrong_contacts.length == 0 %>
+      <%= pb_rails("caption", props: { text: "wrong number" })%>
+      <% object.wrong_numbers.each do |wrong_number| %>
+        <%= pb_rails("contact", props: {
+            contact_type: wrong_number[:contact_type],
+            contact_value: wrong_number[:contact_value],
+        }) %>
+      <% end %>
+    <% end %>
 <% end %>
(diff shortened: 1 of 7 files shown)


app/pb_kits/playbook/pb_person_contact/_person_contact.html.erb
```diff
@@ -15,4 +15,13 @@
           contact_detail: contact[:contact_detail],
       }) %>
     <% end %>
+    <% unless object.wrong_contacts.length == 0 %>
```
thestephenmarshall: `unless` makes things murky to others when reading 

```erb
<% if object.wrong_contact.present? %>
```

app/pb_kits/playbook/pb_person_contact/person_contact.rb
```diff
@@ -9,6 +9,17 @@ class PersonContact
...
       prop :people, type: Playbook::Props::HashArray, default: []
+      prop :wrong_contacts, type: Playbook::Props::HashArray, default: []
+
+      def filter_number
+        if contact_type = "wrong number"
+          wrong_number << contact
+          positon = contacts.find_index(contact)
+          contacts.delete_at(position)
+        end
+        wrong_number
```
thestephenmarshall: I don't think this does anything right?

app/pb_kits/playbook/pb_person_contact/_person_contact.html.erb
kskilt: I tried using `.present?` and it was not actually filtering. `.length == 0` is filtering properly, but I do agree its ugly if there was a cleaner method.

kskilt: So it is passing an array, so we were able to change this to `.empty?` which gave the same functionality and is much cleaner.

KFad11:
Hey @thestephenmarshall - We've got everything good to go for review on this one. Let us know if anything needs to change. We can always take the `contact_detail`s off the docs for now if that is still in limbo with how/if we want to implement that.

PR #430, opened 2019-11-06
Revert "Add contact_details prop to contact kit"

thestephenmarshall:
Reverts powerhome/playbook#408


jasoncypret:
@garettarrowood We need to chat about this. We need to revert people

jasoncypret:
This is when the props were added: https://github.com/powerhome/playbook/pull/369/files

PR #436, opened 2019-11-08
added dark mode to user

evalouderback:


app/pb_kits/playbook/pb_user/_user.html.erb
```diff
@@ -11,7 +11,7 @@
   <% end %>
   <%= content_tag(:div,
       class: "content_wrapper") do %>
-    <%= pb_rails("title", props: { text: object.name, size: object.title_size }) %>
+    <%= pb_rails("title", props: { text: object.name, size: object.title_size, dark: object.dark }) %>
     <%= pb_rails("body", props: {
       text: object.title,
       color: "light"
```
markiearnold: Body kit should also be the dark version

app/pb_kits/playbook/pb_user/docs/example.yml
```diff
@@ -5,7 +5,10 @@ examples:
   - user_text_only: Text Only
   - user_size: Horizontal Size
   - user_vertical_size: Vertical Size
+  - user_dark: Dark
+
```
markiearnold: Above you also created the doc file "_user_text_only_dark.html.erb" - this should be listed here as well

app/pb_kits/playbook/pb_user/user.rb
```diff
@@ -37,6 +39,12 @@ def avatar_size
         end
       end
 
+      def dark_class
```
markiearnold: We no longer need the underscore in `_dark` when using the new syntax of `generate_classname()`. It handles that for use. Instead this can just be "dark".  See [Body Kit,  line 33](https://github.com/powerhome/playbook/blob/master/app/pb_kits/playbook/pb_body/body.rb#L33) for reference

PR #438, opened 2019-11-12
Date Stacked Kit Created

christinaatai:
Props:
- Date
- Size
- Alignment
- Reverse (month and day are flipped)
- Dark

Note: Year is displayed when it is not the current year
Screen Shot 2019-11-13 at 12 57 37 PM


app/pb_kits/playbook/packs/examples.js
```diff
@@ -113,3 +113,6 @@ WebpackerReact.setup (Toggle);
 WebpackerReact.setup (User);
 WebpackerReact.setup (UserBadge);
 WebpackerReact.setup (vertical_nav);
+
+import * as DateStacked from "pb_date_stacked/docs";
+WebpackerReact.setup (DateStacked);
```
drborges: Talking to @xjunior yesterday, we're not sure if `WebpackerReact.setup` here is either correct or even required.

As per [webpacker-react's Readme](https://github.com/renchap/webpacker-react#usage) the `setup` method expects an object that describes the component being registered, alternatively one could do:

```js
WebpackerReact.setup ({ DateStacked });
```

With that said, [it appears that Playbook no longer relies on webpacker-react to render react components](https://github.com/powerhome/playbook/blob/master/app/helpers/playbook/pb_kit_helper.rb#L19:L19) on Rails views, in which case we should remove webpacker-react from playbook which would simplify quick a bit how packs are setup. Thoughts @thestephenmarshall? cc @xjunior

drborges: hum... perhaps `WebpackerReact.setup({ DateStacked })` **is** required... In which case we should just fix the `setup` statements on the `examples.js` file.

xjunior: We stand corrected, @drborges. This seem to be exporting the examples of each component, that happen to be in react. I still think we can do it better:

1) change `examples.js` to just export the kits;
2) convert kit's examples to use `pb_react` to render the examples instead of pure react;
3) remove individual exports of each kit (currently not being used at all).

thestephenmarshall: @drborges @xjunior - Great points. We will have to address that in a different PR however.

thestephenmarshall: I created an issue to track this fwiw https://github.com/powerhome/playbook/issues/440

thestephenmarshall:
@christinaatai before you are done, will you add a screenshot of the new kit to the PR description pls?

PR #446, opened 2019-11-14
Home Tour React Kits - Support branch

markiearnold:
Changes:

### Button Kit
* Enhanced React Button kit to accept custom `id` prop.  (Needed to override styles in Circle Icon Button Kit)

---

### Circle Icon Button Kit: React
* Changed prop type to Object for button
* Added container div to use classnames - so there wasn't an "undefined" class rendered when `className` prop was not passed
* If `text` prop is passed in for Button, make it null, so it will use icon instead

---

### Circle Icon Button Kit: SCSS
* Added id to Button so `!important` tags were not needed
* Cleaned up styles that were not needed (Object-fit:cover, etc)
* Removed styles for icon, this should be taken care of in button, not needed. Targeting `i`, `svg` would also break button's loading icon

---

### Circle Icon Button Kit: Ruby Class
* Made `button` prop required
* Made `icon` prop required
* Removed `private` since there are no private methods

---

### Circle Icon Button Kit: Docs
* Made docs consistent for both Rails and React
* Removed the prop `text` from button, since it is suppose to use icon passed in

---

### Circle Icon Button Kit: Spec Test
* Created spec test for Circle Icon Button kit

---
---

### Card Kit: SCSS
* Clean up mixins 
* Add variables

### Selectable Card Kit: Ruby Class
* Modernized Kit

### Selectable Card Kit: Rails Partial
* Optimize html
* Make sure that label for relates to the id field of input

### Selectable Card Kit: React Kit
* Bring in prop helpers from Toggle
* Make sure event triggers for inputs onChange events
* Icon should only appear if true
* Lots of cleanup and restructure

### Selectable Card Kit: SCSS
* Cleanup
* Make sure all is using variables
* Fix Dark version styles
* Add transitions to be consistent with Card transitions for selectable border

### Selectable Card Kit: Docs
* Make all docs consistent between rails and react
* React docs contain state, make sure we have a working example for both multi and single

### Selectable Card Kit: Spec Test
* Created spec test for kit


jasoncypret:
@jasperfurniss @megantrimble We should get this merged ASAP!

PR #454, opened 2019-11-18
Enhance Layout sidebar colors

markiearnold:
Allow Layout to have a Gradient Sidebar
* Remove prop `dark` and add enum prop: `variant` that accepts light, dark, gradient.
* Cleanup docs to remove default `transparent: false`
* Cleanup docs to have clear examples.
* Convert Layout.jsx to newest jsx syntax, but **wait** on SubComponent Structure


markiearnold:
Currently, Layout Kit has 2 child components (Body, Sidebar), which both take children.  For react, how should we structure this kit, and how should the import/export work in the current index.js?

Referencing the jsx of a similar kit, Card which has a child component (Body) - I found that the child component was combined with the the parent in the [React kit](https://github.com/powerhome/playbook/blob/master/app/pb_kits/playbook/pb_card/_card.jsx). I do not think this approach will work with Layout due to both child components taking children. Any direction or ideas on this one? @thestephenmarshall @drborges @xjunior

app/pb_kits/playbook/pb_layout/_layout.jsx
```diff
@@ -1,54 +1,56 @@
...
+}: LayoutProps) => {
+  const ariaProps = buildAriaProps(aria)
+  const dataProps = buildDataProps(data)
 
-export default Layout;
+  const collapse_class = 'layout'+position_class+'_collapse_'+collapse;
+  const css = buildCss({
+    'pb_layout_kit': true,
+    [size]: true,
+    [position]: true,
```
thestephenmarshall: Clever! 💯

PR #456, opened 2019-11-19
User Kit (React)

christinaatai:


app/pb_kits/playbook/pb_user/_user.jsx
```diff
@@ -1,21 +1,66 @@
...
+
+const User = (props: UserProps) => {
+  const {
+    name='Anna Black',
+    title='',
+    align='left',
+    orientation='horizontal',
+    size='sm',
+    avatar=false,
+    avatar_url,
```
drborges: We should stick to camelcase style on JS code over snake-case.

PR #463, opened 2019-11-25
User kit, Territory Prop Added

christinaatai:


app/pb_kits/playbook/pb_user/_user.jsx
```diff
@@ -52,12 +54,25 @@ const User = (props: UserProps) => {
     }
   }
 
+  const print_territory = (territory) => {
+    if (territory != "") {
```
jasperfurniss: I think this needs to be !==  , but it might even need to be the triple equals  (!===) to have strict equality. Not sure if territory will ever be anything other than a string. Thoughts? @thestephenmarshall

app/pb_kits/playbook/pb_user/user.rb
```diff
@@ -40,6 +41,13 @@ def avatar_size
       def title_size
         size == "lg" ? 3 : 4
       end
+
+      def show_territory
+        if !territory.blank?
```
terryfinn: We can make this more straight forward using the inverse of ‘blank?’.

```suggestion
        if territory.present?
```

app/pb_kits/playbook/pb_user/_user.jsx
thestephenmarshall: `!==` will do. What I would also suggest here is to set the default up top in the argument like so:

```jsx
const print_territory = (territory="") => {
```

wadewinningham:
This this crossing the line into Nitro/Power-specific terminology?

app/pb_kits/playbook/pb_user/user.rb
```diff
@@ -40,6 +41,13 @@ def avatar_size
       def title_size
         size == "lg" ? 3 : 4
       end
+
+      def show_territory
```
terryfinn: I would consider making this a little more declarative and push more of the logic into this object and pull it out of the view. Maybe name this details or territory_and_title.

``` ruby
[territory, title].compact.join(“•”)
```

I would also consider using a unordered list and css to show the “•”.

christinaatai: What are the benefit(s) of using unordered list vs. just using " • "? @terryfinn

jasoncypret:
> This this crossing the line into Nitro/Power-specific terminology?

Would you make this something more generic? Location?

christinaatai:
> Would you make this something more generic? Location?

Location might be too generic as it may be confused with branches. Maybe userTerritory?  @wadewinningham @jasoncypret

wadewinningham:
I'm not opposed to *Territory* as that's a term other businesses frequently have as well. But the way we use it vs. how other companies use that info could differ. If we want to open source this, we have to plan for flexibility.

In this PR's context, *Territory* may be something we can offer as it's generic enough, but by displaying it in a format specifically for us like `"#{object.show_territory}#{object.title}"` isn't necessarily how someone else would want to show it.

I'm just putting this out there as a discussion point so we know where that line is.

* Does *Territory* have context outside our organization to others who may use playbook as an open source project.
* We're displaying territory and title a specific way. Is that okay for now? Or do we want to add some flexibility with how we display that detail?

As we get into more specific stuff, like user titles, etc, those types of questions are ones we should be asking ourselves along the way.

app/pb_kits/playbook/pb_user/user.rb
terryfinn: You avoid the delimiter from being text nodes like the other values. That means they will not be copyable, and they will not be read by a screen reader.

christinaatai:
@wadewinningham @jasoncypret I changed it so that it's details`"#{object.details}"`, where details consist of both territory and title. Do you think this is more clear?

wadewinningham:
I'm fine with that. I still leave final word on that with @jasoncypret

app/pb_kits/playbook/pb_user/user.rb
```diff
@@ -40,6 +41,14 @@ def avatar_size
...
         size == "lg" ? 3 : 4
       end
+
+      def details
+        if territory.present?
+          [territory, title].compact.join(" • ")
+        else
+          title
+        end
+      end
```
terryfinn: You don’t need the constructional. It just needs to be the compact join line.

```ruby
def details
  [territory, title].compact.join(" • ")
end
```

gmfvpereira: Good point, just make sure that the `territory` will not be `""` otherwise you can get to a weird scenario:
```
[1] pry(main)> territory = nil
=> nil
[2] pry(main)> title = "developer"
=> "developer"
[3] pry(main)> [territory, title].compact.join("  - ")
=> "developer"
[4] pry(main)> territory = ""
=> ""
[5] pry(main)> [territory, title].compact.join("  - ")
=> "  - developer"
```

@terryfinn what do you mean by "constructional"?

terryfinn: Auto correct mistake, not "constructional".

Meant to say, you don't need the conditional (`if territory.present?`).

Looks like from you example the value for territory and title are not nil, so it would need to be updated to:

```ruby
def details
  [territory, title]reject(&:blank?).join(" • ")
end
```

gmfvpereira: Got it! Now it makes sense!

PR #466, opened 2019-11-29
Create React version of Playbook Kit for IconCircle

Annihilous:


app/pb_kits/playbook/pb_icon_circle/_icon_circle.jsx
```diff
@@ -1,15 +1,25 @@
...
 
 class IconCircle extends React.Component {
   render() {
+    const { icon, size, variant } = this.props;
+
     return (
-      <h1>{`Coming Soon...`}</h1>
+      <div className={`pb_icon_circle_kit_${size}_${variant}`}>
+        <Icon icon={icon} />
+      </div>
```
thestephenmarshall: 1. Since we are including `flow` and that is the standard, we should use it instead of continuing to use `prop-types`.

1. Since no state is needed here, could we just convert this one to a function like so?:

```jsx
const IconCircle = ({
  icon,
  size='md',
  variant='default',
}: IconCircleProps) => (
  ...JSX HERE
)
```

Also, notice the defaults for `size` and `variant` are taken from the same defaults [here](https://github.com/powerhome/playbook/blob/master/app/pb_kits/playbook/pb_icon_circle/icon_circle.rb#L13) and [here](https://github.com/powerhome/playbook/blob/master/app/pb_kits/playbook/pb_icon_circle/icon_circle.rb#L16).

PR #473, opened 2019-12-04
Add jsx support for Date kit

Annihilous:
SSL

### DIFF
--- a/app/pb_kits/playbook/pb_date/_date.jsx
+++ b/app/pb_kits/playbook/pb_date/_date.jsx
@@ -1,21 +1,55 @@
+/* @flow */
+
 import React from 'react';
-import PropTypes from "prop-types";
+import DateTime from '../pb_kit/dateTime.js'
+import moment from 'moment'
+import {Icon} from '../'
 
-const propTypes = {
-  className: PropTypes.string,
-  id: PropTypes.string
-};
-
-class Date extends React.Component {
-  render() {
-    return (
-      <div className="pb_date">
-        <span>DATE CONTENT</span>
-      </div>
-    )
-  }
+type DateProps = {
+  size?: 'xs' | 'sm' | 'lg',
+  value?: String,
 }
 
-Date.propTypes = propTypes;
 
-export default Date;
+const defaultDateString = value => {
+  const weekday = value.toWeekday().toUpperCase()
+  const month = value.toMonth().toUpperCase()
+  const day = value.toDay()
+
+  return `${weekday} · ${month} ${day}`
+}
+
+const largeDateString = value => {
+  const month = value.toMonth().toUpperCase()
+  const day = value.toDay()
+
+  return `${month} ${day}`
+}
+
+
+const ExtraSmallDate = ({ value }) => (
+  <h3 className='pb_title_kit_4'>{defaultDateString(value)}</h3>
+)
+
+const SmallDate = ({ value }) => (
+  <h3 className='pb_title_kit_4'><Icon icon="calendar" fixedWidth="true" />{defaultDateString(value)}</h3>
+)
+
+const LargeDate = ({ value }) => (
+  <h3 className='pb_title_kit_3'>{largeDateString(value)}</h3>
+)
+
+
+export default function Date(props: DateProps) {
+  const {
+    size,
+    value,
+  } = props
+
+  const dateTimeValue = new DateTime({ value: value });
+
+  if (size == 'xs') return <ExtraSmallDate value={dateTimeValue} />
+  if (size == 'lg') return <LargeDate value={dateTimeValue} />
+
+  return <SmallDate value={dateTimeValue} />
+}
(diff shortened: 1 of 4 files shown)


app/pb_kits/playbook/pb_date/_date.jsx
```diff
@@ -1,21 +1,59 @@
...
+  const {
+    size,
+    value,
+  } = props
+
+  const calendarIcon = function(size, value) {
+    const dateTimeValue = new DateTime({ value: value });
+
+    const monthAndDate = dateTimeValue.toMonth().toUpperCase().toString() + ' ' + dateTimeValue.toDay().toString()
+    const weekdayAndDot = moment(value).format('ddd').toUpperCase() + ' · '
```
stephenagreer: It may be worth adding these two lines as functions in `dateTime.js`. That would also eliminate the need for the `require` statements at the top.

app/pb_kits/playbook/pb_date/_date.jsx
```diff
@@ -1,21 +1,59 @@
...
+
+    const defaultString = weekdayAndDot + monthAndDate
+
+    console.log(dateTimeValue);
```
gmfvpereira: Can you remove this `log`?

Annihilous: Yes! Thank you

app/pb_kits/playbook/pb_date/_date.jsx
Annihilous: Do you think we'll want to use these strings somewhere else? Or are you suggesting the move just to make the code more organized?

app/pb_kits/playbook/pb_date/_date.jsx
```diff
@@ -1,21 +1,59 @@
...
-    return (
-      <div className="pb_date">
-        <span>DATE CONTENT</span>
-      </div>
-    )
-  }
+import DateTime from '../pb_kit/dateTime.js'
+import moment from 'moment'
+require('moment-strftime')
+require('moment-timezone')
```
drborges: Let's not mix `require` and `import` the above can be rewritten as:

```jsx
import "moment-strftime"
import "moment-timezone"
```

app/pb_kits/playbook/pb_date/_date.jsx
```diff
@@ -1,21 +1,59 @@
...
+
+import {
+  Body,
+  Icon,
+} from '../'
+
+type DateProps = {
+  className: String,
```
drborges: It seems like the `className` prop is never used in the component's implementation.

app/pb_kits/playbook/pb_date/_date.jsx
```diff
@@ -1,21 +1,59 @@
...
-
-class Date extends React.Component {
-  render() {
-    return (
-      <div className="pb_date">
-        <span>DATE CONTENT</span>
-      </div>
-    )
-  }
+import DateTime from '../pb_kit/dateTime.js'
```
drborges: Why do you need `DateTime`?

Annihilous: I'm using DateTime to build an object for `dateTimeValue`. I tried deleting this line to see if it was required, and the date string wont display without it. Is there a better way?

drborges: I believe that with the proposal above, we would not need `DateTime`, `moment.js` can handle all date manipulation we need.

gmfvpereira:
I like @drborges suggestion. 
In my opinion, it's not a HUGE improvement in this case because the component itself is not a big one, but it's a good pattern to follow for sure. Even more if by any chance this component increases its size in the future.

app/pb_kits/playbook/pb_date/_date.jsx
thestephenmarshall: @drborges have you guys considered a transition to something like [date-fns](https://date-fns.org/) or some other suggestion from [here](https://github.com/you-dont-need/You-Dont-Need-Momentjs)? (I realize this might not be the place for that discussion)

Just curious 🤔

app/pb_kits/playbook/pb_date/_date.jsx
thestephenmarshall: [I think we should avoid `moment` in PB](https://github.com/powerhome/playbook/pull/473#discussion_r354384331)

drborges: Interesting. The reason we chose `moment` is because that [seemed to be the "norm"](https://github.com/powerhome/playbook/blob/master/app/pb_kits/playbook/pb_kit/dateTime.js) in PB.

Is it the plan to migrate to `date-fns`? If so we can look into replacing the moment implementation with `date-fns`.

drborges: nvm, just saw your comment below. @Annihilous wanna give that a shot?

drborges: @thestephenmarshall have you guys looked into [dayjs](https://github.com/iamkun/dayjs)?

PR #476, opened 2019-12-05
Text Input: Fixed SCSS name & Added classname support

kre8sions:
Renamed the input SCSS styles from pb_text_input_kit to text_input
Removed classname from input element and added object.classname to container


app/pb_kits/playbook/pb_text_input/_text_input.html.erb
```diff
@@ -1,21 +1,17 @@
...
+  <%= content_tag(:div,
+      class: "text_input_wrapper") do %>
+    <%= tag(:input,
+        id: object.id,
+        data: object.data,
+        class: "text_input",
+        name: object.name,
+        placeholder: object.placeholder,
+        type: object.type,
+        value: object.value) %>
```
jasoncypret: This looks like it's missing the children

app/pb_kits/playbook/pb_text_input/_text_input.scss
```diff
@@ -8,51 +8,52 @@
...
+    border-radius: $border-rad-light;
+    display: block;
+    width: 100%;
+    color: $text_lt_default;
+    background: $white;
+    padding: $space_xs $space_sm;
+    transition: 0.2s all;
+    outline: none;
+    font-size: $font_small;
+    margin: .1rem 0 0;
```
jasoncypret: If you pass a block input it won't have this class. You will likely have a similar pattern as Terry made

PR #477, opened 2019-12-05
Fix Dashboard Value Kit

### LINKED ISSUE #470: Dashboard Value is Broken
The most recent version of the dashboard value was modernized. However, this modernization enforced value to be a number. We have several prototypes and live code in Nitro that passes a string. Looking at the use cases I feel this kit should take a string not a number. The currency kit should enforce a number.

https://github.com/powerhome/playbook/tree/master/app/pb_kits/playbook/pb_dashboard_value

thestephenmarshall:
Fixes #470

### DIFF
--- a/app/pb_kits/playbook/pb_dashboard_value/_dashboard_value.html.erb
+++ b/app/pb_kits/playbook/pb_dashboard_value/_dashboard_value.html.erb
@@ -7,7 +7,7 @@
   <% end %>
 
   <% if object.stat_value.present? %>
-    <%= pb_rails("stat_value", props: object.stat_value) %>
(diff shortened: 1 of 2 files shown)


app/pb_kits/playbook/pb_dashboard_value/docs/_dashboard_value_default.html.erb
```diff
@@ -1,21 +1,21 @@
 <%= pb_rails("dashboard_value", props: {
   stat_label: "Decreased Value",
-  stat_value: {value: 1428, unit: "appts"},
+  stat_value: {value: "1428", unit: "appts"},
```
thestephenmarshall: Only updated this doc example for testing purposes.

PR #478, opened 2019-12-05
Select Kit

jasoncypret:
image


app/pb_kits/playbook/pb_select/docs/nitro_theme.gemspec
```diff
@@ -0,0 +1,41 @@
+$LOAD_PATH.push File.expand_path("lib", __dir__)
```
gmfvpereira: What's the intent of this file here?

PR #481, opened 2019-12-05
Fix form kit dependencies

terryfinn:
Make sure that simple form will be available in the production mode
of the playbook web app. This is needed so we can show the example of
how to use playbook with simple form. The previous code only had a
development dependency defined in the gemspec, moving this to the
Gemfile will make it available in the production web app.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,3 +9,4 @@ end
 gemspec
 
 gem "health_check"
+gem "simple_form", ">= 5.0.1", "< 6.0.0"


Gemfile
```diff
@@ -9,3 +9,4 @@ end
 gemspec
 
 gem "health_check"
+gem "simple_form", ">= 5.0.1", "< 6.0.0"
```
wadewinningham: There a Gemfile.lock that needs to be updated?

terryfinn: @wadewinningham No, this was already in the Gemfile.lock from the gemspec.

Issue #483, opened 2019-12-05
Enable ESLint Completely


thestephenmarshall:
This PR was opened recently: https://github.com/powerhome/playbook/pull/482 which would enable ESLint for all of Playbook. ESLint would run during builds and provide helpful insight into warnings and errors on top of what has already been enabled in overcommit.yml for commits. 

As mentioned in the PR comments, we should phase this rollout and likely do the same for Huddle as well. Thanks to @rafbgarcia for kicking this off!

rafbgarcia:
Milestone

- [ ] **STEP 1** Create a PR that updates only the config changes .eslintrc.json, ect and do not make it retroactive #484
- [ ] **STEP 2** Fix the non-kit JS warnings and errors with the above ^ #484
- [ ] **STEP 3** Incrementally update each kit (or perhaps groups of kits) to remove lint errors and warnings - commits should be kit focused, so one commit per kit. While these changes are only fixing lint errors, we all know that if even the slightest possibility exists for a chance to break a kit then we want to be able to at least track the change down quickly. Breaking this out into kits allows for that.

gmfvpereira:
I think it should be safe to have a `STEP 2.1` PR where we change all occurrences that are "auto fixed" by eslint. Do you see any drawbacks to this approach?

rafbgarcia:
@gmfvpereira https://github.com/powerhome/playbook/pull/484 does that for all non-kit files.

gmfvpereira:
What's the problem with doing the auto fix even for the kits?

rafbgarcia:
The only issue I see is possible conflicts with the PRs currently opened. Maybe @thestephenmarshall has other concerns.

PR #482, opened 2019-12-05
Enable eslint

rafbgarcia:
Right now on `master` we have a lot of lint offenses:

rafa_Rafael____power_playbook

### This PR:

- Cleans up all lint offenses;

rafa_Rafael____power_playbook

- Adds some lints to guarantee code standards (I've added the lints that I think are good, but this is flexible, of course, so feedback is welcome);
- Runs eslint on CI.

PR #484, opened 2019-12-06
Update eslint and non kit files

rafbgarcia:
On this PR are steps 1 and 2 of this issue https://github.com/powerhome/playbook/issues/483

- Updated `.eslint` to include the same lints added here https://github.com/powerhome/playbook/pull/482
- Sorted them out
- Fixed offenses for files that are not in the pattern below
- Updated generators to avoid lint offenses on new kits

rafa_Rafael____power_playbook_app_pb_kits_playbook


lib/generators/kit/templates/kit_jsx.erb.tt
```diff
@@ -1,21 +1,39 @@
+<%
```
rafbgarcia: Generated file:
_rafa_jsx_—_playbook

config/webpacker.yml
```diff
@@ -7,7 +7,7 @@ default: &default
   public_output_path: packs
   cache_path: tmp/cache/webpacker
   check_yarn_integrity: false
-  webpack_compile_output: false
+  webpack_compile_output: true
```
thestephenmarshall: We will want to test this setting out again since the reason it was disabled before now eludes me. Likely was related to slow startup which might not be an issue now..

startup.sh
```diff
@@ -1,11 +1,11 @@
 if [ $# -eq 0 ] ; then
   echo '=========== Precompiling'
-  cd spec/dummy; RAILS_ENV=production SECRET_KEY_BASE=does_not_matter_here bin/rails assets:precompile
+  cd spec/dummy; RAILS_ENV=production SECRET_KEY_BASE=does_not_matter_here bin/rails assets:precompile --trace
```
thestephenmarshall: This is another one that I will need to test out. Don't think it will cause any issues though.

config/webpacker.yml
rafbgarcia: @thestephenmarshall if it helps, I set this to true to see CI compilation errors, there's a thread about it (https://github.com/rails/webpacker/issues/955) and they've decided to set it to `true` by default on a next version https://github.com/rails/webpacker/pull/2316

startup.sh
rafbgarcia: Sorry, I don't think we need this, I was trying to see CI errors but I think the `webpack_compile_output ` is enough, let me remove it.

config/webpacker.yml
thestephenmarshall: Okay then if that is the case let's set it to true as well.

startup.sh
rafbgarcia: Done

PR #489, opened 2019-12-06
Eslint is a dev dependency

gmfvpereira:
The eslint version used here depends on a newer version of node when comparing to nitro-web.
Moving this dep to the `devDependencies` is a short term solution to allow playbook_ui to be used with nitro-web.

Long term solution should be to upgrade nitro node version since [it's not gonna be supported anymore](https://nodejs.org/en/about/releases/).

We tested this pointing my nitro-web to my playbook on my local machine and it worked fine. Is there any other test I should do?

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,9 @@
     "eslint-plugin-react": "^7.16.0",
     "extract-text-webpack-plugin": "3.0.0",
     "flow-bin": "0.54.1",
-    "webpack-dev-server": "^3.7.2"
+    "webpack-dev-server": "^3.7.2",
+    "eslint": "^6.6.0",
+    "eslint-plugin-flowtype": "^4.4.1"
   },
   "dependencies": {
     "@babel/cli": "7.2.3",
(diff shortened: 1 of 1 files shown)


thestephenmarshall:
@gmfvpereira I just pulled this down and switched to `8.9.4` and got this:

Perhaps I am missing something..

```sh
➜  playbook git:(fix-eslint-version) yarn 
yarn install v1.13.0
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
info fsevents@1.2.7: The platform "linux" is incompatible with this module.
info "fsevents@1.2.7" is an optional dependency and failed compatibility check. Excluding it from installation.
error eslint@6.6.0: The engine "node" is incompatible with this module. Expected version "^8.10.0 || ^10.13.0 || >=11.10.1". Got "8.9.4"
error Found incompatible module
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
➜  playbook git:(fix-eslint-version) node -v
v8.9.4
```

gmfvpereira:
@thestephenmarshall I'll need your help to release this as 3.0.1

PR #493, opened 2019-12-09
Kit: Flex

markiearnold:
# Flex Kit 
**PR includes Rails only**

## Demo: How to use Kit
Playgrounds Branch ([dedemo/how-to-use-flex-kit](https://github.com/powerhome/playgrounds/pull/66))

## What does it do?
Flex kit has simple props to control the container with flexbox css properties. Creates consistency between row and column alignment options with a simple props "vertical" and "horizontal".

**PROPS:**
* Horizontal - Enum to control both `row` and `column` horizontal alignment
* Vertical - Enum to control both `row` and `column` vertical alignment
* Inline - Boolean to set `display: flex` vs `display: inline-flex`
* Orientation - Enum to set `row` or `column`
* Spacing - Enum to set spacing of child elements to `none`, `around`, `between`, or `evenly`
* Reverse - Boolean to reverse order of child elements, but keep alignment settings
* Wrap - Boolean to set `flex-wrap: wrap` vs `flex-wrap: nowrap`

---

Includes optional child kit, "Flex Item"

Flex Item child kit is a simple, *optional* wrapper for child elements of the Flex kit. 

* Fixed Size - String to set a fixed size of the child (Examples: "100px", "50vh", "33%"). Forces child to stay that size by setting width and flex-basis of the element.
* Grow - Boolean to set `flex-grow: 0` vs `flex-grow: 1`


app/pb_kits/playbook/packs/site_styles/docs/_flex_examples.scss
```diff
@@ -0,0 +1,10 @@
+  [class^=pb_flex_kit] { background: #F3F7FB; }
+  .flex-item,
+  [class^=pb_flex_item_kit] { background: #E4E8F0; min-width: 30px; min-height: 30px; color: #242B42; }
+  .tall {min-height: 200px;}
+
+  // [class^=pb_flex_kit][class*=wide]{
+  //     [class^=pb_flex_item_kit]{
+  //       flex-basis: 300px;
+  //     }
+  // }
```
thestephenmarshall: Do we need to keep this around?

app/pb_kits/playbook/pb_flex/_flex.jsx
```diff
@@ -0,0 +1,48 @@
...
+const Flex = ({
+  children,
+  className,
+  inline,
+  horizontal,
+  orientation,
+  spacing,
+  reverse,
+  vertical,
+  wrap,
```
thestephenmarshall: Let's mirror any of the defaults from `flex.rb` here

PR #494, opened 2019-12-10
Text Input (Dark Mode & Updates)

megantrimble:
-Added Dark version of Text Input
-Revised Text Input to use Textarea mixins and follow Sketch design
-Moved over OnChange function from Home Tour React Kits

~~*Currently experiencing an eslint error that needs to be fixed.*~~

### DIFF
--- a/app/pb_kits/playbook/pb_text_input/_text_input.html.erb
+++ b/app/pb_kits/playbook/pb_text_input/_text_input.html.erb
@@ -1,12 +1,12 @@
 <%= content_tag(:div,
     class: object.classname) do %>
   <% if object.label.present? %>
-    <%= pb_rails("caption", props: {text: object.label}) %>
+    <%= pb_rails("caption", props: {text: object.label, dark: object.dark}) %>
   <% end %>
   <%= content_tag(:div,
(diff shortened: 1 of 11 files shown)


app/pb_kits/playbook/pb_text_input/_text_input.html.erb
```diff
@@ -1,12 +1,12 @@
 <%= content_tag(:div,
     class: object.classname) do %>
   <% if object.label.present? %>
-    <%= pb_rails("caption", props: {text: object.label}) %>
+    <%= pb_rails("caption", props: {text: object.label, dark: object.dark}) %>
```
terryfinn: I like that is is automatically passing the dark option down into the caption kit. Any thoughts of adding support for this to the form kit, so we could specify the form as being a dark mode variant, and having this option to automatically pushed down into the text input kit and button kit?

jasoncypret: Love it @terryfinn . We will start a PR for that

jasoncypret:
@megantrimble can you add a changelog?

PR #499, opened 2019-12-10
Display wrong contacts correctly

KFad11:
Currently, the kit wasn't displaying the wrong number caption, wasn't arranging the contacts with valid contacts above incorrect contacts, nor documenting this functionality in the kit. 

This PR adds that documentation referenced above, it detects valid contacts, then displays them above a wrong number caption and displays the wrong numbers beneath with the correct icon. 

**This in no way is a backward breaking change to the Nitro repo, as it isn't implemented there yet.**

Screen Shot 2019-12-11 at 11 35 03 AM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Sets image build status for Milano compatibility ([#472][] @viniciusgama)
 - Add node binding for node-sass ([#465][] @thestephenmarshall)
 - Enable ESLint in overcommit ([#462][] @thestephenmarshall)
+- Display wrong numbers in Person Contact Kit ([#499][] @KFad11)
 
 [#436]: https://github.com/powerhome/playbook/pull/436
 [#479]: https://github.com/powerhome/playbook/pull/479
@@ -74,6 +75,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#472]: https://github.com/powerhome/playbook/pull/472
 [#465]: https://github.com/powerhome/playbook/pull/465
 [#462]: https://github.com/powerhome/playbook/pull/462
+[#499]: https://github.com/powerhome/playbook/pull/499
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Display wrong numbers in Person Contact Kit ([#499][] @KFad11)
+- Display wrong numbers in Person Contact Kit. The Kit will now display a caption "wrong number" and have the wrong numbers listed beneath the caption ([#499][] @KFad11)
 
 [#499]: https://github.com/powerhome/playbook/pull/499
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,16 +15,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#498]: https://github.com/powerhome/playbook/pull/498
 [#439]: https://github.com/powerhome/playbook/pull/439
 
+### Fixed
+
+- Display wrong numbers in Person Contact Kit ([#499][] @KFad11)
+
+[#499]: https://github.com/powerhome/playbook/pull/499
+
 
 ## [3.0.1] - 2019-12-06
 
 ### Fixed
 
 - Move `eslint` to devDependencies ([#489][] @gmfvpereira)
-- Display wrong numbers in Person Contact Kit ([#499][] @KFad11)
 
 [#489]: https://github.com/powerhome/playbook/pull/489
-[#499]: https://github.com/powerhome/playbook/pull/499
 
 ## [3.0.0] - 2019-12-05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,8 +21,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Move `eslint` to devDependencies ([#489][] @gmfvpereira)
+- Display wrong numbers in Person Contact Kit ([#499][] @KFad11)
 
 [#489]: https://github.com/powerhome/playbook/pull/489
+[#499]: https://github.com/powerhome/playbook/pull/499
 
 ## [3.0.0] - 2019-12-05
 
@@ -63,7 +65,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Sets image build status for Milano compatibility ([#472][] @viniciusgama)
 - Add node binding for node-sass ([#465][] @thestephenmarshall)
 - Enable ESLint in overcommit ([#462][] @thestephenmarshall)
-- Display wrong numbers in Person Contact Kit ([#499][] @KFad11)
 
 [#436]: https://github.com/powerhome/playbook/pull/436
 [#479]: https://github.com/powerhome/playbook/pull/479
@@ -75,7 +76,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#472]: https://github.com/powerhome/playbook/pull/472
 [#465]: https://github.com/powerhome/playbook/pull/465
 [#462]: https://github.com/powerhome/playbook/pull/462
-[#499]: https://github.com/powerhome/playbook/pull/499
 
 ### Fixed
 
--- a/app/pb_kits/playbook/pb_person_contact/_person_contact.html.erb
+++ b/app/pb_kits/playbook/pb_person_contact/_person_contact.html.erb
@@ -14,7 +14,7 @@
     }) %>
   <% end %>
   <% unless object.wrong_contacts.empty? %>
-    <%= pb_rails("caption", props: { text: "wrong number" }) %>
+    <div style="padding-top: 10px"> <%= pb_rails("caption", props: { text: "wrong number" }) %></div>
 
     <% object.wrong_contacts.each do |wrong_number| %>
       <%= pb_rails("contact", props: {
--- a/app/pb_kits/playbook/pb_person_contact/_person_contact.html.erb
+++ b/app/pb_kits/playbook/pb_person_contact/_person_contact.html.erb
@@ -14,7 +14,7 @@
     }) %>
   <% end %>
   <% unless object.wrong_contacts.empty? %>
-    <div style="padding-top: 10px"> <%= pb_rails("caption", props: { text: "wrong number" }) %></div>
+    <div class="wrong_numbers"> <%= pb_rails("caption", props: { text: "wrong number" }) %></div>
 
     <% object.wrong_contacts.each do |wrong_number| %>
       <%= pb_rails("contact", props: {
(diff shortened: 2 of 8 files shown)


app/pb_kits/playbook/pb_person_contact/docs/example.yml
```diff
@@ -4,6 +4,7 @@ examples:
   - person_contact_default: Default
   - person_contact_multiple: Multiple People
   - person_contact_with_detail: With Detail
+  - person_contact_with_wrong_numbers: With Wrong Numbers
```
gmfvpereira: Do we need all those extra lines right above this one?

app/pb_kits/playbook/pb_person_contact/_person_contact.html.erb
```diff
@@ -1,25 +1,26 @@
 <%= content_tag(:div,
```
gmfvpereira: What do you mean by `This in no way is a backward breaking change to the Nitro repo, as it isn't implemented there yet.`? 

I do see this usage:
```
            <%= pb_rails("contact", props: {
              contact_type: "cell",
              contact_value: history.phone_number,
            }) %>
```
this is not using the changed param (wrong number), but I think that until we release this kit someone may use a  `contact_type = "wrong number"`. And that would configure a breaking change. Right? Or I am missing something?

app/pb_kits/playbook/pb_person_contact/docs/example.yml
KFad11: Line 7 adds the documentation to the Playbook UI for the example with the wrong numbers in the contact. Lines 4, 5 & 6 are different examples, showing that there are different options/props you can use with the kit.

app/pb_kits/playbook/pb_person_contact/_person_contact.html.erb
KFad11: The case statement from Line 21 in the `contact.rb` file states which options are available as `contact_type`s. If someone uses anything not currently on the list, the kit will default to the phone icon. So if they use `contact_type = "wrong number"` incorrectly, they'll just get a phone icon with no slash through it.

app/pb_kits/playbook/pb_person_contact/_person_contact.html.erb
```diff
@@ -1,25 +1,26 @@
...
+  <% object.valid_contacts.each do |contact| %>
+    <%= pb_rails("contact", props: {
+        contact_type: contact[:contact_type],
+        contact_value: contact[:contact_value],
+        contact_detail: contact[:contact_detail],
     }) %>
-    <% object.contacts.each do |contact| %>
+  <% end %>
+  <% unless object.wrong_contacts.empty? %>
+    <%= pb_rails("caption", props: { text: "wrong number" }) %>
```
jasoncypret: Looking at the screenshot (thanks for providing) I believe this caption needs some margin-top. Can you match it up with the mocks, and use the spacing variables to give it some separation? Let me know if you need help!

CHANGELOG.md
```diff
@@ -63,6 +63,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Sets image build status for Milano compatibility ([#472][] @viniciusgama)
 - Add node binding for node-sass ([#465][] @thestephenmarshall)
 - Enable ESLint in overcommit ([#462][] @thestephenmarshall)
+- Display wrong numbers in Person Contact Kit ([#499][] @KFad11)
```
rafbgarcia: Hey @KFad11 looks great, this should be moved to the `Unreleased` section though, under the `### Fixed` category.

KFad11:
The example now expects the `wrong-phone` `contact_type` to generate the correct `phone-slash` icon. Also if given an empty string or a type outside the defined list, the icon defaults to `phone`. Rspec goes green after adding them.

app/pb_kits/playbook/pb_person_contact/_person_contact.html.erb
KFad11: I threw the caption kit here into a `div` with some `padding-top: 10px`. Is that an OK way to do that, @jasoncypret ?

CHANGELOG.md
KFad11: Ah HAH! I get it now. Should be good to go.

CHANGELOG.md
```diff
@@ -21,8 +21,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Move `eslint` to devDependencies ([#489][] @gmfvpereira)
+- Display wrong numbers in Person Contact Kit ([#499][] @KFad11)
```
rafbgarcia: Almost there :) Now you're under the `## [3.0.1] - 2019-12-06` release section, you want to put it under the `## [Unreleased]` one.

I know it's not very obvious.

There's an `### Added` category there, you'll need to add a `### Fixed` one, like:

```
## [Unreleased]
### Added
- Introduce CHANGELOG file ([#498][] @rafbgarcia)
- Include Card Separator example in Card kit docs ([#439][] @rruiz85)
[#498]: https://github.com/powerhome/playbook/pull/498
[#439]: https://github.com/powerhome/playbook/pull/439

+ ### Fixed
+
+ - Display wrong numbers in Person Contact Kit ([#499][] @KFad11)
+
+ [#499]: https://github.com/powerhome/playbook/pull/499
```

Notice the `+` (plus) are your changes.

KFad11: So... close! lol. I find it way harder to read in my text editor. I should have given it a closer look in Github, then found my spot in the text editor.

jasoncypret: Also, can we be a bit more descriptive? The Ninjas will look at this to figure out what to test. What should they look for?

app/pb_kits/playbook/pb_person_contact/docs/example.yml
gmfvpereira: Wow, sorry I thought about something but typed other. I meant "the lines below".

CHANGELOG.md
KFad11: I think I was just looking for the first `### Fixed` and adding my fix to that list. I added the new `### Fixed`, right under the current unreleased `### Added` group.

app/pb_kits/playbook/pb_person_contact/_person_contact.html.erb
gmfvpereira: Ok, cool. So this is changing the API but you're handling the exception. Good call.

app/pb_kits/playbook/pb_person_contact/_person_contact.html.erb
```diff
@@ -1,25 +1,26 @@
...
+  <% object.valid_contacts.each do |contact| %>
+    <%= pb_rails("contact", props: {
+        contact_type: contact[:contact_type],
+        contact_value: contact[:contact_value],
+        contact_detail: contact[:contact_detail],
     }) %>
-    <% object.contacts.each do |contact| %>
+  <% end %>
+  <% unless object.wrong_contacts.empty? %>
+    <div style="padding-top: 10px"> <%= pb_rails("caption", props: { text: "wrong number" }) %></div>
```
jasoncypret: We never want to add inline styles. Please move this to the person_contact.scss https://github.com/powerhome/playbook/blob/master/app/pb_kits/playbook/pb_person_contact/_person_contact.scss

Also, use a spacing token for the padding value

app/pb_kits/playbook/pb_person_contact/docs/example.yml
KFad11: Not sure why the multiple spaces... looks like there isn't a solid convention used throughout to separate. I'm a fan of a single space, which is what I'll use here. FWIW, many kits have 1-2 spaces between, without a clear reason why.

app/pb_kits/playbook/pb_person_contact/_person_contact.html.erb
KFad11: `$space-xs` looked better than `$space-md`, let me know what you think, I'll update the screen shot.

app/pb_kits/playbook/pb_person_contact/_person_contact.html.erb
```diff
@@ -1,25 +1,26 @@
...
+  <% object.valid_contacts.each do |contact| %>
+    <%= pb_rails("contact", props: {
+        contact_type: contact[:contact_type],
+        contact_value: contact[:contact_value],
+        contact_detail: contact[:contact_detail],
     }) %>
-    <% object.contacts.each do |contact| %>
+  <% end %>
+  <% unless object.wrong_contacts.empty? %>
+    <div class="wrong_numbers"> <%= pb_rails("caption", props: { text: "wrong number" }) %></div>
```
terryfinn: Why wrap this in a div, and not add a class to the wrong number caption?

If we want Playbook to be generally useful to others, what are our thoughts on handling copy like "wrong number"? This might not be the way others want to describe this, and others might not be able to only use English. I would consider moving things like "wrong number" out of hard coded HTML and make it configurable. I think taking advantage of translations would a good approach. This line could be turned into:

```suggestion
    <div class="wrong_numbers"> <%= pb_rails("caption", props: { text: t("pb_kits/playbook/pb_person_contact/wrong_number") }) %></div>
```

Now others can define alternative language versions of this, and can redefine what should display for English.

Also, do we have a convention for custom css rules? I usually see classes written as "wrong-numbers".

app/pb_kits/playbook/pb_person_contact/_person_contact.html.erb
```diff
@@ -1,25 +1,26 @@
...
+  }) %>
+  <% object.valid_contacts.each do |contact| %>
+    <%= pb_rails("contact", props: {
+        contact_type: contact[:contact_type],
+        contact_value: contact[:contact_value],
+        contact_detail: contact[:contact_detail],
     }) %>
-    <% object.contacts.each do |contact| %>
+  <% end %>
+  <% unless object.wrong_contacts.empty? %>
```
terryfinn: We can probably make this a little clearer with some logic flipping.
```suggestion
  <% if object.wrong_contacts.present? %>
```

rafbgarcia: Agree with @terryfinn. I'm always in favor of the positive version instead of double negation.

app/pb_kits/playbook/pb_person_contact/_person_contact.html.erb
rafbgarcia: @terryfinn I'm new to the repo but I think if we want to go in this path, we need to establish it as a standard first and then start doing it.

Maybe adding a `CONTRIBUTING.md` file with such standards would be helpful so that everyone knows what guidelines they should be following.

Maybe a template for PRs with a checklist would also be helpful.

Thoughts?

KFad11: Totally forgot that we can pass that `classname` prop and get that style on there. Got that and the conditional issue below resolved in the latest commit. I like `if` better than `unless` as well.

KFad11: @terryfinn - As flexible and inclusive as the translations would be, I feel like that would be out of scope here for this PR. Solidifying a standard going forward, as well as a template like @rafbgarcia pointed out would be awesome too.

app/pb_kits/playbook/pb_person_contact/_person_contact.html.erb
KFad11: @jasoncypret I think this is in its final form as is... I would like to get it in the next version raise in Nitro tomorrow. I'm going to head out in a few minutes here but I could try to make any last-minute changes tonight so it could be included tomorrow's version raise. Thanks!

PR #509, opened 2019-12-11
Update `pb_release:version` task to include confirmation and versioning info

thestephenmarshall:
Upon running the command `bin/rails app:pb_release:version` you will now see:

Screen Shot 2019-12-11 at 2 28 13 PM


jasoncypret:
Changelog

PR #511, opened 2019-12-12
Add select to form kit

web-kat:
#### Runway Ticket URL
Tickets: [BOP-309]

#### Breaking Changes
n/a

## Form With Example
form_with_example

## Simple Form Example
simple_form_example

#### Checklist:


app/pb_kits/playbook/pb_select/_select.scss
```diff
@@ -3,7 +3,8 @@
 @import "../tokens/colors";
 
 [class^=pb_select] {
-  margin-bottom: $space_sm;
+  margin: 4px 4px $space_sm 4px;
```
web-kat: @jasoncypret we noticed that the select kit did not match the rest of the form inputs' margins, so we adjusted the margins here to match the rest of them. Thoughts on making this change?

jasoncypret: I think some of the PR's are working against each other, but we shouldn't be adding space on the top, left or right. We would like to enable spacing props to all kits in a future PR. For now, the plan is to only have margin-bottom defined.

app/pb_kits/playbook/pb_textarea/_textarea.scss
```diff
@@ -2,9 +2,9 @@
 @import "./textarea_mixin";
 
 [class^=pb_textarea_kit] {
+  margin: 8px;
```
jasoncypret: Let's revert 😉

app/pb_kits/playbook/pb_select/_select.scss
web-kat: I'll remove this

app/pb_kits/playbook/pb_textarea/_textarea.scss
web-kat: Yep, I'll rebase onto Roxanne's branch for this

PR #514, opened 2019-12-12
Update to version 3.1.0

thestephenmarshall:


rafbgarcia:
We need to update CHANGELOG too.

Thinking about this, if new PRs are merged before we update CHANGELOG, they'll add things to the `Unreleased` section and that may be hard to track.

rafbgarcia:
Never mind I saw you did that on another PR.

PR #521, opened 2019-12-16
Implement React version of IconValue kit

drborges:
#### Runway Ticket URL

Implement the React version of the icon value kit.

#### Beaking Changes

This PR **has no** breaking changes

Screen Shot 2019-12-16 at 2 57 29 PM

Screen Shot 2019-12-16 at 2 57 45 PM

#### Checklist:

### DIFF
--- a/app/pb_kits/playbook/pb_icon_value/_icon_value.jsx
+++ b/app/pb_kits/playbook/pb_icon_value/_icon_value.jsx
@@ -14,12 +14,12 @@ import {
 } from '../utilities/props'
 
 type IconValueProps = {
-  align: 'left' | 'center' | 'right',
-  aria: object,
-  className: string,
-  data: object,
+  align?: 'left' | 'center' | 'right',
+  aria?: object,
+  className?: string,
+  data?: object,
   icon: string,
-  id: number,
+  id?: number,
   text: string,
 }
 
--- a/app/pb_kits/playbook/pb_icon_value/_icon_value.jsx
+++ b/app/pb_kits/playbook/pb_icon_value/_icon_value.jsx
@@ -1,21 +1,57 @@
+/* @flow */
+
 import React from 'react'
-import PropTypes from 'prop-types'
+import classnames from 'classnames'
+import {
+  Icon,
+  Body,
+} from '../'
 
-const propTypes = {
-  className: PropTypes.string,
-  id: PropTypes.string,
+import {
+  buildAriaProps,
+  buildDataProps,
+  buildCss,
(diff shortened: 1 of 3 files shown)


app/pb_kits/playbook/pb_icon_value/_icon_value.html.erb
```diff
@@ -5,5 +5,6 @@
     <%= pb_rails("body", props: { color: "light" }) do %>
       <%= pb_rails("icon", props: { icon: object.icon, fixed_width: true }) %>
       <%= object.text %>
+
```
rafbgarcia: New line by accident?

app/pb_kits/playbook/pb_icon_value/_icon_value.jsx
```diff
@@ -1,21 +1,59 @@
...
+  icon,
+  id,
+  text,
+}: IconValueProps) => {
+  const ariaProps = buildAriaProps(aria)
+  const dataProps = buildDataProps(data)
+  const css = buildCss({
+    'pb_icon_value_kit': true,
+    [align]: true,
+  })
```
rafbgarcia: What do you think of updating `buildCss` implementation to the below? 

```js
export const buildCss = (...rules) => classnames(rules).replace(/\s/g, '_')
```

I did that on my PR because we don't actually use that `delimiter` param anywhere, which I think is good because we keep the standard for the CSS classes, so that param may be unnecessary after all.

If you did that you could do just 
```js
buildCss('pb_icon_value_kit', align)
```

drborges: Sounds like a nice improvement 👍 

I'll give that a shot in the morning, thanks @rafbgarcia 👊

app/pb_kits/playbook/pb_icon_value/_icon_value.html.erb
drborges: likely... thx

app/pb_kits/playbook/pb_icon_value/_icon_value.jsx
drborges: Actually... The current implementation should already allow for:

```jsx
buildCss('pb_icon_value_kit', align)
```

Am I missing something?

rafbgarcia: Yes, the way the function is currently implemented, `align` would be passed as `delimiter` and not joined with `'pb_icon_value_kit'`.

PR #523, opened 2019-12-16
Export missing components

gmfvpereira:
Export missing Kits


gmfvpereira:
Is there any other place that I need to change in order to export a jsx component?

PR #533, opened 2019-12-19
Run eslint autofix on pre commit hook

rafbgarcia:
rafa_Rafael____power_playbook

#### Runway Ticket URL
None

#### Breaking Changes
None

None.

#### Checklist:

### DIFF
--- a/.lintstagedrc
+++ b/.lintstagedrc
@@ -1,0 +1,3 @@
+{
+  "*.{js,jsx}": "yarn lint --fix"
+}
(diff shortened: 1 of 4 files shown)


drborges:
Here's another thought:

I've been using `vscode` with `eslint` (and `flow`) configured to run on every save and it's been working pretty well so far. Code formatting/styling is a no brainer now, I simply let the editor take care of styling things for me based on our rules. With that said, code styling should no longer be a code review "overhead" anymore since what you see in the PR is what eslint enforced.

I realize that not everyone uses `vscode` but it's most likely that every editor has an eslint plugin these days. This could be (IMO it should be) part of every dev's workflow. Thoughts?

rafbgarcia:
I agree and I think people could benefit from using these plugins to learn how to better organize their code as they'll be seeing those code changes in the editor.

I think that's harder to enforce though.

drborges:
Yeah, I would not say enforce but highly recommend people to rely on such workflow since that has the potential to reduce back&forth optimizing their time.

PR #536, opened 2019-12-23
Pb Card: Swap implementation of highlight position values.

drborges:
#### Runway Ticket URL

It seems that the highlight position values (`side` vs `top`) have their implementations inverted (see screenshots below).

This PR makes it so that setting the highlight position to `side`, causes the highlight to be actually rendered on the side of the card rather than on the top. Similarly, the highlight position set to `top` will render the highlight on the top of the card rather than on its side.

#### Breaking Changes

This PR **could** introduce some regression on apps using the `position` prop of the `Card` kit. As of this writing, it seems that at least `nitro-web` **would not** experience any changes to the behavior of card components since there's no usage instance of the position prop.

**Before**

**Rails** Version:

rails-before

**React** Version:

react-before

**After**

**Rails** Version:

rails-after

**React** Version:

react-after

#### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixed broken eslint ref + removed unused file ([#540][] @jasoncypret)
+- Swap implementation of highlight position values for Card Kit ([#536][] @drborges)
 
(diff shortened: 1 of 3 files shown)


drborges:
@terryfinn it does. I'm yet to do some more accurate investigation on the impact of the regression but it should impact any usage of the card component using the `highlight.position` prop.

PR #546, opened 2020-01-06
Fixed radio bug in docs

kre8sions:
#### Runway Ticket URL
https://nitro.powerhrg.com/runway/backlog_items/NUX-516

Screen Shot 2020-01-06 at 1 03 13 PM

#### Breaking Changes

#### Checklist:

### DIFF
--- a/app/pb_kits/playbook/pb_radio/_radio.jsx
+++ b/app/pb_kits/playbook/pb_radio/_radio.jsx
@@ -11,7 +11,7 @@ type RadioProps = {
   label: String,
   name: String,
   value: String,
-  checked?: Boolean,
+  defaultChecked?: Boolean,
   dark?: Boolean,
   text: String,
   children?: Node,
@@ -20,7 +20,7 @@ type RadioProps = {
 
 const Radio = (props: RadioProps) => {
   const {
-    checked = false,
+    defaultChecked = false,
     children,
     className,
     dark = false,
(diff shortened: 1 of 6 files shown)


jasoncypret:
@kre8sions this technically is a prop change and would warrant a major release. @thestephenmarshall Thoughts on this change? Do we need to change the prop name?

kre8sions:
@jasoncypret I'm not sure if the react version of this was even being used should we slate this for 4.0 or should we revert this change.

app/pb_kits/playbook/pb_radio/_radio.jsx
```diff
@@ -11,7 +11,7 @@ type RadioProps = {
   label: String,
   name: String,
   value: String,
-  checked?: Boolean,
+  defaultChecked?: Boolean,
```
thestephenmarshall: Is this solely for the case where the docs are being shown? If so, I'm not really for this change. We should implement a thin state wrapper around the doc to do this instead of modifying the kit itself unless I am missing some other requirement. https://reactjs.org/docs/uncontrolled-components.html#default-values

kre8sions: @thestephenmarshall when I implemented the Block kit I noticed that there was an inversion originally made on the checked prop. So any block kit would be checked if defaultChecked was true. So my change was to use change the prop checked to defaultChecked. If anything I can revert this and implement a different solution.

jasoncypret: Can we not recode this to keep the checked prop...but fix the broken logic? Does the rails version use default_checked as a prop or checked?

thestephenmarshall: > So any block kit wouldn't be checked if defaultChecked was true

Do you mean within the context of `Block`'s documentation or the actual kit?

PR #547, opened 2020-01-07
Typeahead kit

terryfinn:
#### Runway Ticket URL

[BOP-325](https://nitro.powerhrg.com/runway/backlog_items/BOP-325)

Screen Shot 2020-01-09 at 11 39 31 AM

typeahead-github-demo

#### Breaking Changes

None

#### Checklist:

### DIFF
--- a/app/pb_kits/playbook/pb_typeahead/_typeahead.scss
+++ b/app/pb_kits/playbook/pb_typeahead/_typeahead.scss
@@ -1,0 +1,52 @@
+[class^=pb_typeahead_kit] {
+  .pb_typeahead_wrapper {
+    position: relative;
+  }
+
+  .text_input_wrapper {
+    margin-bottom: 0;
+  }
+
+  .pb_item_kit {
+    padding: ($space_xs + 2) 0;
+
+    &:hover {
+      background-color: $bg_light;
+    }
+  }
+
+  &:focus-within [class^=pb_list_kit] {
+    max-height: 18em;
+    overflow-y: auto;
+    overscroll-behavior: contain;
+    position: absolute;
+    left: 0;
+    right: 0;
+    background: white;
+    box-shadow: $shadow_deep;
+    z-index: 100;
+  }
+
+  &:not(:focus-within) [class^=pb_list_kit] {
+    display: none;
+  }
+
+  [class^=pb_list_kit] > [data-pb-typeahead-kit-results] > li {
+    &:focus-within {
+      background: $bg_light;
+    }
+
+    > button {
+      background: none;
+      color: inherit;
+      border: none;
+      padding: 0;
+      font: inherit;
+      cursor: pointer;
+      outline: inherit;
+      width: 100%;
+      height: 100%;
+      text-align: left;
+    }
+  }
+}
--- a/app/pb_kits/playbook/pb_typeahead/docs/_typeahead_default.html.erb
+++ b/app/pb_kits/playbook/pb_typeahead/docs/_typeahead_default.html.erb
@@ -1,0 +1,51 @@
+<%= pb_rails("typeahead", props: { data: { typeahead_example: true } }) %>
+
+<br><br><br>
+
+<%= pb_rails("card", props: { padding: "xl", data: { typeahead_example_selected_option: true } }) do %>
+  <%= pb_rails("body") do %>
+    Use the above input to search for users on Github, and see the results as you type.
+  <% end %>
+
+  <%= pb_rails("body") do %>
+    When you make a selection, you will see it apear in the list below
+  <% end %>
+
+  <div data-selected-option></div>
+<% end %>
+
+<template data-typeahead-example-result-option>
+  <%= pb_rails("user", props: {
+    name: tag(:slot, name: "name"),
+    orientation: "horizontal",
+    align: "left",
+    avatar_url: "placeholder"
+    }) %>
+</template>
+
+<%= javascript_tag defer: "defer" do %>
+  document.addEventListener("pb-typeahead-kit-search", function(event) {
+    if (!event.target.dataset.typeaheadExample) return;
+
+    fetch(`https://api.github.com/search/users?q=${encodeURIComponent(event.detail.searchingFor)}`)
+      .then(response => response.json())
+      .then((result) => {
+        const resultOptionTemplate = document.querySelector("[data-typeahead-example-result-option]")
+
+        event.detail.setResults((result.items || []).map((user) => {
+          const wrapper = resultOptionTemplate.content.cloneNode(true)
+          wrapper.querySelector('slot[name="name"]').replaceWith(user.login)
+          wrapper.querySelector('img').setAttribute("src", user.avatar_url)
+          return wrapper
+        }))
+      })
+  })
+
+
+  document.addEventListener("pb-typeahead-kit-result-option-selected", function(event) {
+    if (!event.target.dataset.typeaheadExample) return;
+
+    document.querySelector("[data-typeahead-example-selected-option] [data-selected-option]").innerHTML = ""
+    document.querySelector("[data-typeahead-example-selected-option] [data-selected-option]").appendChild(event.detail.selected)
+  })
+<% end %>
--- a/app/pb_kits/playbook/pb_typeahead/typeahead.rb
+++ b/app/pb_kits/playbook/pb_typeahead/typeahead.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+module Playbook
+  module PbTypeahead
+    class Typeahead
+      include Playbook::Props
+
+      prop :label
+      prop :placeholder
+      prop :search_term_minimum_length, default: 3
+      prop :search_debounce_timeout, default: 250
+
+      partial "pb_typeahead/typeahead"
+
+      def classname
+        generate_classname("pb_typeahead_kit")
+      end
+
+      def data
+        Hash(values[:data]).merge(
+          pb_typeahead_kit: true,
+          pb_typeahead_kit_search_term_minimum_length: search_term_minimum_length,
+          pb_typeahead_kit_search_debounce_timeout: search_debounce_timeout,
+        )
+      end
+    end
+  end
+end
--- a/spec/pb_kits/playbook/pb_typeahead/typeahead_spec.rb
+++ b/spec/pb_kits/playbook/pb_typeahead/typeahead_spec.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+require_relative "../../../../app/pb_kits/playbook/pb_typeahead/typeahead"
+
+RSpec.describe Playbook::PbTypeahead::Typeahead do
+  subject { Playbook::PbTypeahead::Typeahead }
+
+  it { is_expected.to define_partial }
+
+  it { is_expected.to define_prop(:label) }
+  it { is_expected.to define_prop(:placeholder) }
+  it { is_expected.to define_prop(:search_term_minimum_length).with_default(3) }
+  it { is_expected.to define_prop(:search_debounce_timeout).with_default(250) }
+end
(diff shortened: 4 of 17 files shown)


terryfinn:
> - `PbEnhancedElement` this looks a lot like what you had created for the live form example. I think there might be some reusability there. WDYT?

I think so. This version is more complete than what I tossed together for the example for live form. I think reusing it for that would be very possible.

> - The naming of `PbEnhancedElement` seems broader than the implementation itself which is concerned primarily with observing and reacting to mutations via `mutationObserver` API. Could we perhaps rename this to be more specific?

This was the best name I could come up with at the time. I am open to suggestions. I felt that the element keeping any hints of mutation observer out of it would be a good idea, that seemed far too specific of the implementation.

> - How flexible can we make `app/pb_kits/playbook/pb_typeahead/_typeahead.html.erb` with regards to modification of the `list/item` props and block? Seems like we could improve things there (maybe even in a future iteration) so that different result style combinations could be added by the consumer with ease.

I was keeping it focused, but yeah, we could expose some prop forwarding on the typeadhead to customize the list. I am not sure how much I will be able to do on this first pass though.

> - Could we clean up the implementation so that, instead of leading with `document.addEventListener...`, we could simply spin up an instance? e.g. `new PBTypeAheadKitSearch([options])`

Not sure I follow. Can you expand on this?

> - Things like the result template, debounce, etc I would love to see as optional params so that they may be configurable

I already have a minimum length for triggering the search as a configurable option, I can expose it as a prop directly, instead of being simple a data attribute. I can do the same for debounce timeout.

 I assume that your talking about the content of what shows up in the list. If so, I don't have a any templateing for it. Right now, results are 100% the responsibility of something else. The typeadhed is only responsible to allowing a sending events when a user types, show results that other code has fetched and prepaired, and allowing a user to select a result.

> - Similar to how `pbChart js` was made available to Nitro, this too will have to be implemented in the same way. Currently, that is done via `app/pb_kits/playbook/vendor.js`, but this may be a good time to enhance that route as well.

You mean assign the object to a property on the window. Seems like it would be useful. I can do that. What are your thoughts on "... may be a good time to enhance that route as well."?

terryfinn:
What we want to target for feature and functionality right now is:

- [X] User inputs trigger a event that allows other code to provide the typeadhead with options.
- [X] User can cycle through options via tab, shift + tab, arrow up, arrow down, and scroll.
- [ ] User can select a option via enter or click.
- [ ] User selection triggers a event that allows other code to handle the selection.
- [X] Configurable minimal characters to trigger search.
- [ ] Configurable debounce timeout.

Does that sound sufficiently complete for this first pass?

terryfinn:
@jasoncypret @thestephenmarshall 

For completeness of I want to share the alternative design I considered.

The typeahead could look like this:

```js
import { debounce } from 'lodash'

export default class PbTypeahead extends HTMLElement {
  connectedCallback() {
    this.addEventListener('keydown', (event) => this.handleKeydown(event))
    this.searchInput.addEventListener('focus', () => this.debouncedSearch())
    this.searchInput.addEventListener('input', () => this.debouncedSearch())
  }

  handleKeydown(event) {
    if (event.key === 'ArrowUp') {
      event.preventDefault()
      this.focusPreviousOption()
    } else if (event.key === 'ArrowDown') {
      event.preventDefault()
      this.focusNextOption()
    }
  }

  search() {
    if (this.searchTerm.length < this.searchTermMinimumLength) return this.clearResults()

    this.dispatchEvent(new CustomEvent('pb-typeahead-kit-search', { bubbles: true, detail: this }))
  }

  get debouncedSearch() {
    return this._debouncedSearch = (this._debouncedSearch || debounce(this.search, 250).bind(this))
  }

  showResults(results) {
    this.clearResults()

    for (const result of results) {
      this.resultsElement.appendChild(this.newResultOption(result))
    }
  }

  clearResults() {
    this.resultsElement.innerHTML = ''
  }

  newResultOption(content) {
    const resultOption = this.resultOptionTemplate.content.cloneNode(true)
    resultOption.querySelector('slot[name="content"]').replaceWith(content)
    return resultOption
  }

  focusPreviousOption() {
    const currentIndex = this.resultOptionItems.indexOf(this.currentSelectedResultOptionItem)
    const previousIndex = currentIndex - 1
    const previousOptionItem = this.resultOptionItems[previousIndex] ||
                               this.resultOptionItems[this.resultOptionItems.length - 1]
    previousOptionItem.focus()
  }

  focusNextOption() {
    const currentIndex = this.resultOptionItems.indexOf(this.currentSelectedResultOptionItem)
    const nextIndex = currentIndex + 1
    const nextOptionItem = this.resultOptionItems[nextIndex] ||
                           this.resultOptionItems[0]
    nextOptionItem.focus()
  }

  get resultOptionItems() {
    return Array.from(this.resultsElement.querySelectorAll('[data-result-option-item="true"]'))
  }

  get currentSelectedResultOptionItem() {
    return document.activeElement.closest('[data-result-option-item="true"]')
  }

  get searchInput() {
    return this._searchInput = (this._searchInput || this.querySelector('input[type="search"]'))
  }

  get searchTerm() {
    return this.searchInput.value
  }

  get searchTermMinimumLength() {
    return this.dataset.pbTypeaheadKitSearchTermMinimumLength
  }

  get resultsElement() {
    return this._resultsElement = (this._resultsElement || this.querySelector('[data-pb-typeahead-kit-results]'))
  }

  get resultOptionTemplate() {
    return this._resultOptionTemplate = (
      this._resultOptionTemplate ||
      this.querySelector('template[data-pb-typeahead-kit-result-option]')
    )
  }
}

customElements.define("pb-typeahead", PbTypeahead);
```

The advantage of this approach is it removes the need for the mutation observer and enhanced element classes. The downside is we are defining everything directly on a element object, and that is a little more restrictive, and couples the element with the enhancements.

thestephenmarshall:
@terryfinn I liked the idea of using the classes from the previous approach potentially for helping to build the HTML form validation that I am currently working on. I do, however, see the advantage of the more concise approach in your latest proposal. 

Also, here are some responses to the above:

> Not sure I follow. Can you expand on this?

I was just suggesting that instead using an event listener to use this, perhaps a slightly different approach. Really not an issue the way it is though. 

> ...What are your thoughts on "... may be a good time to enhance that route as well."?

This was just a reference to some cleanup I may want to focus on if we feel that this particular ~~route~~ area of Playbook might be a little vague in approach. Likely there is nothing to do immediately in this PR.

terryfinn:
> @terryfinn I liked the idea of using the classes from the previous approach potentially for helping to build the HTML form validation that I am currently working on. I do, however, see the advantage of the more concise approach in your latest proposal.

Just to be clear, I am not proposing we use custom elements, I was just sharing an alternative I had considered. I custom elements, and the whole Web Component API, and think they are grate, but I think they also have some slight disadvantages. First, and least important, they have slightly lower support than Mutation Observers. That said, this support gap is small, and will only get smaller as time goes on. The bigger disadvantage I see, is that our customization code and the element code become one in the same object. This means we have less flexibility, as we need the element to still be a valid element object. The advantage of the custom element approach is we end up with little over 100 lines of code from the base library to support the enhanced element going away, and get to rely on more native browser code.

thestephenmarshall:
@terryfinn Got it. I think we should continue to keep Web Components in our peripherals. From my experience, there are definitely benefits from being more declarative.

thestephenmarshall:
@terryfinn would you say that `PbEnhancedElement` is at a point where I could begin to leverage it safely within my form validation work or is there more to come?

terryfinn:
@thestephenmarshall I think `PbEnhancedElement` is fairly complete at this point, so yes, I think you can start to leverage it in form validations.

app/pb_kits/playbook/pb_typeahead/_typeahead.scss
```diff
@@ -0,0 +1,52 @@
+.pb_typeahead {
```
thestephenmarshall:
```suggestion
[class^=pb_typeahead_kit] {
```

app/pb_kits/playbook/pb_typeahead/_typeahead.scss
```diff
@@ -0,0 +1,52 @@
...
+  .pb_typeahead_wrapper {
+    position: relative;
+  }
+
+  .text_input_wrapper {
+    margin-bottom: 0;
+  }
+
+  .pb_item_kit {
+    padding: 10px 0;
```
thestephenmarshall: We should probably use spacing tokens from `app/pb_kits/playbook/tokens/_spacing.scss` here instead of static px values.

app/pb_kits/playbook/pb_typeahead/_typeahead.scss
```diff
@@ -0,0 +1,52 @@
...
+
+  &:focus-within [class^=pb_list_kit] {
+    max-height: 18em;
+    overflow-y: auto;
+    overscroll-behavior: contain;
+    position: absolute;
+    left: 0;
+    right: 0;
+    background: white;
+    box-shadow: 0px 8px 16px 0px rgba(0,0,0,0.2);
```
thestephenmarshall: This should be leveraging shadow values that have already been UX approved `app/pb_kits/playbook/tokens/_shadows.scss`

app/pb_kits/playbook/pb_typeahead/typeahead.rb
```diff
@@ -0,0 +1,28 @@
...
+
+      prop :label
+      prop :placeholder
+      prop :search_term_minimum_length, default: 3
+      prop :search_debounce_timeout, default: 250
+
+      partial "pb_typeahead/typeahead"
+
+      def classname
+        generate_classname("pb_typeahead")
```
thestephenmarshall:
```suggestion
        generate_classname("pb_typeahead_kit")
```

app/pb_kits/playbook/pb_typeahead/docs/_typeahead_default.html.erb
```diff
@@ -0,0 +1,51 @@
...
+    name: tag(:slot, name: "name"),
+    orientation: "horizontal",
+    align: "left",
+    avatar_url: "placeholder"
+    }) %>
+</template>
+
+<%= javascript_tag defer: "defer" do %>
+  document.addEventListener("pb-typeahead-kit-search", function(event) {
+  if (!event.target.dataset.typeaheadExample) return;
```
thestephenmarshall: Some missing indentation here.

app/pb_kits/playbook/pb_typeahead/_typeahead.scss
jasoncypret: I feel like this could also be a results kit...but we could abstract in a later release

app/pb_kits/playbook/pb_typeahead/_typeahead.scss
terryfinn: @thestephenmarshall What value would you suggest?

spec/pb_kits/playbook/pb_typeahead/typeahead_spec.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+require_relative "../../../../app/pb_kits/playbook/pb_typeahead/typeahead"
+
+RSpec.describe Playbook::PbTypeahead::Typeahead do
+  subject { Playbook::PbTypeahead::Typeahead }
+
+  it { is_expected.to define_partial }
```
thestephenmarshall: Be sure to add some prop coverage here

app/pb_kits/playbook/pb_typeahead/_typeahead.scss
terryfinn: @jasoncypret I agree, I am sure we could extract out more than just results as a new kit. I like the idea of punting that for now.

app/pb_kits/playbook/pb_typeahead/_typeahead.scss
thestephenmarshall: Try `$space_xs + 2 0`

app/pb_kits/playbook/pb_typeahead/_typeahead.scss
terryfinn: @thestephenmarshall Any suggestions for the shadows?

thestephenmarshall: @terryfinn try `box-shadow: $shadow_deep;` first

PR #583, opened 2020-01-31
DevDoc Fix

kre8sions:
Updated generator, doc helper, doc example to expose MENU.yml in Nitro Web


kre8sions:
@thestephenmarshall Yup

PR #592, opened 2020-02-05
Ignore timezone on DateStacked

gmfvpereira:
#### Runway Ticket URL
Without these changes a usage like this
```
      <DateStacked
          date="2018-03-20"
          size="md"
      />
```
would create different results depending on the timezone, it could say that the date is "March 19th, 2018", for instance. 

Since this is a Date only kit my assumption is that timezones are not desired.

#### Breaking Changes

This should not break anything. This kit still accepts `new Date` or formatted date strings like `"2020-08-30"` as a valid date.

The `/compliance` page in Nitro uses this Kit, so we can verify that it's still working as expected:
- If the date is in the same year as the current year, it should only show day and month
- If the date is NOT in the same year as the current year, it should show the year as well.

#### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+- Fixed timezone issue on DateStacked ([#592][] @gmfvpereira)
 
 ## [4.2.0] 2020-2-6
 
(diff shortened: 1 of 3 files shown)


gmfvpereira:
I'll need some guidance on how to add an entry in the CHANGELOG. 
Which one should I choose: `CHANGED/REMOVED/FIXED`?

Also, in the previous version, the dependency on `moment` was encapsulated on `dateTime.js`. I thought about doing the same with a `date.js` file, but I was not sure if it's desired or not. Thoughts? If it's desired, I'm accepting suggestions for the class name, since `Date` is already in user.

gmfvpereira:
@jasoncypret I rebased on v4.3.0.

gmfvpereira:
@jasoncypret is it good now?

PR #596, opened 2020-02-07
[BOLT-71] Select Kit: Fix misuse of selected vs value props

drborges:
#### Runway Ticket URL

https://nitro.powerhrg.com/runway/backlog_items/BOLT-71

Fixes the following warning:

Screen Shot 2020-02-07 at 2 56 17 PM

#### Breaking Changes

1. Renamed the option `valueText` attribute to `text` for a more concise API.
1. Removed the ability to directly mark an option as `selected` in favor of the `value` prop on the `select` element.

There seems to be only one usage of the `Select` kit in `nitro-web` in the installation crews profile page which will need to be updated to meet the new `Select` API and be regression tested.

#### Checklist:

### DIFF
--- a/app/pb_kits/playbook/pb_select/_select.jsx
+++ b/app/pb_kits/playbook/pb_select/_select.jsx
@@ -20,7 +20,6 @@ type SelectOption = {
   value: string,
   valueText: string,
   disabled?: boolean,
-  selected?: boolean,
 }
 
 type SelectProps = {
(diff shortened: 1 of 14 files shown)


drborges:
@ghaydarov, @miniyakkos I noticed a [reference to this component](https://github.com/powerhome/nitro-web/blob/bolt-59/install-schedule-territory-filter/components/home_tour/app/components/FormModal/Rooms.jsx#L42:L42) in the home tour but it seems that my changes here would not cause trouble there, however, that code also makes use of the invalid `selected` prop on the `option` element.

Please, consider removing the child `select` in favor of passing in the list of `options` to the `Select` component, [like in this example](https://github.com/powerhome/playbook/blob/select/fix-misuse-of-selected-vs-value-props/app/pb_kits/playbook/pb_select/docs/_select_disabled_options.jsx#L39:L39).

PR #606, opened 2020-02-12
Ignore timezone on date stacked

jasoncypret:
#### Runway Ticket URL
Without these changes a usage like this
```
      <DateStacked
          date="2018-03-20"
          size="md"
      />
```
would create different results depending on the timezone, it could say that the date is "March 19th, 2018", for instance. 

Since this is a Date only kit my assumption is that timezones are not desired.

Screen Shot 2020-02-13 at 2 42 08 PM

#### Breaking Changes

This should not break anything. This kit still accepts `new Date` or formatted date strings like `"2020-08-30"` as a valid date.

The `/compliance` page in Nitro uses this Kit, so we can verify that it's still working as expected:
- If the date is in the same year as the current year, it should only show day and month
- If the date is NOT in the same year as the current year, it should show the year as well.

#### Checklist:

### DIFF
--- a/app/pb_kits/playbook/pb_date_stacked/_date_stacked.jsx
+++ b/app/pb_kits/playbook/pb_date_stacked/_date_stacked.jsx
@@ -2,8 +2,8 @@
 
 import React from 'react'
 import classnames from 'classnames'
+import moment from 'moment'
 
-import DateTime from '../pb_kit/dateTime.js'
 import { buildCss } from '../utilities/props'
 import { Caption, Title } from '../'
 
@@ -36,20 +36,20 @@ const DateStacked = ({
     'reverse': reverse,
   }))
 
-  const currentYear = new Date().getFullYear().toString()
-  const dateTimestamp = new DateTime({ value: date })
-  const inputYear = dateTimestamp.toYear().toString()
+  const currentYear = moment().year()
+  const dateObject = moment(date)
+  const inputYear = dateObject.format('Y')
 
   return (
     <div className={classes}>
       <div className="pb_date_stacked_day_month">
         <Caption
-            text={dateTimestamp.toMonth().toUpperCase()}
+            text={dateObject.format('MMM').toUpperCase()}
         />
         <Title
             dark={dark}
             size={sizes[size]}
-            text={dateTimestamp.toDay()}
+            text={dateObject.format('D')}
         />
       </div>
       <If condition={currentYear != inputYear}>
--- a/app/pb_kits/playbook/pb_date_stacked/_date_stacked.jsx
+++ b/app/pb_kits/playbook/pb_date_stacked/_date_stacked.jsx
@@ -2,7 +2,7 @@
 
 import React from 'react'
 import classnames from 'classnames'
-import moment from 'moment'
+import DateTime from '../pb_kit/dateTime.js'
 
 import { buildCss } from '../utilities/props'
 import { Caption, Title } from '../'
@@ -36,20 +36,20 @@ const DateStacked = ({
     'reverse': reverse,
   }))
 
-  const currentYear = moment().year()
-  const dateObject = moment(date)
-  const inputYear = dateObject.format('Y')
+  const currentYear = new Date().getFullYear().toString()
+  const dateTimestamp = new DateTime({ value: date })
+  const inputYear = dateTimestamp.toYear().toString()
 
   return (
     <div className={classes}>
       <div className="pb_date_stacked_day_month">
         <Caption
-            text={dateObject.format('MMM').toUpperCase()}
+            text={dateTimestamp.toMonth().toUpperCase()}
         />
         <Title
             dark={dark}
             size={sizes[size]}
-            text={dateObject.format('D')}
+            text={dateTimestamp.toDay()}
         />
       </div>
       <If condition={currentYear != inputYear}>
--- a/app/pb_kits/playbook/pb_date_stacked/docs/_date_stacked_not_current_year.jsx
+++ b/app/pb_kits/playbook/pb_date_stacked/docs/_date_stacked_not_current_year.jsx
@@ -13,7 +13,7 @@ const DateStackedNotCurrentYear = () => {
       <br />
 
       <DateStacked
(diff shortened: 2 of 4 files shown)


app/pb_kits/playbook/pb_date_stacked/_date_stacked.jsx
```diff
@@ -2,8 +2,8 @@
 
 import React from 'react'
 import classnames from 'classnames'
+import moment from 'moment'
```
thestephenmarshall: We shouldn't be directly importing `moment` here. Originally, the intent was to make use of `dateTime.js` which imports the lib.

app/pb_kits/playbook/pb_date_stacked/_date_stacked.jsx
```diff
@@ -36,20 +36,20 @@ const DateStacked = ({
...
+  const currentYear = moment().year()
+  const dateObject = moment(date)
+  const inputYear = dateObject.format('Y')
 
   return (
     <div className={classes}>
       <div className="pb_date_stacked_day_month">
         <Caption
-            text={dateTimestamp.toMonth().toUpperCase()}
+            text={dateObject.format('MMM').toUpperCase()}
```
thestephenmarshall: `dateTime.js` was created to prevent verbosity here. If need be, we should develop new exports/methods there.

app/pb_kits/playbook/pb_date_stacked/docs/_date_stacked_not_current_year.jsx
```diff
@@ -13,7 +13,7 @@ const DateStackedNotCurrentYear = () => {
       <br />
 
       <DateStacked
-          date={new Date('20 Mar 2018')}
+          date="2018-03-20"
```
thestephenmarshall: We want to be able to send either a `Date` object or a `String` here instead of just `String`.

jasoncypret:
@gmfvpereira Let me know if you have any questions here. We felt like the fix should be in dateTime.js so all kits could benefit from it. Plus we have more reuse that way.

gmfvpereira:
About not using `moment` directly on the class that's precisely my question on the [original PR](https://github.com/powerhome/playbook/pull/592). I'll address it.

About using `dateTime.js` that's probably not desirable. As I said before, dates have no timezones. If someone uses a `DateStacked` kit to show a user's hire date on "Jan 5th", that should be January 5th in any place of the world, it doesn't matter the timezone. 

Since right now we're considering timezones for a date kit, we potentially have a bug where BAMs outside Philadelphia timezone are seeing wrong start dates for new hires. No one complained yet despite myself or any brazilian that's working -2 hours from philly timezone. That's likely to be happening in other timezones as well.

So, my first idea was to create a `date.js` file, but obviously `Date` is a reserved word. So, I end up adding moment itself. I'll need some help to come up with a good name for it. Ideas?

> but I am also curious how this PR made it's way out beyond UX approval.

Curious about what specifically @thestephenmarshall ? 
Here's what happened: I tried to create a conversation, after 6 days I had no answer,  but I got Diego's review on it and I had at least @jasoncypret and @kre8sions eyes on it as well. So I hit merge. 

Is there a process that I should follow that I'm not aware of? When you say "UX Approval", who are these people exactly?

app/pb_kits/playbook/pb_date_stacked/docs/_date_stacked_not_current_year.jsx
gmfvpereira: We do accept both. There was no examples with String, I just added one.

If you scroll up you can see a Date example
```
      <DateStacked
          date={new Date('20 Mar 2018')}
          size="sm"
      />

      <br />

      <DateStacked
          date="2018-03-20"
          size="md"
      />
```

gmfvpereira:
@thestephenmarshall as you had suggested before, this is making the `zone` defaulting to `null` instead of a specific timezone. Let me know if this is a better approach.

thestephenmarshall:
@gmfvpereira 

> So, my first idea was to create a date.js file, but obviously Date is a reserved word. So, I end up adding moment itself. I'll need some help to come up with a good name for it. Ideas?

How about `dates.js`? Within that, `moment` could be added (although it would be great to phase out moment and replace with https://date-fns.org instead) and all is well. 

> Curious about what specifically @thestephenmarshall ?
Here's what happened: I tried to create a conversation, after 6 days I had no answer, but I got Diego's review on it and I had at least @jasoncypret and @kre8sions eyes on it as well. So I hit merge.

Yep. Makes sense. So UX was specifically added to the CODEOWNERS and it appears as though GH isn't honoring this anymore. We likely can change that no problem, but that is ultimately a GH feature blunder. Likely, we will have to make settings changes as yet another part of our release process 🙃.

> Is there a process that I should follow that I'm not aware of? When you say "UX Approval", who are these people exactly?

@powerhome/ux-team - this is the group listed in CODEOWNERS.

gmfvpereira:
@thestephenmarshall thanks for clarifying everything. 

What's the next step, can I merge it already or there is another step required?

thestephenmarshall:
@gmfvpereira let's get an approval from @jasoncypret

app/pb_kits/playbook/pb_date_stacked/docs/_date_stacked_not_current_year.jsx
jasoncypret: We still need to show this works off a date object. Can you either add it back or add another example using a string?

gmfvpereira: That's here already: https://github.com/powerhome/playbook/pull/606/files#diff-b8e86b8e3fe5e0700f457960999baf69R9

(line 9 on the same file)

jasoncypret: Ahh sorry, I missed that. Perfect then!

gmfvpereira: Thanks guys!

PR #621, opened 2020-02-18
Add a context option for the typeahead

terryfinn:
Allow the typeahead to use a context to make searches distinct when
using the same typeahead to provide results with different
conditions.

The typeahead has a cache, but that might not be the best name for
it. This structure was primarily added to provide coordination for
asynchronous searches. We can get the benefits of a cache,
and always provide the user with fresh results. We change the search
to always show the cached values when results are present but also
always issuing a event to get new results.

#### Runway Ticket URL
[BOP-357](https://nitro.powerhrg.com/runway/backlog_items/BOP-357)

#### Checklist:


thestephenmarshall:
Can we update the docs with an example pls?

PR #626, opened 2020-02-20
Use mutation observer to bind behavior of PbPopover

indiebrain:
#### Runway Ticket URL

Make PbPopover behavior work with turbolinks navigation events. 

https://nitro.powerhrg.com/runway/backlog_items/GOG-83

#### Breaking Changes

Should be a transparent-to-the-client change to the way behavior is bound to the PbPopover so long as the client is importing the playbook webpack. 

Additional concerns would be the - hopefully unlikely case - that the client explicitly messes with the `window.PbPopover` which has been removed from the `vendor.js`.

Click on uses of the PbPopover kit. The behavior should not change.

Screen Shot 2020-02-20 at 7 53 56 AM

#### Checklist:

- [x] **SPECS** Please cover your changes with specs (existing coverage)


app/pb_kits/playbook/vendor.js
```diff
@@ -7,7 +7,3 @@ import './pb_form/pb_form_validation'
 
 // Lazy image loading
 import 'lazysizes'
-
-// Popovers
-import PbPopover from './pb_popover/_popover.js'
-window.PbPopover = PbPopover
```
thestephenmarshall: This was removed and now the kit is broken in Nitro.

PR #633, opened 2020-02-24
DNM: Fix SVG display

thestephenmarshall:
# DO NOT MERGE


thestephenmarshall:
Overall, I do not think that this kit should provide the actual SVG images as props. Reason being (and @jasoncypret correct me if I am off-track) is that any outside users of this library will not want those badges. Forcing those badges would only make this kit useful to Power. 

The proposed solution to the above is to make this kit a "block" kit which means we are passing in children elements (the SVG image) as a block to be yielded in the HTML. The implementor of this kit will hold the responsibility of ensuring the SVG is correct for their application. One thing to note here, is that we will need to be sure to add some notes to the example that talks about ensuring that the SVGs to be badged will contain necessary attributes: `viewBox` and `preserveAspectRatio` as seen in this PR. This will allow our kit to be useful with regards to sizing `width` and `height` properties from SCSS. 

@jasoncypret I'm also looking for some feedback from you here as well. This isn't set in stone. Here is one idea I have for this kit *if we do want to include some default badges of our own*:

- Provide the badges with props such as `badge: veteran` where the `veteran` string prop would automatically load the SVG. (this will only be useful for Power and we still need to determine the best way to source in the SVG images via Webpacker for JSX cases)
- Alternatively, use the block to yield custom SVG images for those who will not find any use from our badges.

Looking for some thoughts here...

app/pb_kits/playbook/pb_user_badge/_user_badge.html.erb
```diff
@@ -2,7 +2,7 @@
     id: object.id,
     data: object.data,
     class: object.classname) do %>
-  <%= content_tag(:div, class: "pb__user_badge_wrapper") do %>
-    <%= render partial: object.display_badge %>
+  <%= content_tag(:div, class: "pb_user_badge_wrapper") do %>
+    <%= capture(&object.children) %>
```
thestephenmarshall: This is allowing block content to yield here.

app/pb_kits/playbook/pb_user_badge/_user_badge.jsx
```diff
@@ -4,18 +4,18 @@ import React from 'react'
 type UserBadgeProps = {
   className?: String,
   id?: String,
-  badge?: 'million-dollar' | 'veteran' ,
+  badge: React.ReactNode,
```
thestephenmarshall: We might even try `'million-dollar' | 'veteran' | React.ReactNode` which will better follow the proposal in the comments..

app/pb_kits/playbook/pb_user_badge/_user_badge.jsx
```diff
@@ -4,18 +4,18 @@ import React from 'react'
...
 }
 
 const UserBadge = ({
-  badge = 'million-dollar',
+  badge,
   size = 'md',
-}) => (
+}: UserBadgeProps) => (
   <div className={`pb_user_badge_kit_${size}`}>
-    <If condition={UserBadgeProps.badge == "veteran"}>
```
thestephenmarshall: You error was coming from `UserBadgeProps` attempting to be directly accessed. It is simply `badge` instead.

app/pb_kits/playbook/pb_user_badge/_user_badge.scss
```diff
@@ -17,7 +17,7 @@ $badge-sizes: (
 
   @each $name, $size in $badge-sizes {
     &[class*=_#{$name}] {
-      .pb__user_badge_wrapper {
```
thestephenmarshall: Let's lose the extra underscore here.

app/pb_kits/playbook/pb_user_badge/badges/_veteran.html.erb
```diff
@@ -0,0 +1,8 @@
+<svg viewBox="0 0 200 250" preserveAspectRatio="xMidYMid meet">
```
thestephenmarshall: `viewBox` here is very important along with `preserveAspectRatio` and [here is why](https://css-tricks.com/scale-svg/#article-header-id-1)

app/pb_kits/playbook/pb_user_badge/docs/_user_badge_default.jsx
```diff
@@ -1,12 +1,32 @@
...
+    <path
+        d="M99.3,125.4V33.5L1.6,102.7l-0.1,0.1c-1,0.7-1.5,1.8-1.5,3v0v67.4v0c0,1.2,0.6,2.3,1.5,3l0.1,0.1l93.5,66.3c1.7,1.2,4.1,0,4.1-2.1v-28.9l40.5-28l-40.5,0.7v-25h73.7l25.4-16.9l0.2-16.8H99.3z"
+    />
+    <rect
+        height="24.3"
+        width="99.3"
+        x="99.3"
+        y="101.1"
+    />
+  </svg>
```
thestephenmarshall: This would be better if we can determine the best way to import the same SVG that we render as a partial on the Rails side.

app/pb_kits/playbook/pb_user_badge/user_badge.rb
```diff
@@ -7,9 +7,9 @@ class UserBadge
 
       partial "pb_user_badge/user_badge"
 
-      prop :badge, type: Playbook::Props::Enum,
-                   values: %w[million-dollar veteran],
-                   default: "million-dollar"
+      # prop :badge, type: Playbook::Props::Enum,
+      #              values: %w[million-dollar veteran],
+      #              default: "million-dollar"
```
thestephenmarshall: Kept this here bc we could still end up using this except without a default and more as a string instead potentially.

app/pb_kits/playbook/pb_user_badge/user_badge.rb
```diff
@@ -18,9 +18,9 @@ def classname
         generate_classname("pb_user_badge_kit", size)
       end
 
-      def display_badge
-        "pb_user_badge/badges/#{badge}"
-      end
+      # def display_badge
+      #   "pb_user_badge/badges/#{badge}"
+      # end
```
thestephenmarshall: This could be added back, but see the above for caveats.

jasoncypret:
I'd say let's have a chat about this. I'm reluctant to make this a block kit

app/pb_kits/playbook/pb_user_badge/_user_badge.jsx
eabousaif: Oh okay, I wasn't aware that badge would be accessible like that. Good to know

PR #661, opened 2020-03-12
Multi select kit

jasoncypret:
#### Breaking Changes

#### Checklist:


app/pb_kits/playbook/kits/pb_form_pill.js
```diff
@@ -0,0 +1,4 @@
+import FormPill from '../pb_form_pill/_form_pill.jsx'
+
+import WebpackerReact from 'webpacker-react'
+WebpackerReact.setup({ FormPill })
```
thestephenmarshall: this should not be here

app/pb_kits/playbook/pb_form_pill/_form_pill.jsx
```diff
@@ -0,0 +1,74 @@
...
+name: String,
+onClick?: EventHandler,
+avatar?: Boolean,
+avatarUrl?: String,
+}
+const FormPill = ({
+  className,
+  text,
+  name,
+  avatar = false,
```
thestephenmarshall: This might cause unintended bugs - perhaps we can not add the prop if it is blank

PR #680, opened 2020-03-23
Time Range Inline kit

evertoncunha:
Screenshot_2020-03-27 Playbook Design System(1)Screenshot_2020-03-27 Playbook Design System(4)

Screenshot_2020-03-27 Playbook Design System(2)Screenshot_2020-03-27 Playbook Design System(3)

#### Runway Ticket URL

https://nitro.powerhrg.com/runway/backlog_items/NUX-637

#### Breaking Changes

#### Checklist:


app/pb_kits/playbook/pb_time_range_inline/_time_range_inline.scss
```diff
@@ -0,0 +1,27 @@
...
+  }
+  .align-center {
+    display: block;
+    margin: auto;
+    width: max-content;
+  }
+  .align-left {
+    display: block;
+    width: max-content;
+  }
```
thestephenmarshall: For these, you might be better off using the `Flex` kit to achieve the same/similar results.

app/pb_kits/playbook/pb_time_range_inline/_time_range_inline.html.erb
```diff
@@ -0,0 +1,23 @@
...
+        <% text_kit = "caption" %>
+        <% text_timezone_color = nil %>
+        <% if object.size == "sm" %>
+          <% text_kit = "body" %>
+          <% text_timezone_color = "light" %>
+        <% end %>
+        <% html_icon = "" %>
+        <% if object.icon == true %>
+          <% html_icon = pb_rails("icon", props: { icon: "clock", dark: object.dark, classname:"space-right", fixed_width: true, size: object.size }) %>
+        <% end %>
```
thestephenmarshall: We should move any variable assignments to methods with the same name. You won't have to pass any arguments as they will already be available via `object` in `app/pb_kits/playbook/pb_time_range_inline/time_range_inline.rb`

spec/pb_kits/playbook/kits/time_range_inline_spec.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require_relative "../../../../app/pb_kits/playbook/pb_time_range_inline/time_range_inline"
+
+RSpec.describe Playbook::PbTimeRangeInline::TimeRangeInline do
+  subject { Playbook::PbTimeRangeInline::TimeRangeInline }
+
+  it { is_expected.to define_partial }
+
+  # Do not leave this file blank. Use other spec files for example tests.
```
thestephenmarshall: You will want to add your test coverage here

app/pb_kits/playbook/pb_time_range_inline/_time_range_inline.scss
evertoncunha: Using `Flex` now, thank you.

spec/pb_kits/playbook/kits/time_range_inline_spec.rb
evertoncunha: Included test coverage, thanks.

app/pb_kits/playbook/pb_time_range_inline/_time_range_inline.html.erb
evertoncunha: @thestephenmarshall I don't know if the current version of the code satisfies this, but I couldn't find a way to remove the `if` logic from the `erb` file. Could you help me to achieve that? Thanks.

app/pb_kits/playbook/pb_time_range_inline/_time_range_inline.jsx
```diff
@@ -0,0 +1,134 @@
...
+            tag="span"
+        />
+      </If>
+    )
+  }
+
+  return (
+    <div className={classnames('pb_time_range_inline_kit_' + alignment, className)}>
+      <div className="pb_time_range_inline_wrapper">
+        <If condition={size == 'xs'}>
```
gregblake: I wasn't familiar with the `<If>` component prior to this PR, but it seems like it could be a handy tool for some use cases. Is this component part of React proper, or part of another library?

thestephenmarshall: @gregblake this is [`jsx-control-statements`](https://www.npmjs.com/package/jsx-control-statements) which has been in PB for a while now and is actually part of `nitro-react` as well!

PR #686, opened 2020-03-26
Mobile scaling + table responsive settings + avatar issues

jasoncypret:
#### Runway Ticket URL
https://nitro.powerhrg.com/runway/backlog_items/NUX-763


jasoncypret:
@thestephenmarshall I don't think 4.9.0 was rebased off 4.8.2 as it's showing changes that were made there. I believe if we rebase 4.9.0 off 4.8.2 those extra files will go away here.

PR #689, opened 2020-04-02
WIP: File upload kit

thestephenmarshall:
#### Runway Ticket URL

# WIP

#### Breaking Changes

#### Checklist:


app/controllers/playbook/pages_controller.rb
```diff
@@ -64,5 +65,12 @@ def set_kit
         redirect_to root_path, flash: { error: "That kit does not exist" }
       end
     end
+
+    def ensure_rails_kit_exists
+      # TODO: unsure why we cannot simply use the helpers that are included in ApplicationController - fix this
```
thestephenmarshall, line 70: @jasoncypret any idea why we can't use `Playbook::PbDocHelper` instead of re-creating this? I tried all kinds of things here and could not succeed.

app/controllers/playbook/pages_controller.rb
```diff
@@ -64,5 +65,12 @@ def set_kit
...
+      erbfiles = File.join("**", "*.erb")
+      kit_files = Dir.glob(erbfiles, base: "#{Playbook::Engine.root}/app/pb_kits/playbook/pb_#{@kit}/docs").present?
+      redirect_to action: "kit_show_react" unless kit_files.present?
```
thestephenmarshall, line 73: Since we default to `rails` kit version from the nav, we should redirect if that doesn't exist (small edge case, but relevant)

thestephenmarshall:
Closing in lieu of https://github.com/powerhome/playbook/pull/730

PR #692, opened 2020-04-07
Move onClick prop out of Icon and in to FormPill

thestephenmarshall:
@gmfvpereira I was running your branch for a PR that I need to create regarding cleanup of `onClick` handlers across the board. Upon testing, I noticed that the `<i onClick=...` does not actually fire at all (which is sort of surprising - sorta not). Since it isn't doing anything, I moved the prop to the wrapping `<div>` within `FormPill`. This time, the `onClick` is fired as seen here in the doc examples for React. Let me know if you have any questions or concerns. I would like to know if you had this working for you locally and perhaps there is some strange delta I am seeing here.. Thanks!

BTW - I did create the [revert PR](https://github.com/powerhome/playbook/pull/693)

### DIFF
--- a/app/pb_kits/playbook/pb_form_pill/_form_pill.jsx
+++ b/app/pb_kits/playbook/pb_form_pill/_form_pill.jsx
@@ -47,11 +47,13 @@ const FormPill = ({
             text={text}
         />
       </If>
-      <div className="pb_form_pill_close">
+      <div
+          className="pb_form_pill_close"
+          onClick={onClick}
+      >
         <Icon
             fixedWidth
             icon="times"
-            onClick={onClick}
         />
       </div>
     </div>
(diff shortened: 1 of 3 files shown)


gmfvpereira:
My bad. 
I couldn't make my local server works in order to create these docs/tests. 
So I simply changed straight in the `node_modules/playbook` files and when I saw that the mouse pointer was changing to indicate a link I thought it was done. But you're right, the onClick is not triggering. Any ideas why?

thestephenmarshall:
@gmfvpereira I tried finding the answer, but all I can assume is that `<i>` (similar to `<span>`?) isn't considered a "clickable" DOM element in HTML - so JSX doesn't get `onClick` for those..

PR #695, opened 2020-04-07
[WIP] Add card header prop using react sub-components

christinaatai:
#### Breaking Changes

#### Checklist:

### DIFF
--- a/app/pb_kits/playbook/pb_card/_card_test.jsx
+++ b/app/pb_kits/playbook/pb_card/_card_test.jsx
@@ -1,105 +1,103 @@
-/* @flow */
+// /* @flow */
 
-import React from 'react'
-import classnames from 'classnames'
-import { buildCss } from '../utilities/props'
+// import React from 'react'
+// import classnames from 'classnames'
+// import { buildCss } from '../utilities/props'
 
-//  Card props
-/*
-type CardPropTypes = {
-  children: Array<React.ReactNode> | React.ReactNode,
-  className?: String,
-  header?: {
-    padding?: 'none' | 'xs' | 'sm' | 'md' | 'lg' | 'xl',
-    color?: String
-  }
-  bodyPadding?: 'none' | 'xs' | 'sm' | 'md' | 'lg' | 'xl',
-}
+// //  Card props
+// /*
+// type CardPropTypes = {
+//   children: Array<React.ReactNode> | React.ReactNode,
+//   className?: String,
+//   header?: {
+//     padding?: 'none' | 'xs' | 'sm' | 'md' | 'lg' | 'xl',
+//     color?: String
+//   }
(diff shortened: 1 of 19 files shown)


app/pb_kits/playbook/pb_card/_card_test.jsx
```diff
@@ -0,0 +1,105 @@
...
+  bodyPadding?: 'none' | 'xs' | 'sm' | 'md' | 'lg' | 'xl',
+}
+
+// Establishing css classes
+const CardTest = ({
+  children,
+  className,
+  header = {},
+  bodyPadding = 'md',
+}: CardPropTypes) => {
```
gillianwenhold, line 25: I think this stuff is important to have in there!

app/pb_kits/playbook/pb_card/_card_test.jsx
```diff
@@ -0,0 +1,105 @@
...
+    const childType =
+    child && child.type && (child.type.displayName || child.type.name);
+    if (type.includes(childType)) {
+    result.push(child);
+    }
+});
+
+/* Then we go through each React children, if one of matches the name of the sub-component we’re looking for we put it in the result array */
+return result[0];
+};
```
gillianwenhold, line 51: This looks great! If I'm understanding correctly, this is being used as a boolean that will resolve to `true` if a header should be displayed, and `false` if it should not be displayed. If so, the `.find` method could also work! It would return `undefined` if the `childType` was not there, then return the first object it finds if any match the criteria. Something like this: 

`children.find(({ type }) => (type.displayName || type.name) === childType)`

gillianwenhold: Also - I think it may be possible that this function exists already without having to declare it...unclear though

app/pb_kits/playbook/pb_card/_card_test.jsx
```diff
@@ -0,0 +1,105 @@
...
+		return null;
+	}
+
+	render () {
+		const { children, className } = this.props;
+		return (
+			<CardTest>
+			</CardTest>
+		);
+	}
```
gillianwenhold, line 98: I think since this render is inside of the `CardTest` component, the return value would be calling the `renderHeader` and `renderBody` components inside of a div (maybe with a kit name)? Ex: 

```
render (
  <div>
    this.renderHeader()
    this.renderBody()
  </div>
)
```

gmfvpereira:
@christinaatai I was not able to test locally because my `make start` [always fail](https://gist.github.com/gmfvpereira/3a2794ee36ce2fab999f99491bae221f).

But I *think* [this is the way you wanna do it](https://github.com/powerhome/playbook/pull/695/commits/c30d62ad064dbc978a69ff5998ef4eb28c1307fa).

(I commented out the card_test.jsx since linter was complaining about it, my changes are all in the Card itself)

PR #702, opened 2020-04-08
Add filter templates + React Version

kre8sions:
#### Breaking Changes

#### Checklist:


app/pb_kits/playbook/pb_filter/_filter.jsx
```diff
@@ -1,86 +1,222 @@
...
+  background?: Boolean,
   className?: String,
   data?: String,
   id?: String,
   children?: React.Node,
   filters?: Array<Hash>,
-  sortMenu?: Array<Hash>
+  results?: Number,
+  sortMenu?: Array<Hash>,
+  template?: 'default' | 'filter_only' | 'sort_only' | 'single'
```
thestephenmarshall, line 28: Nice work! Quick question here: are all of these props truly optional? `?` indicates it doesn't have to be passed to the kit in order for the kit to function. If so, then the deconstructed prop list below should set defaults for every optional prop.

kre8sions: @thestephenmarshall they are optional for the reason you can have only filters or only sorting , you can show results if you choose and the template is set to default if nothing is given.

jasoncypret: Makes sense

PR #746, opened 2020-04-20
Adding card header prop to card kit & category colors

christinaatai:
#### Runway Ticket URL
https://nitro.powerhrg.com/runway/backlog_items/NUX-778

#### Images
Card header prop
Screen Shot 2020-04-21 at 12 35 49 PM

Category colors
Screen Shot 2020-04-21 at 12 33 04 PM

#### Breaking Changes

#### Checklist:


app/pb_kits/playbook/pb_card/_card.jsx
```diff
@@ -18,6 +18,50 @@ type CardPropTypes = {
...
+  color?: 'category_1' | 'category_2' | 'category_3' | 'category_4' | 'category_5' | 'category_6' | 'category_7' | 'category_8' |
+          'category_9' | 'category_10' | 'category_11' | 'category_12' | 'category_13' | 'category_14' | 'category_15' | 'category_16' |
+          'category_17' | 'category_18' | 'category_19' | 'category_20' | 'category_21',
+}
+
+type CardBodyProps = {
+  children: Array<React.ReactNode> | React.ReactNode | String,
+  className?: String,
+  padding?: 'none' | 'xs' | 'sm' | 'md' | 'lg' | 'xl',
+}
```
thestephenmarshall, line 34: We move over the types from `card_test.jsx`

app/pb_kits/playbook/pb_card/_card.jsx
```diff
@@ -18,6 +18,50 @@ type CardPropTypes = {
...
+  padding = 'sm',
+}: CardBodyProps) => {
+  const bodyCSS = buildCss('pb_card_body_kit', padding)
+  return (
+    <div className={classnames(bodyCSS, className)}>
+      {children}
+    </div>
+  )
+}
+
```
thestephenmarshall, line 64: Move the sub-components from `card_test.jsx`

app/pb_kits/playbook/pb_card/_card.jsx
```diff
@@ -36,13 +80,30 @@ const Card = ({
     [`highlight_${highlight.color}`]: highlight.color,
   })
 
+  // coerce to array
+  const leChildren = typeof(children) === 'object' && children.length ? children : [children]
```
thestephenmarshall, line 84: We need to ensure that (if we are going to loop through a *list* of children) we indeed always have a list of children 😀

app/pb_kits/playbook/pb_card/_card.jsx
```diff
@@ -36,13 +80,30 @@ const Card = ({
...
+
+  const subComponentTags = (tagName) => {
+    return leChildren.filter((c) => {
+      return c.type && c.type.displayName === tagName
+    }).map((child, i) => {
+      return React.cloneElement(child, { key: `${tagName.toLowerCase()}-${i}` })
+    })
+  }
```
thestephenmarshall, line 92: Reuseable method that can be leverage later down the road as well. This method just needs the argument `tagName` (e.g. "Header") which correlates to the `Card.Header` subcomponent.

app/pb_kits/playbook/pb_card/_card.jsx
```diff
@@ -36,13 +80,30 @@ const Card = ({
...
+
+  const nonHeaderChildren = leChildren.filter((child) => !child.type || child.type.displayName !== 'Header')
```
thestephenmarshall, line 94: This variable contains the non-Header children so that we may separate them from the header positionally. This means that even if the `Card.Header` tag is passed in as the last child, it will always be at the top..

app/pb_kits/playbook/pb_card/_card.jsx
```diff
@@ -36,13 +80,30 @@ const Card = ({
...
+
   return (
     <div className={classnames(cardCss, className)}>
+      {subComponentTags('Header')}
```
thestephenmarshall, line 98: ...always at the top ^

app/pb_kits/playbook/pb_card/_card.jsx
```diff
@@ -36,13 +80,30 @@ const Card = ({
...
       <div className={bodyCSS}>
-        {children}
+        {nonHeaderChildren}
```
thestephenmarshall, line 100: Children that are not `type === 'Header'`

app/pb_kits/playbook/pb_card/_card.jsx
```diff
@@ -36,13 +80,30 @@ const Card = ({
...
       </div>
     </div>
   )
 }
 
+Card.Header = Header
+Card.Body = Body
```
thestephenmarshall, line 107: Just like `card_test.jsx`, we simply assign the subcomponents to `Card` here

drborges:
From the screenshot, it seems that the header's `background-color` does not match the `border-color`, giving the impression that the background-color does not fill up the container entirely:

Screen Shot 2020-04-21 at 2 28 55 PM

Would it make more sense to match the header's border color with its background color?

christinaatai:
@drborges I did a quick mockup of how cards would look like if the card border color was the same as the header bg color.
Screen Shot 2020-04-21 at 2 18 04 PM
 I feel like the border stands out too much. Thoughts on this?

drborges:
@christinaatai, sorry, I meant only changing the border color for the header not the entire card container.

Basically the container (wrapping header + body) would not have a border at all, the header would have the border matching its background-color and the card body would have the regular border color (gray?). By doing that, the header and body borders would align pixel-wise. Does that make sense?

app/pb_kits/playbook/pb_card/_card.jsx
```diff
@@ -36,13 +79,30 @@ const Card = ({
...
+    }).map((child, i) => {
+      return React.cloneElement(child, { key: `${tagName.toLowerCase()}-${i}` })
+    })
+  }
+
+  const nonHeaderChildren = cardChildren.filter((child) => !child.type || child.type.displayName !== 'Header')
+
   return (
     <div className={classnames(cardCss, className)}>
+      {subComponentTags('Header')}
```
drborges, line 97: What if we pass the actual reference to the component so we don't have to rely on `type.display_name` string comparisons?

```suggestion
      {subComponentTags(Header)}
```

Then type checking would simply be:

```js
const subComponentTags = Type => cardChildren.filter(c => c.type === Type)
                                             .map((child, i) => React.cloneElement(child, { key: `${Type.displayName.toLowerCase()}-${i}` }))

const nonHeaderChildren = cardChildren.filter(child => child.type !== Header)
```

christinaatai:
@drborges Ohh yes, makes sense!

app/pb_kits/playbook/pb_card/_card.jsx
christinaatai: I implemented your suggested and the card header wouldn't render. I think we need to rely on string comparisons for React subcomponents, as all the other examples I found online uses string comparison as well.

christinaatai:
@drborges I timeboxed card border color to be the same as header bg color, and it's out of my scope (coding skills and initial scope for card headers). We can potentially come back to this in the future.

However, I was able to make the rounded corners of card header to be more snug.
Screen Shot 2020-04-22 at 8 21 57 AM

app/pb_kits/playbook/pb_card/_card.jsx
```diff
@@ -14,10 +13,54 @@ type CardPropTypes = {
...
   dark?: Boolean,
 }
 
+type CardHeaderProps = {
+  children: Array<React.ReactNode> | React.ReactNode,
+  className?: String,
+  padding?: 'none' | 'xs' | 'sm' | 'md' | 'lg' | 'xl',
+  color?: 'category_1' | 'category_2' | 'category_3' | 'category_4' | 'category_5' | 'category_6' | 'category_7' | 'category_8' |
+          'category_9' | 'category_10' | 'category_11' | 'category_12' | 'category_13' | 'category_14' | 'category_15' | 'category_16' |
+          'category_17' | 'category_18' | 'category_19' | 'category_20' | 'category_21',
```
jasoncypret, line 26: Thoughts about making this `category: 1` instead. I feel like that's closer to what we're doing with spacing

app/pb_kits/playbook/pb_card/_card.jsx
```diff
@@ -36,13 +79,30 @@ const Card = ({
     [`highlight_${highlight.color}`]: highlight.color,
```
jasoncypret, line 79: Should highlights also take category colors? Could highlights be used as categories in some cases?

app/pb_kits/playbook/pb_card/card_header.rb
```diff
@@ -0,0 +1,25 @@
...
+        partial "pb_card/child_kits/card_header"
+  
+        prop :padding, type: Playbook::Props::Enum,
+                       values: %w[none xs sm md lg xl],
+                       default: "none"
+        prop :color, type: Playbook::Props::Enum,
+                     values: %w[category_1 category_2 category_3 category_4 category_5 category_6 category_7 
+                     category_8 category_9 category_10 category_11 category_12 category_13 category_14 category_15
+                     category_16 category_17 category_18 category_19 category_20 category_21],
+                     default: "category_1"
```
jasoncypret, line 17: Same thought here. If we are leaning towards color being category only could this be simplified?

christinaatai:
@jasoncypret I don't think I can update spec test doc because card header is a sub/child component and spec doc only tests the main container.

christinaatai:
@jasoncypret Or would I just make another spec doc for card_header and card_body?

app/pb_kits/playbook/pb_card/_card.jsx
christinaatai: I think highlight can take category colors. There's one _distant_ use case I can think of - course listing on Learning Dojo (admin side). Even though it uses table row side highlight, I can see it working with card side highlight with category colors.

jasoncypret: Sounds good.

let's chat at dailies...perhaps there's a better way to define the options since our list is really long

PR #749, opened 2020-04-22
Limit webpacker CPU & memory on local dev

thestephenmarshall:
Playbook will currently run with as much resource as it possibly can which can potentially cause over-allocation and all of the not-so-happy side effects that come along (hanging, unresponsiveness and even crashing). For this reason, I created this PR which adds resource restrictions to the Docker `webpacker` instance during local development (based on the fairly standard Power-issued dev Macbook Pro):

- CPU: min 2 / max 4
- Memory: min 2G / max 6G (~4-5G is normal w/ no latency)

This first pass is more lenient than perhaps we may end up settling on. 

¯\\\_(ツ)_/¯

Screen Shot 2020-04-21 at 11 08 55 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 # Unreleased
+- Limit webpacker CPU & memory on local dev ([#749](https://github.com/powerhome/playbook/pull/749) @thestephenmarshall)
 
 ## [4.12.0] 2020-4-17
 - New Kit - File Upload Kit ([#730](https://github.com/powerhome/playbook/pull/730) @thestephenmarshall)
--- a/config/webpack/loaders/erb.js
+++ b/config/webpack/loaders/erb.js
@@ -6,6 +6,10 @@ module.exports = {
     loader: 'rails-erb-loader',
     options: {
       runner: (/^win/.test(process.platform) ? 'ruby ' : '') + 'bin/rails runner',
+      env: {
+        ...process.env,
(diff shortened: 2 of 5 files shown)


Jenkinsfile
```diff
@@ -11,7 +11,7 @@ app.build(
     requestCpu: '1',
     limitCpu: '2',
     requestMemory: '6Gi',
-    limitMemory: '12Gi',
+    limitMemory: '8Gi',
```
benlangfeld, line 14: Does the compilation fit into this much memory? Unlike CPU, exceeding this will cause your builds to be aborted, not to run slower.

benlangfeld: Turns out no, it doesn't fit. This PR's CI build was killed because of OOM. Charts in [HQ](https://grafana-hq.powerapp.cloud/d/hYTBQhQiz/pod-resource-usage?orgId=1&refresh=10s&from=now-2d&to=now&var-namespace=jenkins-builds&var-filter=jenkins-powerhome-playbook&var-pod=All) and [Gotham](https://grafana-gm.powerapp.cloud/d/hYTBQhQiz/pod-resource-usage?orgId=1&refresh=10s&from=now-2d&to=now&var-namespace=jenkins-builds&var-filter=jenkins-powerhome-playbook&var-pod=All) show that memory utilisation generally reaches up to around 11.5Gi, so if anything you'll need to increase this, not reduce it.

Of course, making other fixes that reduce the actual memory consumption would be great, but you'd have to do that and watch that consumption fall on those charts before you could lower the limit.

thestephenmarshall: Good to know. I will back this one out.

thestephenmarshall:
@benlangfeld thanks for the input. We would really like to decrease from `12Gi` as that seems like far too much allocation for what Playbook is offering. There will be a follow-up PR to start addressing that.

PR #778, opened 2020-05-13
Upgrade react-popper lib to latest

thestephenmarshall:
#### Runway Ticket URL

https://nitro.powerhrg.com/runway/backlog_items/NUX-960

#### Breaking Changes

None

N/A

#### Checklist:


app/pb_kits/playbook/pb_popover/_popover.jsx
```diff
@@ -73,13 +79,13 @@ const Popover = ({
...
   static defaultProps = {
-    modifiers: {},
+    modifiers: [],
     offset: false,
     placement: 'left',
     portal: 'body',
     show: false,
     shouldClosePopover: noop,
-    usePortal: false,
+    usePortal: true,
```
thestephenmarshall, line 88: We should use this by default everywhere for this kit. This means that now the popover content will be sourced in as the last element inside `<body>` ensuring that no z-index issues occur.

app/pb_kits/playbook/pb_popover/_popover.jsx
```diff
@@ -156,13 +161,15 @@ export default class PbReactPopover extends React.Component<PbPopoverProps> {
...
           </PopperReference>
         </If>
-        <If condition={usePortal}>
-          {ReactDOM.createPortal(
-            popoverComponent,
-            document.querySelector(portal)
-          )}
-          <Else />
-          {popoverComponent}
+        <If condition={show}>
```
thestephenmarshall, line 164: The only other alternative here is not a good one at all and is potentially the reason for the memory consumption. That is using `forceUpdate` which will cause render recursion.

app/pb_kits/playbook/pb_popover/docs/index.js
```diff
@@ -1,7 +1,6 @@
 export { default as PopoverDefault } from './_popover_default.jsx'
 export { default as PopoverWithButton } from './_popover_with_button.jsx'
 export { default as PopoverList } from './_popover_list.jsx'
-export { default as PopoverPortal } from './_popover_portal.jsx'
```
thestephenmarshall, line 4: We do not need an example for this since it's default behavior now.

PR #789, opened 2020-05-18
Filter and table page example

evalouderback:
#### Breaking Changes

#### Checklist:


app/pb_kits/playbook/data/menu.yml
```diff
@@ -87,4 +88,3 @@ kits:
     - title_detail
     - weekday_stacked
   - user
-  - user_badge
```
thestephenmarshall, line 90: This will need to be added back.

Issue #791, opened 2020-05-18
Text Input kit does not support data & aria props in playbook


garettarrowood:
Aria & Data props are available but text input is not implementing the correctly

#### Related Runway Ticket

https://nitro.powerhrg.com/runway/backlog_items/NUX-977

garettarrowood:
Resolved in #812

PR #798, opened 2020-05-20
Fix and add aria, data, id props in text input kit

kellyeryan:
#### Screens

Screen Shot 2020-05-20 at 12 47 53 PM

Screen Shot 2020-05-20 at 12 46 24 PM

#### Breaking Changes

Yes. I made changes so that aria, id, and data props are the same between the React and Ruby sides. 

There were some discrepancies in the code before I started working on it. For example, id was missing from the props on the React side, so it had to be added; additionally, Ruby was passing id and a custom classname in the input tag (when it should be the hardcoded "text_input" classname as it was in React), but it made more sense to put them in the outer div.

Note: validation_data should still be working properly on the Ruby side, even though the method was renamed. It had to be renamed, because it prevented custom data from being passed.

`{...props}` is still getting passed into the actual `<input>` on the React side, which means anything passed to `TextInput` will make it in there. This doesn't seem right, but I've left it.

Let me know if you'd like any changes.

#### Github Issue

#791 

#### Checklist:

- [ ] **SPECS** Please cover your changes with specs - N/A


app/pb_kits/playbook/pb_text_input/_text_input.html.erb
```diff
@@ -1,5 +1,8 @@
 <%= content_tag(:div,
-    class: object.classname) do %>
+    aria: object.aria,
+    class: object.classname,
+    data: object.data,
+    id: object.id) do %>
```
jasoncypret, line 5: I'm not sure about the id change. Going to check with some folks but I believe the ID will be needed for the input in many cases, not the wrapping div. An example is select: https://github.com/powerhome/playbook/blob/v4.15.0/app/pb_kits/playbook/pb_select/_select.html.erb#L22

drborges:
> {...props} is still getting passed into the actual <input> on the React side, which means anything passed to TextInput will make it in there. This doesn't seem right, but I've left it.

That's actually by design, you can find more info on https://github.com/powerhome/playbook/pull/772.

garettarrowood:
@drborges - That design on the React side means there would be duplicate classnames, data, id & aria on the outer div. Should none of those props be present on the outer div (React)? If so, for the Ruby implementation to match, `classname` would need to be removed from the outer div. And custom `data` would have to be merged into `validation_data`.

drborges:
> That design on the React side means there would be duplicate classnames, data, id & aria on the outer div

It does not seem to be the case @garettarrowood, those props are being destructed, thus they won't be present within the `props` variable being spread over the underlying `input`.

Did I get that right?

app/pb_kits/playbook/pb_text_input/_text_input.html.erb
thestephenmarshall, line 5:
```suggestion
    data: object.data do %>
```

app/pb_kits/playbook/pb_text_input/_text_input.html.erb
```diff
@@ -12,9 +15,8 @@
       <% end %>
     <% else %>
       <%= tag(:input,
-          id: object.id,
-          class: object.classname,
-          data: object.data,
+          class: "text_input",
```
thestephenmarshall, line 18:
```suggestion
          id: object.id,
          class: "text_input",
```

garettarrowood:
Ah, yes. I was too vague in my comment, as I was referring to the projected changes in this PR. And also I was wrong about classname, because JS props deconstruction will ignore the custom classname on the input itself. For clarity's sake, and please correct me where this is off, here is a breakdown of how aria, classname, data, & id compare between the Ruby and React versions of this kit.

## `master`

### Ruby 
#### outer div
custom classname
#### input 
duplicate custom classname
id
data overwritten by validation prop object

#### problems
duplicate custom classname
missing aria
inability to pass data

### React
#### outer div
custom classname
#### input
classname of "text_input"
aria (but ignoring `buildPrefixedProps` convention)
data (but ignoring `buildPrefixedProps` convention)
id

#### problems
maybe nothing, other than change in convention for passing data and aria

## As things stand now on this PR

### Ruby
#### outer div
aria
custom classname
data
id
#### input
classname of "text_input"
data overwritten by validation prop object

#### problems
id needs to be moved to input

### React
#### outer div
aria
custom classname
data
id
#### input
classname of "text_input"
aria (but ignoring `buildPrefixedProps` convention)
data (but ignoring `buildPrefixedProps` convention)
id

#### problems
everything is duplicated


## Projected ideal design

### outer div
aria
data
custom classname

### input
id
classname of "text_input"

## Remaining questions
1. Do we want the validation prop? It's only on the Ruby implementation and not the React one. Should this be removed?
1. Do we want to remove the ability to pass aria and data to the input in React? Do we ignore the repetition, and tolerate that we have duplicate data and aria with the outer div?
1. Do we care that aria and data are not handled the same way for React on kits with inputs?

drborges:
Very good points there, @garettarrowood, I very much agree with the projected ideal design.

The `id` being placed on the `input` is aligned with the points I've made previously as well 👍 

> 1. Do we want the validation prop? It's only on the Ruby implementation and not the React one. Should this be removed?

Looking at the implementation, it seems that the `validation` prop is an abstraction for setting the `pattern` attribute of an `input` and the validation `message`.

Again, back to the proposal of bringing Playbook form elements API as close as possible to the standard React form elements API, I'd prefer to drop the `validation` prop entirely and let users combine `pattern` and `error` props in order to achieve the same result.

With that said, before making any calls, it would be nice to better understand the current use cases for the `validation` prop. cc @powerhome/ux-team.

thestephenmarshall:
Great discussion here. Let's bump this one over to `4.16.0` (next week) so that we can continue to discuss. 👍

kellyeryan:
Rebasing this branch against `v4.16.0` (and also I tried `master`) produces merge conflicts in `milano.yml`, `config/deploy/staging/values.yaml`, `config/deploy/production/values.yaml`, `config/deploy/prs/values.yaml` and `bin/deploy`. I don't really know anything about these files. I will open up a new PR and cherry-pick my commits.

I'm sorry to lose all of the comments on this PR though. Is there way to rebase a PR from a cut version of Playbook to the unreleased version of Playbook? I'm new to this. Could documentation be added for how to accomplish this?

PR #838, opened 2020-06-04
Add id, aria, data props to React Currency kit, aria props to Rails

kellyeryan:
#### Issues

React Currency kit did not have aria, data, or id props. The Rails Currency kit did not include aria props. There we also a few minor inconsistencies between the Rails and React kits that I corrected.

#### Screenshots

React

I added to the examples aria, data, id props to show that they are working in the html. I removed them before pushing up the PR.

Screen Shot 2020-06-04 at 11 40 14 AM

Screen Shot 2020-06-04 at 11 40 24 AM

Screen Shot 2020-06-04 at 11 40 42 AM

Screen Shot 2020-06-04 at 11 40 34 AM

Rails

Screen Shot 2020-06-04 at 11 54 11 AM

Screen Shot 2020-06-04 at 11 54 21 AM

Screen Shot 2020-06-04 at 11 54 30 AM

Screen Shot 2020-06-04 at 11 54 41 AM

#### Breaking Changes

None

#### Checklist:


app/pb_kits/playbook/pb_currency/docs/_currency_small.jsx
```diff
@@ -3,24 +3,27 @@ import { Currency } from '../../'
 
 const CurrencySmall = () => {
   return (
-    <div>
+    <>
```
thestephenmarshall, line 6: Since we use bootstrap and may have other styles which could affect layout in our main application, this would potentially be a breaking change. I would suggest leaving the `<div>` in place here though we are not against the use of React fragments in new kits.

kellyeryan: OK! I thought since it was just in the documentation, it would not be a breaking change. I'll change it and keep it in mind for the future.

PR #853, opened 2020-06-11
Add aria, data, id props to React Title kit and aria to Rails Title kit

kellyeryan:
#### Issue

React Title kit was missing aria, data, id props and Rails Title kit was missing aria props.

#### Screens

RAILS

Screen Shot 2020-06-11 at 10 29 24 AM

Screen Shot 2020-06-11 at 10 29 30 AM

REACT

Screen Shot 2020-06-11 at 10 29 59 AM

Screen Shot 2020-06-11 at 10 30 04 AM

#### Breaking Changes

None

#### Checklist:

### DIFF
--- a/app/pb_kits/playbook/pb_title/_title.jsx
+++ b/app/pb_kits/playbook/pb_title/_title.jsx
@@ -13,7 +13,7 @@ type TitleProps = {
   data?: object,
   id?: String,
   size?: 1 | 2 | 3 | 4,
-  tag?: "h1" | "h2" | "h3" | "h4" | "h5" | "h6" | "p" | "div" | "span",
+  tag?: "h1" | "h2" | "h3" | "h4" | "h5" | "h6" | "div",
   text?: String,
 }
 
--- a/app/pb_kits/playbook/pb_title/_title.jsx
+++ b/app/pb_kits/playbook/pb_title/_title.jsx
@@ -2,38 +2,45 @@
 
 import React from 'react'
 import classnames from 'classnames'
+import { buildAriaProps, buildCss, buildDataProps } from '../utilities/props'
 import { spacing } from '../utilities/spacing.js'
 
 type TitleProps = {
+  aria?: object,
   className?: String,
   children?: Array<React.ReactNode> | React.ReactNode,
   dark?: Boolean,
+  data?: object,
+  id?: String,
   size?: 1 | 2 | 3 | 4,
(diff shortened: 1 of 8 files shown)


app/pb_kits/playbook/pb_title/_title.jsx
```diff
@@ -2,38 +2,45 @@
...
 
 type TitleProps = {
+  aria?: object,
   className?: String,
   children?: Array<React.ReactNode> | React.ReactNode,
   dark?: Boolean,
+  data?: object,
+  id?: String,
   size?: 1 | 2 | 3 | 4,
+  tag?: "h1" | "h2" | "h3" | "h4" | "h5" | "h6" | "p" | "div" | "span",
```
thestephenmarshall, line 16: What might the usecase be for `tag="p"`? Would we want to instead settle for `div` instead of `span` or is there a usecase for both?

kellyeryan: Hi, Thanks for the question. I'm not sure what the usecase would be for either; it does seem a little strange for a Title to be in a "span" or "p" tag. I just put these in to mimic the rails implementation.  

https://github.com/powerhome/playbook/blob/1f34586588b17ed9ede3a74d282807d114e5a51e/app/pb_kits/playbook/pb_title/title.rb#L15-L16

If you'd like, I can fix both sides to remove the "p" and "span" tags.

thestephenmarshall: That would be great. Thank you!

app/pb_kits/playbook/pb_title/title.rb
```diff
@@ -13,7 +13,7 @@ class Title
                   values: [1, 2, 3, 4],
                   default: 3
       prop :tag, type: Playbook::Props::Enum,
-                 values: %w[h1 h2 h3 h4 h5 h6 p div span],
+                 values: %w[h1 h2 h3 h4 h5 h6 div],
```
jasperfurniss, line 16: This could possibly break Nitro kits that are using the `tag` prop as `p` or `span`. See this for more: https://github.com/powerhome/playbook/pull/860/files#r442490239

kellyeryan: Hi! Thanks for the comment. I'm good with whatever you think should happen so if you think that I should put the tags back, I'll do it.

PR #860, opened 2020-06-12
Add props to Caption

kellyeryan:
####

React Caption kit was missing aria, data, id props. There were also some enum values missing from the tag prop.

#### Screens

Screen Shot 2020-06-12 at 11 30 49 AM

#### Breaking Changes

None

#### Checklist:


app/pb_kits/playbook/pb_caption/_caption.jsx
```diff
@@ -2,29 +2,37 @@
...
+  aria?: object,
   className?: String,
   children: Array<React.ReactNode> | React.ReactNode,
   dark?: Boolean,
+  data?: object,
+  id?: String,
   size?: "xs" | "sm" | "md" | "lg" | "xl",
-  tag: String,
-  text: String,
+  tag?: "h1" | "h2" | "h3" | "h4" | "h5" | "h6" | "p" | "span" | "div",
```
jasperfurniss, line 16: This could be a possible breaking change... if people are already using this Caption kit on Nitro and are using a tag prop with a different element than one in your given list, it would break that kit instance. @thestephenmarshall @jasoncypret Thoughts on this one. I can check Nitro manually-- but that might take a while to go through every kit usage of Caption in react, depending on how many there are.

jasoncypret: Yes you could argue it is...but I can't imagine any other tags we would want to endable

kellyeryan: Hi! Thanks for your comment. For what it's worth, I added these tags in to reflect the Ruby implementation.

https://github.com/powerhome/playbook/blob/5d626afa846a450d3822c19eb57a1a9e4dc36c2f/app/pb_kits/playbook/pb_caption/caption.rb#L14-L16

I am happy to change it back.

thestephenmarshall: @jasperfurniss these will work more like very shoddy IDE warnings at best for the React side. They are only suggestions and will allow any string that can be mapped to a JSX node in the end. If this were Typescript, you might see a compiler error if you tried to wander outside the enumerable values given.

thestephenmarshall: Summary: 

ship it

PR #870, opened 2020-06-16
Add ability to toggle if the legend is clickable

haydenrou:
This PR allows for the legend's click functionality on the Bar and Line graph to be toggled on and off.

I have set the default to true because it is being used in several places around Nitro and I don't want to have to change them all.

### DIFF
--- a/app/pb_kits/playbook/pb_bar_graph/_bar_graph.jsx
+++ b/app/pb_kits/playbook/pb_bar_graph/_bar_graph.jsx
@@ -21,7 +21,7 @@ type BarGraphProps = {
   title: String,
   type?: String,
   legend?: Boolean,
-  legendClick?: Boolean,
+  toggleLegendClick?: Boolean,
   height?: String,
 }
 
@@ -30,7 +30,7 @@ export default class BarGraph extends React.Component<BarGraphProps> {
     className: 'pb_bar_graph',
     type: 'column',
     legend: false,
-    legendClick: true,
+    toggleLegendClick: true,
   }
 
   componentDidMount() {
@@ -48,7 +48,7 @@ export default class BarGraph extends React.Component<BarGraphProps> {
       type,
       legend,
       height,
-      legendClick,
+      toggleLegendClick,
     } = this.props
 
     new pbChart(`.${className}`, {
@@ -63,7 +63,7 @@ export default class BarGraph extends React.Component<BarGraphProps> {
       yAxisMin: yAxisMin,
       yAxisMax: yAxisMax,
       legend: legend,
-      legendClick: legendClick,
+      toggleLegendClick: toggleLegendClick,
       height: height,
     })
   }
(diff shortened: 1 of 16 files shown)


app/pb_kits/playbook/pb_bar_graph/_bar_graph.jsx
```diff
@@ -21,13 +21,16 @@ type BarGraphProps = {
...
+  legendClick?: Boolean,
   height?: String,
 }
 
 export default class BarGraph extends React.Component<BarGraphProps> {
   static defaultProps = {
     className: 'pb_bar_graph',
     type: 'column',
+    legend: false,
+    legendClick: true,
```
thestephenmarshall, line 33: @jasoncypret should we instead keep the existing functionality and make `legendClick: false` by default?

app/pb_kits/playbook/pb_bar_graph/docs/_bar_graph_legend_non_clickable.jsx
```diff
@@ -0,0 +1,24 @@
...
+}]
+
+const BarGraphLegendNonClickable = () => (
+  <div>
+    <BarGraph
+        axisTitle="Number of Employees"
+        chartData={chartData}
+        id="bar-test-3"
+        legend
+        legendClick={false}
```
thestephenmarshall, line 16: You wouldn't need this here if `false` by default

jasoncypret: Yes, I agree. Let's keep existing default so no clients are disrupted

app/pb_kits/playbook/pb_bar_graph/_bar_graph.jsx
haydenrou: There was a discussion in the playbook connect chat the other day about this. When setting a prop that will change default behaviour, it's dangerous to existing functionality within Nitro. If we make this `false` by default, I'm going to have to submit a PR changing _everywhere_ it's being used in Nitro. That or we leave it as an opt-in.

The reason I'm concerned about this is because the functionality for `legend` was set to default to `false` last week, and nitro was not touched. This has caused a customer facing page to lose the legend, making the graph confusing to the customer.

Thoughts?

app/pb_kits/playbook/pb_bar_graph/docs/_bar_graph_legend_non_clickable.jsx
haydenrou: Per my above comment, the current behaviour is that it is enabled, hence why I set the default to `true`.

app/pb_kits/playbook/pb_bar_graph/_bar_graph.jsx
thestephenmarshall: Thanks for the quick reply! Ok, based on this info, that doesn't sound good then. I had this backwards. LGTM

PR #884, opened 2020-06-25
Bumps popover z-index above Nitro sidebar

jasperfurniss:
#### Screens
Screen Shot 2020-06-25 at 3 04 07 PM


jasoncypret:
Let's move this fix to Nitro

PR #904, opened 2020-07-10
Add props to Body React kit

kellyeryan:
#### Issue

Body React kit was missing id, aria, data props; Body Rails kit was missing aria props. I also changed the React side to match more closely the Rails side.

#### Screens

Screen Shot 2020-07-10 at 11 07 06 AM

Screen Shot 2020-07-10 at 11 08 03 AM

#### Breaking Changes

None

#### Checklist:

- [X] **SPECS** Please cover your changes with specs - N/A


app/pb_kits/playbook/pb_body/_body.jsx
```diff
@@ -2,50 +2,59 @@
...
-  const colorStyle = color !== '' ? `_${color}` : ''
-
-  const themeStyle = dark === true ? '_dark' : ''
-
-  const statusStyle = status !== '' ? `_${status}` : ''
-
-  return 'pb_body_kit' + colorStyle + themeStyle + statusStyle
+  id?: String,
+  status?: 'negative' | 'neutral' | 'positive',
+  tag?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'p' | 'span' | 'div',
```
jasperfurniss, line 20: This will possibly break things in Nitro. It's a possible breaking change if any Body kit is currently using a tag not in this list. I think let's hold off on releasing this and take some time to go through the 109 occurences of this kit in Nitro, and double-check that none of those instances are using the tag prop in a manner excluded from this list of options. Here is the link to all the occurrences that we will need to look at for regression: https://github.com/powerhome/nitro-web/search?l=JSX&q=%3CBody+tag%3D%22a%22

Let me know if there are further questions. I'm happy to pair on this. It will probably take a half hour at least to go through and make sure. @kellyeryan cc: @thestephenmarshall @jasoncypret

thestephenmarshall: @jasperfurniss AFAIK these are merely suggestions and won't impact anything in Nitro (unless something there has significantly changed that I am unaware of). Technically, someone could send `iframe` and it would attempt to do that since there really isn't any true type guardian, but rather Flow is an advisor. I would love to be proven wrong of course.

jasperfurniss: @thestephenmarshall You're right! I was mixing this one up with the .rb file changes made  previously.

PR #910, opened 2020-07-14
Gauge Kit

bh247484:
#### Screens

Rails Default

React State Update

#### Breaking Changes

N/A

#### Runway Ticket URL

[Link to Runway Story](https://nitro.powerhrg.com/runway/backlog_items/NUX-925)

#### How to test this

Try out doc examples.

#### Checklist:


app/pb_kits/playbook/_playbook.scss
```diff
@@ -81,3 +81,4 @@
 @import 'pb_time_stacked/time_stacked';
 @import 'pb_weekday_stacked/weekday_stacked';
 @import './utilities/spacing';
+@import 'pb_gauge/gauge';
```
jasoncypret, line 84: Can you alphabetize

app/pb_kits/playbook/data/menu.yml
```diff
@@ -18,6 +18,7 @@ kits:
     - line_graph
     - distribution_bar
     - circle_chart
+    - gauge
```
jasoncypret, line 21: Can you alphabetize

app/pb_kits/playbook/pb_dashboard/pbChartsLightTheme.js
```diff
@@ -144,6 +144,20 @@ const highchartsTheme = {
...
+    //     // borderWidth: 0,
+    //     color: colors.text_lt_default,
+    //     // enabled: true,
+    //     style: {
+    //       fontFamily: typography.font_family_base,
+    //       fontWeight: typography.bold,
+    //       fontSize: typography.heading_4,
+    //     },
+    //   },
+    // },
```
jasoncypret, line 159: What's the plan here

app/pb_kits/playbook/pb_gauge/_gauge.jsx
```diff
@@ -0,0 +1,106 @@
...
+          chart.series[0].setData([chartData[0].value])
+        }
+      })
+    } else {
+      componentDidMount.current = true
+    }
+  })
+
+  return (
+    <div
```
jasoncypret, line 97: Spacing here?

app/pb_kits/playbook/plugins/pb_chart.js
```diff
@@ -1,8 +1,12 @@
 import Highcharts from 'highcharts'
```
jasoncypret, line 1: Let's work on getting this separated to keep this clean

app/pb_kits/playbook/pb_dashboard/pbChartsLightTheme.js
bh247484: Managed to get this sorted.  Was a silly order of operations issue.  Moved gauge formatting/style options over into pbChartsLightTheme.

app/pb_kits/playbook/plugins/pb_chart.js
bh247484: Got everything requiring Sass imports moved over into Theme file.

app/pb_kits/playbook/pb_gauge/docs/_gauge_min_max.jsx
```diff
@@ -0,0 +1,23 @@
...
+}]
+
+const GaugeMinMax = () => (
+  <div>
+    <Gauge
+        chartData={data}
+        id="gauge-min-max"
+        // max defaults to 100
+        max={5}
+        // min defaults to 0
```
thestephenmarshall, line 15: Could we maybe add this info to the doc example text itself so that it is present and viewable in the browser?

app/pb_kits/playbook/pb_gauge/_gauge.jsx
```diff
@@ -0,0 +1,105 @@
...
+
+  const css = buildCss({
+    'pb_gauge_kit': true,
+  })
+  // Runs first time component Renders
+  useEffect(() => {
+    chartData.forEach((obj) => {
+      obj.y = obj.value
+      delete obj.value
+    })
```
xjunior, line 63: Since chartData is coming as a prop, I think it's better if you transform this collection into another one

app/pb_kits/playbook/pb_gauge/_gauge.jsx
```diff
@@ -0,0 +1,105 @@
...
+
+    new pbChart('.selector', {
```
xjunior, line 65: Is this an external dom manipulation?

app/pb_kits/playbook/pb_gauge/_gauge.jsx
bh247484: I vaguely remember trying to use .map() and assigning it a new variable like `formattedChartData`.  Seem to remember it not working for whatever reason but let me give it another try.  Have made some structural changes since then so whatever the issue was might have resolved itself.

app/pb_kits/playbook/pb_gauge/_gauge.jsx
bh247484: External to React, yes.  It instantiates a pbChart obj defined in [this file here](https://github.com/powerhome/playbook/blob/9c853a9c306b453deab4d0c1c4943d5c8999d4fe/app/pb_kits/playbook/plugins/pb_chart.js).  It relies on the [Highcharts library](https://www.highcharts.com/demo/gauge-speedometer) which in turn manipulates the dom.  Tried to use the Highcharts api to directly affect/update the chart and simply pass React state to it (to avoid unexpected Highcharts behavior when React re-renders).

xjunior: Have you considered https://github.com/highcharts/highcharts-react ? It's not a good idea to mess with dom manipulation as it's react job to rewrite that whenever it feels it should.

bh247484: Had taken a look at highcharts-react.  It was a breeze to setup in codesandbox.  Ultimately we want to move all our react Highcharts kits (Bar Graph, Circle Chart, Line Graph, Circle Chart) over to use highcharts-react but that effort was outside the scope of this project.  

Agreed though, I ran into many of the issues why it's not a good idea to mess with dom manipulation outside of React's concern as I was building the kit.  Long term we will likely be migrating React kits over to react-highcharts.

app/pb_kits/playbook/pb_gauge/_gauge.jsx
bh247484: Forgot to return the object after reformatting it inside `.map()`.  Thanks for drawing my attention back to this one, feels good to have that mystery resolved.  I will push this fix back up later this week.  Don't want to reset the Milano build, I have it linked as a doc reference for others reviewing the alpha.

app/pb_kits/playbook/pb_gauge/_gauge.jsx
```diff
@@ -0,0 +1,105 @@
...
+      showLabels: showLabels,
+      style: style,
+      tooltipHtml: tooltipHtml,
+      type: 'gauge',
+    })
+  }, [])
+
+  const componentDidMount = useRef(false)
+  // Doesn't run the first time but runs every subsequent render
+  useEffect(() => {
```
drborges, line 84: Is it the intent to have this second `useEffect` executed on **every** subsequent mount or only if the chart data is changed?

app/pb_kits/playbook/pb_gauge/_gauge.jsx
drborges: Why would we want that? I was actually thinking we could use the incoming `chartData` to determine when rerenders are required.

app/pb_kits/playbook/pb_gauge/_gauge.jsx
drborges: If that's outside `React`'s context, where will that `selector` class be defined exactly? I guess I'm trying to understand how [`this.element`](https://github.com/powerhome/playbook/blob/9c853a9c306b453deab4d0c1c4943d5c8999d4fe/app/pb_kits/playbook/plugins/pb_chart.js#L30) fits into the implementation.

app/pb_kits/playbook/pb_gauge/_gauge.jsx
bh247484: Looks like `chartData` still determines rerenders, it will just be reformatted as it's passed in the first time.  Highcharts is particular (and not very semantic) about how data has to be formatted.  Subsequent renders will pass it's value field directly to the highcharts api to [update the chart's data](https://github.com/powerhome/playbook/blob/9c853a9c306b453deab4d0c1c4943d5c8999d4fe/app/pb_kits/playbook/pb_gauge/_gauge.jsx#L88).

drborges: got it 👍🏼

app/pb_kits/playbook/pb_gauge/_gauge.jsx
bh247484: Much of pbChart was inherited code, so I'm not entirely what the purpose of the `'.selector'` class is.  Looks like the instance variable `element` is never even used?

app/pb_kits/playbook/pb_gauge/_gauge.jsx
bh247484: Only if chartData prop is changed, is it possible to scope rerenders to a specific prop?  Maybe by adding it to that second array argument in `useEffect()`?

app/pb_kits/playbook/pb_gauge/_gauge.jsx
bh247484: This has got me thinking though, might there be cases where you'd like to change not just the value in chartData but also its' name?  i.e. `chartData={[{name: 'init name', value: 'init value'}]}` mutating to `chartData={[{name: 'modified name', value: 'modified value'}]}`.  Right now I'm only accounting for changes in value, not changes to name.

bh247484: Updated functionality to include concerns in the comment above.  Now includes the ability to change name as well as value.

drborges: yeah, that's hard to tell without an actual use-case... With that said, given that React components in general, are functions of `data` to `UI`, e.g. `fn: Props -> JSX`, I'd expect that a different set of props would give me a different UI state.

app/pb_kits/playbook/pb_gauge/_gauge.jsx
drborges: that's what I was trying to pinpoint :) not sure either.

bh247484: @jasperfurniss do you know what the `'.selector'` argument in pbChart's constructor does?  It's in there for all highcharts instances in Playbook.  I copy and pasted it when setting up the Gauge kit and never got around to investigating.  Maybe it was originally created to assign a selector to an outer div?  Looks like it was never implemented.

app/pb_kits/playbook/pb_gauge/_gauge.jsx
drborges: That's correct 👍🏼 you can scope rerenders to specific props by passing a dependencies array as the second argument of useEffect with the props. That will tell React to only trigger the effect block when any of the props in that array have their reference updated. [Here is a pretty insightful resource](https://overreacted.io/a-complete-guide-to-useeffect/) on different React hooks topics in case you feel like digging a little deeper in the subject 😁

jasoncypret:
@bh247484 Changed the base to master since we don't know how long testing will be. When it gets closer to releasing we can change it back to the target base.

thestephenmarshall:
@bh247484 Lots of conversation happening here. We might want to sync on this one again.

app/pb_kits/playbook/pb_gauge/_gauge.jsx
jasperfurniss: @bh247484 Just now seeing this in my notifications. 
I don't believe there is a purpose for the `'.selector'` that I'm aware of. I considered removing it myself. We would need to test, and do some digging to see what it actually does or was intended to do.

Issue #878, opened 2020-06-22
Bug on Icon on NavItem (?)


gmfvpereira:
I found two scenarios where I think that the `iconRight` on the NavItem is not working as it should. 

1 - The page breaks if we remove an Icon from a NavItem. 
Use case: The Icon should only appear on the active NavItem.
Example:
<img width="276" alt="Screen Shot 2020-06-22 at 11 21 15 AM" src="https://user-images.githubusercontent.com/386460/85298568-a24a2780-b47a-11ea-8fb6-0f32796f1f99.png">

Error: `Uncaught DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.`

Reproducible with:
```
import React, { useState } from "react"
import { NavItem } from "playbook-ui"

const NavItemBug = () => {
  const [ showIcon, setShowIcon ] = useState(true)

  return (
    <If condition={showIcon}>
      <NavItem
          iconRight="ellipsis-v"
          onClick={() => setShowIcon(!showIcon)}
          text="A nav item"
      />
      <Else />
        <NavItem
            onClick={() => setShowIcon(!showIcon)}
            text="A nav item"
        />
      </If>
  )
}

export default NavItemBug

```

This other usage also raises the same error:
```
<NavItem
  iconRight={showIcon ? "ellipsis-v" : ""}
  onClick={() => setShowIcon(!showIcon)}
  text="A nav item"
/>
```

2 - The icon does not update to a new icon, it remains the same forever:
```
<NavItem
  iconRight={selected ? "angle-down" : "angle-right"}
  onClick={() => setShowIcon(!showIcon)}
  text="A nav item"
/>
```

jasoncypret:
@gmfvpereira Just confirming your planning on fixing (just asked Christina). I don't want overlapping fixes. If so I'll assign to you. Thanks!

gmfvpereira:
We'll fix it! Sorry for the delay on getting back to you @jasoncypret

jasoncypret:
Great work @KatherineMuedas !

PR #800, opened 2020-05-21
Fix button loading

gmfvpereira:
#### Screens

Fixes this bug https://github.com/powerhome/playbook/issues/755 

#### Breaking Changes

None

#### Runway Ticket URL
https://nitro.powerhrg.com/runway/backlog_items/LGA-148

#### Checklist:

- [ ] **DEPLOY** Please add the `Milano` label when you are ready for a review
- [ ] **SCREENSHOT** Please add a screen shot or two
~~**SPECS** Please cover your changes with specs~~
- [x] **CHANGELOG** Please add an entry on `[Unreleased]` section to every functionality `ADDED`/`CHANGED`/`DEPRECATED`/`REMOVED`/`FIXED`

PR #913, opened 2020-07-15
Fix Nav Item Icon to handle conditions for icon

KatherineMuedas:
#### Screens

Fixes this bug #878 

#### Breaking Changes
None

#### Runway Ticket URL
https://nitro.powerhrg.com/runway/backlog_items/LGA-201

#### How to test this
```
import React, { useState } from "react"
import { Nav } from '../../'
import NavItem from '../_item.jsx'

const WithIconsNav = () => {
  const [ showIcon, setShowIcon ] = useState(true)
  const [ currentIcon, setCurrentIcon ] = useState("")
  const handleDisplayCurrentIcon = (icon) => {
      setShowIcon(!showIcon)
      setCurrentIcon(icon)
    }

  return (
    <Nav
        link="#"
        title="Browse"
    >
      <NavItem
        iconRight={currentIcon}
        onClick={() => handleDisplayCurrentIcon(showIcon ? "angle-down" : "angle-right")}
        text="Other item"
      />

    </Nav>
  )
}

export default WithIconsNav
```
#### Checklist:

- **SCREENSHOT** Please add a screen shot or two
- **SPECS** Please cover your changes with specs


KatherineMuedas:
To fix this issue, I had to wrap the icon with a div and also add a key to the div. A pretty similar fix was added previously to Buttons https://github.com/powerhome/playbook/pull/800.  I am wondering if we should add the div and key to the icon component instead of only adding it to the nav item with icon file. If we do that all components that uses icons could benefit from it.

app/pb_kits/playbook/pb_nav/_item.jsx
```diff
@@ -32,21 +32,31 @@ const NavItem = ({
...
       >
         <If condition={iconLeft}>
-          <Icon
-              className="pb_nav_list_item_icon_left"
-              fixedWidth
-              icon={iconLeft}
-          />
+          <div
+              className="pb_nav_list_item_icon_section"
+              key={iconLeft}
```
jasperfurniss, line 37: I'm trying to make sense out of using a key here. I think this will need to be a little more unique. For example, what if two different icons want to use the same icon... Their keys would be the same, and that would create a problem, right?

KatherineMuedas: I see your point. What about if we add ```key={`L-${iconLeft}`}``` and ```key={`R-${iconRight}`}``` ? I think that could solve the problem

gmfvpereira: I think that's not a problem @jasperfurniss since React require unique key only within siblings: https://reactjs.org/docs/reconciliation.html#keys

gmfvpereira: And the usage of the key is to force re-rendering when the icon changes.

PR #953, opened 2020-08-06
Add checkbox to form helper

bh247484:
#### Screens

checkbox

validation

#### Runway Ticket URL

[Runway Ticket](https://nitro.powerhrg.com/runway/backlog_items/NUX-1299)

#### Checklist:

- [x] **CHANGELOG** Add appropriate labels


app/pb_kits/playbook/pb_form/form_builder/checkbox_field.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module Playbook
+  module PbForm
+    module FormBuilder
+      module CheckboxField
+        def check_box(name, data: {}, props: {})
+          value = props[:value] ||= 1
+          input = super(name, { required: props[:required], data: data }, value)
+          label_text = name.to_s.gsub("_", " ")
```
terryfinn, line 10: We use Rails provided label method for this. That provides i18n lookup, and provides a uniform way of generating them label text.
https://github.com/powerhome/playbook/blob/3155894e73deeede3074a735f4bb307e2e9abe92/app/pb_kits/playbook/pb_form/form_builder/form_field_builder.rb#L9
```suggestion
          props[:label] = @template.label(@object_name, name) if props[:label] == true 
```

app/pb_kits/playbook/pb_form/form_builder/checkbox_field.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module Playbook
+  module PbForm
+    module FormBuilder
+      module CheckboxField
+        def check_box(name, data: {}, props: {})
+          value = props[:value] ||= 1
+          input = super(name, { required: props[:required], data: data }, value)
```
terryfinn, line 9: One of things we worked hard with the other inputs has been to make the form system as thin of a wrapper over what rails provides out of the box as possible, and that means forwarding all none props to the super class. I would suggest you capture them like this.

```suggestion
        def check_box(name, props: {}, **options, &block)
          value = props[:value] ||= 1
          input = super(name, **options, &block)
```

app/pb_kits/playbook/pb_form/form_builder/form_field_builder.rb
```diff
@@ -12,6 +12,7 @@ def initialize(method_name, kit_name:)
             options[:skip_default_ids] = false unless options.key?(:skip_default_ids)
             options[:required] = true if props[:required]
             options[:placeholder] = props[:placeholder] || ""
+            options[:value] = props[:value]
```
terryfinn, line 15: Are you using this?

app/pb_kits/playbook/pb_form/form_builder.rb
```diff
@@ -14,9 +14,11 @@ module FormBuilder
         prepend(FormFieldBuilder.new(:password_field, kit_name: "text_input"))
         prepend(FormFieldBuilder.new(:url_field, kit_name: "text_input"))
         prepend(FormFieldBuilder.new(:text_area, kit_name: "textarea"))
+        # prepend(FormFieldBuilder.new(:check_box, kit_name: "checkbox"))
```
terryfinn, line 17: Remember to remove this if you not keeping.

app/pb_kits/playbook/pb_filter/docs/_filter_default.html.erb
```diff
@@ -28,8 +28,12 @@
     ]
   %>
   <%= pb_rails("form", props: { form_system_options: { scope: :example, method: :get } }) do |form| %>
-    <%= form.text_field :example_text_field, props: { label: true } %>
-    <%= form.collection_select :example_collection_select, example_collection, :value, :name, props: { label: true } %>
+    <%# <%= form.text_field :example_text_field, props: { label: true } %>
+    <%# <%= form.collection_select :example_collection_select, example_collection, :value, :name, props: { label: true } %>
```
terryfinn, line 32: Why comment these out? Are you just trying to stay focused on your testing?

bh247484: Wanted try using it inside the filter (BOP's use case), see how they oriented themselves and what the spacing looked like.  I'll discard these changes once and clean up the docs once it comes time to merge this into Playbook.

app/pb_kits/playbook/pb_form/form_builder/form_field_builder.rb
bh247484: Had it setup such that value would be passed in directly as a prop.  I'll switch it to accept a value field in the options object instead.

app/pb_kits/playbook/pb_form/form_builder/checkbox_field.rb
bh247484: Ahh okay, I'll make those changes.  That super method was a bit of a black box in terms of which arguments it was parsing and in what order.  I'll mostly ended up with line 9 by trial and error.  I'll make those changes.  Should clean things up and help make it more uniform/conventional.

app/pb_kits/playbook/pb_form/form_builder/checkbox_field.rb
bh247484: Think I remember having difficulty implementing that `@template.label` but I may well be confusing that with that date picker form field I've been working on.  I'll give this another shot.

app/pb_kits/playbook/pb_form/form_builder/checkbox_field.rb
bh247484: @terryfinn updated to use the double asterisk options argument.  Everything looks good except the value attribute isn't applied to the html input even after including `options[:value] = props[:value]`.  Where exactly is the `super` method defined?  Would make it easier to debug if I could see how the arguments are being parsed.

bh247484: Quick follow up.  Looks like it's the default behavior to override any value attribute on a checkbox input and replace it with `1`.  So things are working as expected.  Devs will just have to work with a `1` - `0` boolean value system.

jasoncypret:
@bh247484 Do you still need this deployed to Milano, or can I remove the tag?

bh247484:
> @bh247484 Do you still need this deployed to Milano, or can I remove the tag?

Sorry just saw this.  Wanted to leave it up so that BOP could use the milano deploy doc examples.

PR #974, opened 2020-08-10
Add dark mode to Icon Value

kellyeryan:
#### Runway Ticket URL

NUX-1310

#### Screens

Screen Shot 2020-08-10 at 3 47 14 PM

Screen Shot 2020-08-10 at 3 46 37 PM

#### Breaking Changes

None

#### Checklist:

- [X] **SPECS** Please cover your changes with specs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+- Dark mode added to Icon Value kit ([#974](https://github.com/powerhome/playbook/pull/974)  @kellyeryan)
+
 ### Fixed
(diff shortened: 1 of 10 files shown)


kellyeryan:
> I think the dark prop should be passed to the child kits so they can use the correct color for dark mode

kellyeryan:
Even

thestephenmarshall:
@kellyeryan please add the appropriate labels

jasoncypret:
> @jasoncypret Could you point me towards an example of the color that icon value should use for dark mode? Is it the same as Icon in dark mode? Here is what it currently looks like in dark mode using the dark prop. Is this incorrect?
> 
> Screen Shot 2020-08-12 at 10 16 29 AM

The icon will automatically become the color set by its parent (unless the sass is doing something funky).

kellyeryan:
@jasoncypret So is the code correct or incorrect?

PR #978, opened 2020-08-11
Fix tooltip bug

kellyeryan:
#### Runway Ticket URL

NUX-1227

Below is screenshot of the bug--the tooltips would get stuck open. Since I don't have the original environment in which that was created, I made a table and played around with buttons and tooltips within the table. I think I was able to reproduce the issue. In the course of this, I discovered that there is a problem with using `event.toElement` in the firefox browser that doesn't exist in the chrome browser. When I changed`event.toElement` to `event.target`, the bug went away.

Based on the information I have, this is one possible fix for the bug. It was suggested that perhaps there were duplicate ids for the tooltips on the page and this made all the tooltips open at once, but I was not able to reproduce that. 

#### Screens

This was the bug issue:

tooltips_overlapping

Reproduction of issue:

Screen Shot 2020-08-11 at 1 47 45 PM

#### Checklist:

- [ ] **SPECS** Please cover your changes with specs - NA


coleerikson:
Great fix! Seems like `toElement` is defunct and not well supported.

thestephenmarshall:
I am holding this one back so that I can test it further on my own. I need to better understand why `toElement` isn't the proper choice here. @kellyeryan where did the suggestion to change this originate from?

kellyeryan:
@thestephenmarshall  I tested out the bug in Firefox (the browser I usually use) and was able to replicate the difficulties that the PR demonstrated, but when I tested it in Chrome, it didn't seem like there was a problem. So I looked up the error I was receiving -- `event.toElement` is undefined -- and it suggested using `event.target` because Firefox doesn't support `toElement`. I switched it out and it started working. I tried to duplicate the error again when I was working on the tooltips in the popover kit and it was again a good fix. 

If you want to keep testing it, that's fine; I would just suggest making sure to use the Firefox browser, since that is where I found the difficulties and it is also what nitro uses.

PR #991, opened 2020-08-13
Create online status dark mode

kellyeryan:
#### Runway Ticket URL

NUX-1330

#### Screens

Screen Shot 2020-08-13 at 1 24 30 PM

Screen Shot 2020-08-13 at 1 24 53 PM

#### Breaking Changes

Only additive changes.

#### Checklist:

- [X] **SPECS** Please cover your changes with specs

### DIFF
--- a/app/pb_kits/playbook/pb_online_status/_online_status.jsx
+++ b/app/pb_kits/playbook/pb_online_status/_online_status.jsx
@@ -4,9 +4,10 @@ import React from 'react'
 import classnames from 'classnames'
 import { globalProps } from '../utilities/globalProps.js'
 
-import { buildDataProps } from '../utilities/props'
+import { buildAriaProps, buildCss, buildDataProps } from '../utilities/props'
 
 type OnlineStatusProps = {
+  aria?: object,
   className?: string,
   data?: object,
   id?: string,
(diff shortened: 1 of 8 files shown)


app/pb_kits/playbook/pb_online_status/_online_status.jsx
```diff
@@ -4,24 +4,34 @@ import React from 'react'
...
+    className,
+    data = {},
+    id,
+    status = 'offline',
+  } = props
+
+  const ariaProps = buildAriaProps(aria)
   const dataProps = buildDataProps(data)
-  const css = classnames([`pb_online_status_kit_${status}`, className], globalProps(props))
+  const classes = classnames(buildCss('pb_online_status_kit', status), className, globalProps(props))
```
jasoncypret, line 28: I believe `className` needs to come last so it will work properly with CSS Modules in Nitro. Correct @jasperfurniss ?

kellyeryan: It might have more to do with the parentheses.

jasoncypret, line 28:
```suggestion
  const classes = classnames(buildCss('pb_online_status_kit', status),globalProps(props), className)
```

kellyeryan: @jasoncypret Ok, I'll make the change and push it up right away. But if this is the case--that the `className` needs to be at the end--we will need to change most of the kits because `globalProps(props)` was added to the end of all of the class-building code and therefore `className` won't be at the end.  Does that make sense?

PR #1006, opened 2020-08-19
Updated Date Kit with new styles and props

christinaatai:
#### Screens

Screen Shot 2020-08-20 at 11 23 45 AM
Screen Shot 2020-08-20 at 11 23 53 AM

#### Breaking Changes

Yes, this will be a breaking change. We are removing size "xs" and renaming the prop "value" to "date" in the React kit to match the prop name in the Rails kit.

If we want to keep everything as similar to what is currently in prod as possible, all Date Kits with size="xs" will should be changed so that size prop is removed, "show_day_of_week"/"showDayOfWeek" is true. All remaining Date Kits (of size "sm"), should be changed so that "show_icon"/"showIcon" and "show_day_of_week"/"showDayOfWeek" is true.

#### Runway Ticket URL

https://nitro.powerhrg.com/runway/backlog_items/NUX-638

#### Checklist:


app/pb_kits/playbook/pb_date/_date.jsx
```diff
@@ -2,82 +2,85 @@
...
-  </h3>
-)
+const PbDate = (props: PbDateProps) => {
+  const {
+    date,
+    className,
+    showIcon = false,
+    showDayOfWeek = false,
+    alignment = 'left',
+  } = props
```
kellyeryan, line 25: Maybe you could include data, aria, and id props?

app/pb_kits/playbook/pb_date/_date.html.erb
```diff
@@ -2,13 +2,36 @@
...
     data: object.data,
     class: object.classname) do %>
-    <% if object.size == "lg" %>
-      <%= pb_rails("title", props: { text: object.lg_date, size: 3 }) %>
-    <% elsif object.size == "sm" %>
-      <%= pb_rails("icon", props: { icon: "calendar", fixed_width: true }) %>
-      <%= pb_rails("title", props: { tag: "span", text: object.sm_date, size: 4 }) %>
-    <% else %>
-      <%= pb_rails("title", props: { text: object.xs_date, size: 4 }) %>
+
```
kellyeryan, line 5: Could you add an aria prop here?

app/pb_kits/playbook/pb_date/_date.jsx
coleerikson: Added 👍

app/pb_kits/playbook/pb_date/_date.html.erb
coleerikson: added.

PR #1023, opened 2020-08-24
WIP: Add radio cards to pb_card

stephenmckeon:
#### Screens

[INSERT SCREENSHOT]

#### Breaking Changes

#### Checklist:

- [ ] **SCREENSHOT** Please add a screen shot or two


garettarrowood:
I think the objective of this addition needs some more clarity.

As it stands now, this looks like work in the direction of adding a radio input into a card. 

I would think that something like this would instead be property for a `Radio` input to be wrapped in a card, or a new kit altogether.

@jasoncypret - What do you think?

jasoncypret:
Yes, this is expected to be a new kit, built similar to selectable cards. The only difference is that this kit would show the inputs where the selectable card has those as hidden inputs

garettarrowood:
Merging this functionality into the `CheckboxCard` kit and renaming.

PR #1036, opened 2020-08-26
[NUX-638] - Updated Date Kit with new styles and props

coleerikson:
#### Screens

Screen Shot 2020-08-20 at 11 23 45 AM
Screen Shot 2020-08-20 at 11 23 53 AM

#### Breaking Changes

Yes, this will be a breaking change. We are removing size "xs" and renaming the prop "value" to "date" in the React kit to match the prop name in the Rails kit.

If we want to keep everything as similar to what is currently in prod as possible, all Date Kits with size="xs" will should be changed so that size prop is removed, "show_day_of_week"/"showDayOfWeek" is true. All remaining Date Kits (of size "sm"), should be changed so that "show_icon"/"showIcon" and "show_day_of_week"/"showDayOfWeek" is true.

#### Runway Ticket URL

https://nitro.powerhrg.com/runway/backlog_items/NUX-638

#### Checklist:


app/pb_kits/playbook/pb_kit/pb_date_time.rb
```diff
@@ -36,7 +36,7 @@ def to_year
       end
 
       def to_month
-        @value.strftime("%^b")
+        @value.strftime("%b")
```
coleerikson, line 39: @christinaatai this is cool but it's causing a test to break in rspec.

christinaatai: @coleerikson Let me look into it

PR #1043, opened 2020-08-27
Button & Circle Icon Button Cleanup

megantrimble:
#### Changes
**Button:**

1. Changed pb_button_link_dark variant text to white
2. Removed xl margin and dark prop from Default Primary on React side 
3. Removed duplicate Disabled button on React side
4. Added spacers between react buttons on default, dark, link and loading versions
5. Add spacer in <IF> statement between icon and text/child element on react button block content

**Circle Icon Button:**
1. Link Section had two primary buttons on rails, switched second one to secondary.
2. Added <br/> spacing between the two link buttons

#### Screens

Screen Shot 2020-08-27 at 11 58 49 AM
Screen Shot 2020-08-27 at 11 59 15 AM
Screen Shot 2020-08-27 at 12 01 16 PM
Screen Shot 2020-08-27 at 11 54 35 AM

#### Breaking Changes

Regression testing needed for the spacer placed inside the <IF> statement. Inserted to add spacing between the icon and text/child element on react button block content.

#### Checklist:


app/pb_kits/playbook/pb_button/docs/_button_default.jsx
```diff
@@ -4,26 +4,22 @@ import { Button } from '../../'
 const ButtonDefault = () => (
   <div>
     <Button
-        dark
-        marginRight="xl"
         onClick={() => alert('button clicked!')}
         text="Button Primary"
     />
+    {' '}
```
christinaatai, line 10: Is there a better way to add spacers instead of a `{' '}`?

christinaatai: Have you tried using class pb--doc-demo-row?

megantrimble: I had tried that and it creates them to all be spaced out. I added spacers on react side since rails is displaying the spacing without adding in breaks into the code. If we want to use what you mentioned above across the board for every button example (rails and react) I can add that. But it will no longer left align all the buttons going down the page. Thoughts?

megantrimble: Screen Shot 2020-08-27 at 1 02 17 PM

megantrimble: Could also potentially wrap them in a flex kit as well to create the separation between them.

megantrimble:
@thestephenmarshall I added the regression testing bit to the breaking changes section.

I also tested in FireFox and all is looking well. It is just enforcing a whitespace. Image below.
Screen Shot 2020-08-27 at 1 15 19 PM

PR #1049, opened 2020-08-28
Decrease prs resource allocation

iblackman:
The review envs use almost no resource, there is no need to allocated that much. In order to increase the number of review envs for playbook I'm decreasing the resource allocation.

Screen Shot 2020-08-28 at 2 45 22 PM

#### Screens

[INSERT SCREENSHOT]

#### Breaking Changes

#### Checklist:

- [ ] **SCREENSHOT** Please add a screen shot or two


benlangfeld:
What about resources consumed during upgrades? I don't see any such peaks in the charts. Grab a longer period?

config/deploy/values.yaml
```diff
@@ -1,3 +1,12 @@
 appConfig:
   rackEnv: "production"
   railsEnv: "production"
+resources:
```
benlangfeld, line 4: Why not leave this constant across all environments.

iblackman: because it doesn't use anything and we want to have the resources allocation as real as possible so we can use our resources the best way possible. And since they want to increase from 3 review envs to 10. This would make a big different in the resource allocation.

iblackman: > What about resources consumed during upgrades? I don't see any such peaks in the charts. Grab a longer period?

Even during the spin up of the environment it doesn't use much.
Screen Shot 2020-08-28 at 3 40 01 PM

iblackman: other dashboard as example:
Screen Shot 2020-08-28 at 3 47 24 PM

benlangfeld: > because it doesn't use anything and we want to have the resources allocation as real as possible so we can use our resources the best way possible. And since they want to increase from 3 review envs to 10. This would make a big different in the resource allocation.

But does production really use more?

iblackman: ohh okay. Now I understood your point. I'm gonna check that.

iblackman: yeah. Production doesn't use that either. We can definitely decrease in all envs. I'm gonna adjust the PR @benlangfeld .
Screen Shot 2020-08-28 at 4 05 23 PM
Screen Shot 2020-08-28 at 4 05 17 PM

PR #1080, opened 2020-09-09
Add timezone prop to date, expand docs

jasoncypret:
The Date kit wasn't supporting Timezones. When you added a strict `Date` class it's not a concern but `DateTime` would have problems and default all days to adhere to EST timezone.

- This PR updates the method to leverage existing functionality, the kit just needed a Timezone prop.
- Also, since Dates, Time, Timezones are confusing as is, we needed to add a little more documentation around how our kits leverage those classes in RoR to render dates, times, and timezones.

#### Screens

**Updated details**
title

**Default Example**
date-default-example

**Timezone Examples**
timezone-examples

#### Breaking Changes

Should be ok. New timezone props is optional.

#### Runway Ticket URL
https://nitro.powerhrg.com/runway/backlog_items/NUX-1370

#### How to test this
Use kit, enter various dates and timezones.

#### Checklist:

### DIFF
--- a/app/pb_kits/playbook/pb_date/date.rb
+++ b/app/pb_kits/playbook/pb_date/date.rb
@@ -13,6 +13,7 @@ module Playbook
       prop :size, type: Playbook::Props::Enum,
                   values: %w[lg sm xs],
                   default: "sm"
+      prop :timezone, default: "America/New_York"
 
       def classname
         generate_classname("pb_date_kit")
@@ -33,7 +34,7 @@ module Playbook
     private
 
       def pb_date_time
-        Playbook::PbKit::PbDateTime.new(date)
+        Playbook::PbKit::PbDateTime.new(date, timezone)
       end
     end
   end
--- a/app/pb_kits/playbook/pb_date/docs/_date_default.html.erb
+++ b/app/pb_kits/playbook/pb_date/docs/_date_default.html.erb
@@ -12,6 +12,13 @@
 <br>
 
 <%= pb_rails("date", props: {
-  date: "2012-08-02T15:49:29Z",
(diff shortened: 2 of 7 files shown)


bh247484 (approved):

app/pb_kits/playbook/pb_date/date.rb
```diff
@@ -33,7 +34,7 @@ def sm_date
     private
 
       def pb_date_time
-        Playbook::PbKit::PbDateTime.new(date)
+        Playbook::PbKit::PbDateTime.new(date, timezone)
```
jasoncypret, line 37: @coleerikson I wonder if someone would have passed timezone with date previously and this would cause it to pass `PbDateTime.new(date + timezone, timezone)`. Do you think this would cause a problem with the kit?

coleerikson: **I think** if they passed the TZ, it's being ignored.

It's leveraging this: https://github.com/powerhome/playbook/blob/6.6.0/app/pb_kits/playbook/pb_kit/pb_date_time.rb#L8

Which has a second parameter, but as you can see it's not using it, therefore it's defaulting to EST.

coleerikson:
@jasoncypret Here's an example of a timezone sensitive value being passed to the date kit.

With out timezone, it's defaulting to East Coast, just like before. With timezone, it's checking the DateTime in that TZ which is tomorrow.

missing-timezone

coleerikson:
Here, I removed the `zone` parameter on `Playbook::PbKit::PbDateTime.new(date)`. And as you can see, just the value with TZ doesn't work as expected.

timezone-bug

PR #1106, opened 2020-09-17
React Circle Chart

jasperfurniss:
#### Screens
Screen Shot 2020-09-17 at 9 21 33 AM

#### Breaking Changes
No.

#### Runway Ticket URL
https://nitro.powerhrg.com/runway/backlog_items/NUXE-143

#### How to test this
1) Dev docs react page.
2) Also, for making sure nothing was affected by a classname change that I had to make to get past Lint Errors (not warning, but error)... please check the following pages in nitro which is the only spot the circle chart is currently being used. There should be nothing broken about them at all, because I verified that we are not using kits incorrectly, and I looked at their scss to make sure they were not targeting any classnames from the kit. 

https://github.com/powerhome/nitro-web/blob/2f46e50cc69948659b9067f766874ed18d6426ef/components/contact_center/app/views/contact_center/agent_stats/_stats.html.erb

https://github.com/powerhome/nitro-web/blob/2f46e50cc69948659b9067f766874ed18d6426ef/components/contact_center/app/views/contact_center/agent_stats/show.html.erb

#### Checklist:

- [X] **LABELS** Add a label: `enhancement`, `bug`, `improvement`, `new kit`, `depreciation`, or `breaking`. See [Changelog & Labels](https://github.com/powerhome/playbook/wiki/Changelog-&-Labels) for details.
- [X] **URGENCY** Please select a release milestone
- [X] **DEPLOY** Please add the `Milano` label when you are ready for a review.
- [X] **SCREENSHOT** Please add a screen shot or two.
- [X] **SPECS** Please cover your changes with specs.
- [X] **READ DOCS** Please make sure you have read and understand the [Playbook Release Process](https://github.com/powerhome/playbook/wiki/Playbook-Releases)


app/pb_kits/playbook/pb_circle_chart/_circle_chart.jsx
```diff
@@ -0,0 +1,144 @@
...
+      dataLabelHtml: dataLabelHtml,
+      dataLabels: dataLabels,
+      headerFormat: headerFormat,
+      tooltipHtml: tooltipHtml,
+      useHTML: useHtml,
+      minPointSize: minPointSize,
+      maxPointSize: maxPointSize,
+      innerSize: innerSizeFormat(innerSize),
+      zMin: zMin,
+      startAngle: startAngle,
```
thestephenmarshall, line 108:
```suggestion
      id: id,
      colors,
      borderColor: roundedBorderColor,
      borderWidth: roundedBorderWidth,
      chartData: formattedChartData,
      title,
      type: style,
      showInLegend: legend,
      dataLabelHtml,
      dataLabels,
      headerFormat,
      tooltipHtml,
      useHtml,
      minPointSize,
      maxPointSize,
      innerSize: innerSizeFormat(innerSize),
      zMin,
      startAngle,
```

app/pb_kits/playbook/pb_circle_chart/_circle_chart.jsx
```diff
@@ -0,0 +1,144 @@
...
+      componentDidMount.current = true
+    }
+  }, [chartData])
+  return (
+    <div id={`wrapper-circle-chart-${id}`}>
+      <div
+          id={id}
+          {...ariaProps}
+          {...dataProps}
+          className={classnames('pb_circle_chart', globalProps(props), className)}
```
thestephenmarshall, line 135:
```suggestion
          id={id}
          className={classnames('pb_circle_chart', globalProps(props), className)}
          {...ariaProps}
          {...dataProps}
```

thestephenmarshall: Linter should have caught this one.

app/pb_kits/playbook/pb_circle_chart/docs/_circle_chart_colors.jsx
```diff
@@ -0,0 +1,29 @@
...
+    name: 'After call',
+    value: 10,
+  },
+]
+
+const CircleChartColors = () => (
+  <div>
+    <CircleChart
+        chartData={dataWithColors}
+        colors={['data-6', 'data-4', 'data-2']}
```
thestephenmarshall, line 23: I wonder if there is a more direct way to provide these props. It doesn't seem like devs will immediately make the connection to tokens.

app/pb_kits/playbook/pb_circle_chart/_circle_chart.html.erb
```diff
@@ -15,7 +15,7 @@
   <% end %>
 
   <% if object.children %>
-    <div class="pb_circle_chart_block">
+    <div class="pb-circle-chart-block">
```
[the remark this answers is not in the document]
jasperfurniss: Thanks! It was causing problems with my linter... So i had to change it. 
I've verified that it won't cause any problems in Nitro-- there's only one usage of the Circle Chart rails, and I've already reached out to the devs that worked on it. And we will triple check the milano build for that page as well.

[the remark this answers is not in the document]
jasperfurniss: It would only break if there were targeting the kit class itself, which from me searching their scss and reaching out-- it appears we are good.

app/pb_kits/playbook/pb_circle_chart/_circle_chart.jsx
```diff
@@ -0,0 +1,144 @@
...
+    id,
+    innerSize = 'md',
+    legend = false,
+    maxPointSize = null,
+    minPointSize = null,
+    rounded = false,
+    startAngle = null,
+    style = 'pie',
+    title = '',
+    tooltipHtml = '<span style="font-weight: bold; color:{point.color};">●</span>{point.name}: ' +
```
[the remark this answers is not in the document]
jasperfurniss: It's been working for the rails kit. So i think we are safe to use it here. 
RE: did some digging, and it even saves time, and is better for searching. https://stackoverflow.com/a/9809321/4467704

app/pb_kits/playbook/pb_circle_chart/_circle_chart.jsx
```diff
@@ -0,0 +1,144 @@
...
+      useHTML: useHtml,
+      minPointSize: minPointSize,
+      maxPointSize: maxPointSize,
+      innerSize: innerSizeFormat(innerSize),
+      zMin: zMin,
+      startAngle: startAngle,
+    })
+  }, [])
+
+  const componentDidMount = useRef(false)
```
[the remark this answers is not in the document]
jasperfurniss: The ref can be used as instance variable for a function component in React whenever we need to track some kind of state without using React's re-render mechanism. For example, we can track whether a component has been rendered for the first time or whether it has been re-rendered.

It's a nice way to tap into something that useEffect does not provide out of the box-- the ability to track after an initial render.

app/pb_kits/playbook/pb_circle_chart/docs/_circle_chart_colors.jsx
jasperfurniss: Yeah Jason and I threw out a bunch of ideas when we were building the extension of custom colors, and this was the syntax we landed on. I agree though. Could definitely be revisited. Fwiw, I didn't implement that logic on this PR, so i'm guessing it would need to be a separate ticket?

app/pb_kits/playbook/pb_circle_chart/_circle_chart.jsx
jasperfurniss: This one is actually fine because it's inside of the classnames function. The rails kit is already targeting pb_circle_chart  too, so my thought is... The less I have to change, the less likely it will be to break something else. However, it is only being used in one spot on Nitro. I'm happy to update and test thoroughly or leave the same, since it's not causing any breakage and is passing CI.

thestephenmarshall (approved):

PR #1125, opened 2020-09-30
[NUXE-155] Date Picker Patch 3

### LINKED ISSUE #1116: Date Picker Kit Form Label
**Describe the bug**
The date picker kit does not create a label.

**To Reproduce**
1. Create a form kit with a date picker input
1. The label text for the input will display but is not a label

**Expected behavior**
The data picker kit integrates into the the form system like the other inputs and associate a label for the input, and the label should be generated and default based on the name of the input using the generation strategy rails uses.

**Screenshots**
Screen Shot 2020-09-28 at 9 14 11 PM

**Desktop (please complete the following information):**
 - OS: All
 - Browser All
 - Version All

**Smartphone (please complete the following information):**
 - Device: All
 - OS: All
 - Browser All
 - Version All

**Additional context**
None

### LINKED ISSUE #1118: Date picker constraint validaiton
**Describe the bug**
The date picker does not display constraint validation errors.

**To Reproduce**
1. Use the date picker kit in a form with a constraint validation, like being required.
2. Violate the constraint validation
3. Not see the constraint validation violation

**Expected behavior**
When a input has a constraint validation error we should see it reported.

**Screenshots**
Screen Shot 2020-09-28 at 9 47 37 PM

**Desktop (please complete the following information):**
 - OS: Mac OS
 - Browser Firefox Developer Edition
 - Version 82.0b1 (64-bit)

**Smartphone (please complete the following information):**
Not Tested

**Additional context**
I think this is an issue do the the input for the date picker having the readonly attribute set.

bh247484:
#### What does this PR do?

1. Fixes a date picker readonly validation bug
2. Ensures a default empty lambda function for rails onChange prop
3. Allows for Rails generated form labels
4. Fixes form reset bug
5. Reimagines default date null handling. Now null values leave the date picker input blank instead of defaulting to today's date.  Updates docs accordingly.
6. Removes unused doc examples

Closes #1118
Closes #1116

#### Breaking Changes

Should be backwards compatible.  Biggest change is in `defaultValue` prop handling but this won't affect use cases where a prop was passed.  Only cases in which one isn't past.  Even though it shouldn't break functionality it just won't have today's date as the default value.

All other code is edge case specific or extensions of existing functionality.

#### Runway Ticket URL

[Runway Story](https://nitro.powerhrg.com/runway/backlog_items/NUXE-155)

#### How to test this

I already tested an alpha in Nitro but it's worth having the ninjas look at the [Daily Assignments](https://nitro.powerhrg.com/marketing/daily_assignments) page to double check the date picker's functionality there.

#### Checklist:

- [x] **LABELS** Add a label: `enhancement`, `bug`, `improvement`, `new kit`, `depreciation`, or `breaking`. See [Changelog & Labels](https://github.com/powerhome/playbook/wiki/Changelog-&-Labels) for details.


app/pb_kits/playbook/pb_form/form_builder/date_picker_field.rb
```diff
@@ -8,6 +8,7 @@ def date_picker(name, props: {})
           prefix = @object_name
           html_attribute_name = "#{prefix}[#{name}]"
           html_id = "#{prefix}_#{name}"
+          props[:label] = @template.label(@object_name, name) if props[:label] && !props[:label].is_a?(String)
```
terryfinn, line 11: We should treat true as the value that triggers this. That is how other inputs work, and a string value could be a user provided label, and this would take that away the ability to provide a prepared label value.
```suggestion
          props[:label] = @template.label(@object_name, name) if props[:label] == true
```

bh247484: Ah so the `==` operator type checks as well.  That's a much more elegant solution.  Didn't realize Ruby could do that.

My line 11 was checking if the value was 'truthy' and if it wasn't a String.  Which would preserve the 'ability to provide a prepared label value' and still allow a user to pass `true` for a Rails generated label.  But it doesn't account for all other variable types.  For example someone could pass an integer and it would just be registered as a `truthy` value and trigger the Rails generated label.  Thanks for catching this.

jasperfurniss (approved):

PR #1113, opened 2020-09-24
[CMPN-22] Allow all HTML options on input in PbSelectableIcon, PbSelectableCard and PbSelectableCardIcon Rails kits

josephinef9:
#### Runway Link

https://nitro.powerhrg.com/runway/backlog_items/CMPN-22 

#### Screens
Selectable card: 

Screen Shot 2020-09-29 at 8 49 10 AM
Screen Shot 2020-09-24 at 4 27 35 PM

Selectable Icon: 

Screen Shot 2020-09-29 at 8 49 37 AM
Screen Shot 2020-09-24 at 4 26 59 PM

Selectable Card Icon: 

Screen Shot 2020-09-29 at 8 49 22 AM
Screen Shot 2020-09-24 at 4 25 46 PM
 

#### Breaking Changes

[No (Explain)] - Only additive 

#### How to test this
Rspec tests have been written to cover changes. 

#### Checklist:

- [x] **LABELS** Add a label: `enhancement`, `bug`, `improvement`, `new kit`, `depreciation`, or `breaking`. See [Changelog & Labels](https://github.com/powerhome/playbook/wiki/Changelog-&-Labels) for details.
- [x] **URGENCY** Please select a release milestone
- [x] **DEPLOY** Please add the `Milano` label when you are ready for a review.
- [x] **SCREENSHOT** Please add a screen shot or two.
- [x] **SPECS** Please cover your changes with specs.
- [x] **READ DOCS** Please make sure you have read and understand the [Playbook Release Process](https://github.com/powerhome/playbook/wiki/Playbook-Releases)

*The normal release cut off deadline is 3p EDT each week. Please reach out to the release team if you have an urgent request or need a release off cycle.*

PR #1087, opened 2020-09-10
[CMPN-16] Allow all HTML options on input in PbCheckbox Rails kit

stephenmckeon:
#### Screens
Screen Shot 2020-09-10 at 4 18 58 PM
Screen Shot 2020-09-10 at 4 20 22 PM

#### Breaking Changes

No. This PR adds the `input_options` prop to the checkbox kit so PlayBook users can pass every single kind of HTML element option to the input element. This PR does not remove any props as to not break any existing usages of the checkbox kit.

Read [this](https://github.com/powerhome/rfcs/blob/master/0052-playbook-html-attribute-standards.md) RFC for more details.

#### Runway Ticket URL

https://nitro.powerhrg.com/runway/backlog_items/CMPN-16

#### How to test this

[INSERT TESTING DETAILS]

#### Checklist:

- [x] **LABELS** Add a label: `enhancement`, `bug`, `improvement`, `new kit`, `depreciation`, or `breaking`. See [Changelog & Labels](https://github.com/powerhome/playbook/wiki/Changelog-&-Labels) for details.
- [x] **DEPLOY** Please add the `Milano` label when you are ready for a review.
- [x] **SCREENSHOT** Please add a screen shot or two.
- [ ] **SPECS** Please cover your changes with specs.

PR #1130, opened 2020-10-01
[NUXE-81] - Textarea Refactor

coleerikson:
- Refactors the kit to accept `input_options` as an open hash to accept native attributes including:  `disabled`, `required`, `min-length`, and `max-length`.
- Adds `{...props}` to react textarea kit so dark mode works properly in playbook.
- Fixes `error` to be boolean, not a string
- Adds `error_message` props with default state.

#### Screens
textarea-attributes

textarea-attributes-code-rails

textarea-attributes-code-react

#### Breaking Changes

Yes, the textarea kit is used in one location in Nitro:
https://github.com/powerhome/nitro-web/blob/a50a379bd315585ebbdeb0c187f4ce866565c1de/components/nitro_survey/app/views/nitro_survey/blasts/_prompt.html.erb#L45

#### Runway Ticket URL

https://nitro.powerhrg.com/runway/backlog_items/NUXE-81

#### How to test this

Add prop and inspect raw HTML output.

#### Checklist:

- [x] **LABELS** Add a label: `enhancement`, `bug`, `improvement`, `new kit`, `depreciation`, or `breaking`. See [Changelog & Labels](https://github.com/powerhome/playbook/wiki/Changelog-&-Labels) for details.


jasperfurniss (approved):

app/pb_kits/playbook/pb_textarea/_textarea.html.erb
```diff
@@ -15,10 +15,14 @@
         :object,
         :method,
```
garettarrowood, line 16: The first two options in this `text_area` helper method are hard coded will not work. They need to be `object.object` and `object.method`.

garettarrowood (changes requested):
Besides the comment above, and since this PR specifically is about attributes on this kit, let's get the `input_options` convention going before more people in `nitro-web` start using this.

Check out [this RFC about playbook kits with inputs](https://github.com/powerhome/rfcs/blob/master/0052-playbook-html-attribute-standards.md) and [this example](https://github.com/powerhome/playbook/pull/1113) to the SelectableCard kits and a more basic one [done to PbCheckbox](https://github.com/powerhome/playbook/pull/1087).

garettarrowood:
Currently - the Rails kit for textarea is only used once in Nitro Web, in Survey. I have/had a card for Champions to update this to make the options behave like PbCheckbox (not backwards compatibly) to gain more consistently with input kits in Playbook. One usage should be easy to manage with an upgrade.

coleerikson:
@garettarrowood Sounds great! Thanks for taking a look.

coleerikson:
Hey @garettarrowood, I just committed the changes I think you were interested in. Do you mind taking another look?

app/pb_kits/playbook/pb_textarea/textarea.rb
```diff
@@ -8,22 +8,34 @@ class Textarea
       partial "pb_textarea/textarea"
 
       prop :error
```
garettarrowood, line 10: Error prop should be a boolean

app/pb_kits/playbook/pb_textarea/docs/_textarea_error.html.erb
```diff
@@ -1 +1,6 @@
-<%= pb_rails("textarea", props: { error: "This field has an error!", label: "Label", rows: 4}) %>
+<%= pb_rails("textarea",
+    props: {
+        error: "This field has an error!",
```
coleerikson, line 3: @garettarrowood One clarification per our convo. You mentioned error as a prop being a boolean, but here and in the tests it's being passed as a string which is the error message. I fully understand the boolean concept, with that in mind should we start adding a `error_message` prop as well? Or am I missing the pattern?

garettarrowood: This is a perfect example of why this prop should have a type. Yay!

If UX wants to make the error configurable, then an `error_message` prop sounds good to me. I guess we'd have the option to:

1. Continue to use `error` to conditionally render an error.  Then have a default `error_message` that could be specified by the end user passing that additional prop. OR
2. Only have an `error_message` prop that is indeed a `String`, with the value getting used in the template. And infer that if such a prop is provided that indeed there is an error.

Both of these possibilities sound good to me. I don't know if custom error messages are a thing in other playbook kits.

coleerikson: Going with option 1 and a default message of "This field has an error."

coleerikson:
@garettarrowood Updated with error message..

Also, found textarea in Nitro:
https://github.com/powerhome/nitro-web/blob/master/components/nitro_survey/app/views/nitro_survey/audience_filters/_connect_blast_modal.html.erb#L18

garettarrowood:
That should be the only one.

app/pb_kits/playbook/pb_textarea/textarea.rb
```diff
@@ -7,33 +7,45 @@ class Textarea
 
       partial "pb_textarea/textarea"
 
-      prop :error
-      prop :object
+      prop :error, type: Playbook::Props::Boolean,
+                   default: false
+      prop :error_message, type: Playbook::Props::String,
+                           default: "This field has an error."
+      prop :input_id
```
garettarrowood, line 14: All your changes look great!

This is the last bit. I think an `input_id` prop is a bad pattern because it makes sense to pass anything that goes on the input to the `input_options` prop. That is how this is accomplished on PbCheckbox, PbToggle, PbRadio, PbSelectableCard, PbSelectableIcon, and PbSelectableCardIcon (some of those are yet to be merged). I know we have a couple kits with inputs that use this prop, but I'd like to deprecate them. The Champions team would be happy to do this, but we've temporarily halted our Playbook work until our outstanding PRs get some love.

There may have been some followup UX discussion on this convention I've not been a part of. I'd be happy to talk it out with the team, and happy to be convinced otherwise if that becomes the consensus.

coleerikson:
Based on these updates this is now a candidate for a major release. v8.0.0

spec/pb_kits/playbook/kits/textarea_spec.rb
```diff
@@ -7,24 +7,19 @@
...
+  it { is_expected.to define_boolean_prop(:dark).with_default(false) }
+  it { is_expected.to define_hash_prop(:input_options).with_default({}) }
 
   describe "#classname" do
     it "returns namespaced class name", :aggregate_failures do
       expect(subject.new({}).classname).to eq "pb_textarea_kit resize_none"
-      expect(subject.new({dark: true}).classname).to eq "pb_textarea_kit dark resize_none"
-      expect(subject.new({error: "Something is wrong"}).classname).to eq "pb_textarea_kit error resize_none"
-      expect(subject.new({dark: true, error: "Something is wrong"}).classname).to eq "pb_textarea_kit dark error resize_none"
+      expect(subject.new({dark: true}).classname).to eq "pb_textarea_kit resize_none dark"
```
garettarrowood, line 20: Nit: When only key/value pairs are passed in Ruby, it implicitly makes them a hash. So this is the same thing, and slightly slicker (not a blocker).

```suggestion
      expect(subject.new(dark: true).classname).to eq "pb_textarea_kit resize_none dark"
```

app/pb_kits/playbook/pb_textarea/_textarea.html.erb
```diff
@@ -1,26 +1,17 @@
 <%= content_tag(:div,
-    id: object.id,
+    aria: object.aria,
+    class: object.classname,
     data: object.data,
-    class: object.classname) do %>
+    id: object.id) do %>
   <% if object.label.present? %>
-    <%= pb_rails("caption", props: {text: object.label, dark: object.dark}) %>
+    <%= pb_rails("caption", props: {text: object.label}) %>
```
jasoncypret, line 7: I believe dark needs to be passed to the children. Otherwise setting dark would not set dark on the caption

app/pb_kits/playbook/pb_textarea/textarea.rb
jasoncypret: Makes sense

jasoncypret (changes requested):

jasoncypret:
@coleerikson Can you write a better title? Thinking about the changelog and this being a breaking change

jasoncypret:
Lets remove the error additions. That is being taken care of in a separate PR

garettarrowood (approved):
Approving to remove my change request. Happy to see this work moving forward!

thestephenmarshall:
This one is pretty old. What do we think we should do with it? Still worth updating and pushing @powerhome/ux-team ?

kre8sions:
Closing this PR but keeping the branch alive for the future @jasoncypret

PR #1170, opened 2020-10-23
Fix production links to logs and metrics in the Milano UI

indiebrain:
Playbook production has moved to the new app-prod-hq cluster. As such,
its telemetry is now being reported to the Kibana and Grafana
instances in the new cluster. Here we update the links so consumers
can still use the Milano UI to get to the relevant information for
playbook-production.


bh247484 (approved):

indiebrain:
A tip-o-the-hat to you, sir!

PR #1183, opened 2020-11-01
Don't issue new certs for review envs

benlangfeld:
Each APP cluster has a wildcard cert issued for playbook-prs which is provisioned into each review namespace. This avoids having to issue a new cert for each review environment, which results in regular notifications of cert expiry for these discarded certs.


benlangfeld:
@powerhome/ux-team You've got some tests which depend on the time of year there.

thestephenmarshall (approved):

PR #1269, opened 2020-12-16
[NUXE-48] - changelog page

andrew-k9:
#### Screens

Screen Shot 2020-12-16 at 2 35 14 PM

#### Runway Ticket URL

[NUXE-48](https://nitro.powerhrg.com/runway/backlog_items/NUXE-48)

#### How to test this

Make sure the page loads, the info matches the `CHANGELOG.md` file, and the links work!


jasoncypret (changes requested):
Design updates. I did this in the browser. 

image

andrew-k9:
@jasoncypret just updated and [checked in demo environment](https://pr1269.playbook.beta.gm.powerapp.cloud/changelog), let me know if this looks good!

playbook/app/controllers/playbook/pages_controller.rb
```diff
@@ -31,6 +32,15 @@ def enable_dark_mode
...
     def getting_started; end
 
+    def changelog
+      markdown = Redcarpet::Markdown.new(Redcarpet::Render::HTML, extensions = {})
+      @data = File
+              .read("../../CHANGELOG.md")
+              .to_s
+              .split(/##\s\[/)
+              .map { |change| markdown_chunk(markdown, change) }[1..-1]
+    end
```
jasoncypret, line 42: @bh247484 Can you weigh in here. The changelog is already in markdown. It seems strange to change this toa string then convert back to markdown.

I'm wondering if we should just display the markdown in as is in a single card.  markdown(../../CHANGELOG.md)`

Example in view: https://github.com/powerhome/playbook/blob/4f64274c26be3581583ef59840496c2e3322eba2/playbook/app/views/playbook/pages/kit_show.html.erb#L4

jasoncypret (changes requested):
I'm worried about the string conversion to back to markdown. I'd like to either just display the markdown as is in 1 single card...or perhaps do this split in the markdown helper.

https://github.com/powerhome/playbook/blob/912d8c067ffa681907ca5bfd6cf5ff75b71d0378/playbook/app/helpers/playbook/redcarpet_helper.rb

playbook/app/controllers/playbook/pages_controller.rb
bh247484: Agreed.  I like the idea of splitting `CHANGELOG.md` into each constituent release but until we find other ways to get value out of that (beyond splitting them into separate cards) I think just outputting the `.md` into a single card makes more sense.

Again, cool idea though.  Would like to revisit it in the future.

andrew-k9: @jasoncypret

1) The conversion is a bit odd, but I'm unsure of a more safe/ straightforward way todo this. It probably is just a me being new dev issue and can be done more thoughtfully! 😅 I would love feedback on that portion in particular - it seems to also slow down the page load a tad since there is regex applied to a rather large file, though in the demo environment I didn't notice the slow-down...

2)  I'll play around with having it be displayed in one card and just displaying the markdown as-is and post it here!

andrew-k9: @jasoncypret @bh247484 added the screenshot for the single card view in the design feedback room [here](https://nitro.powerhrg.com/connect/api/v4/files/1615521/body). Also just merged the code here. I tried to just do `markdown("../../CHANGELOG.md") but it's expecting a file so I still am doing a `File.read` call

Lmk what ya'll think!

playbook/app/views/playbook/pages/changelog.html.erb
```diff
@@ -0,0 +1,19 @@
+<div class="changelog-card-collection">
+  <%= pb_rails("flex", props: { orientation: "column", horizontal: "stretch" }) do %>
+    <%= pb_rails("flex/flex_item", props: { margin_top: "md", margin_bottom: "sm", margin_: "md" }) do %>
+      <%= pb_rails("title", props: { text: "Changelog", size: 1, margin_x: "sm" })%>
+    <% end %>
+    <%= pb_rails("flex/flex_item", props: { margin_left: "sm", margin_bottom: "xs" }) do %>
+      <%= pb_rails("body", props: { text: "We release a new version every Thursday!" }) %>
```
jasoncypret, line 7:
```suggestion
      <%= pb_rails("body", props: { text: "Playbook releases operates on a weekly release schedule. We typically create a new release every Thursday." }) %>
```

jasoncypret (approved):

PR #1272, opened 2020-12-17
Introduce view component

xjunior:
#### Screens

As explained on [RFC-0056](https://github.com/powerhome/rfcs/blob/master/0056-playbook-project-organization.md#playbook-rails), playbook-rails implements a supporting library to allow us to represent a portion of the page as a component, similar to React. The initial plan was to, in the future, extract this supporting library to a separate gem so other apps could take advantage of this abstraction, without having to also include the Playbook itself. Such a supporting library would also provide the same abstraction to apps like Nitro and the Playbook Website itself.  [ViewComponent](https://viewcomponent.org/) is precisely that.

This PR introduces ViewComponent to Playbook, and also replaces our current rendering process with ViewComponent in almost all components, except for Forms that might require have more involved testing.

#### Why?

Replacing our own rendering of a component by ViewComponent has the benefit of removing the small partials we have to represent each part of the screen by a cached/compiled mechanism, removing the template step process from the rendering. This also makes the lookup of templates faster for other views that are not componentized (such as the page view) by removing a large number of partials from the lookup.


garettarrowood:
This looks fantastic!

xjunior:
Benchmark on nitro's `/dev_docs/playbook`

| view_component | DISABLE_PLAYBOOK_PERFORMANCE_FIX | time |
|------------------|--------------------------------------------|------|
| true | true | 148248.07 |
| true | false | 72667.76 |
| false | true | 897164.60 |
| false | false | 94722.25 |

playbook/app/pb_kits/playbook/pb_avatar_action_button/avatar_action_button.rb
```diff
@@ -2,13 +2,11 @@
...
   module PbAvatarActionButton
-    class AvatarActionButton
-      include Playbook::Props
-
+    class AvatarActionButton < Playbook::KitBase
       prop :action, type: Playbook::Props::String,
                     default: "add"
       prop :image_url, type: Playbook::Props::String
-      prop :link_url, type: Playbook::Props::String
+      prop :link_url, type: Playbook::Props::String, default: "#"
```
thestephenmarshall, line 9:
```suggestion
      prop :link_url, type: Playbook::Props::String
```

This change seems unrelated to the PR.

xjunior: @thestephenmarshall the link being `nil` makes the `link_to` in the view generate a URL for the current page, making it always link somewhere regardless of the explicit desire of the kit user. The problem was exposed in this change because the view renderer is trying to generate a URL using the Rails application routes, while the route where this kit is rendered is covered by the Engine, causing an error.

thestephenmarshall: Ah ok then if we were receiving an error pls ignore my comment.

thestephenmarshall (approved):
@xjunior I want to make this PR into it's own release so that we can better manage any unforseen issues if that makes sense.

garettarrowood (approved):
+1 to @thestephenmarshall

xjunior:
It does. And if that's the case, there's more work that I also wanted to include on the same release, then:

1) The conversion of the form kit (which I left to a separate PR to ease the review)
2) The complete removal of any view path configuration, so playbook paths are not included in any view file lookup
   * The removal of some partials used by kits (either by making them a sub kit or just inlining its contents)

They should be two small PR's coming after this. This can go to master without a release, though.

PR #1283, opened 2020-12-23
[NUXE-90] generator kit alphabetize content

andrew-k9:
[NUXE-90](https://nitro.powerhrg.com/runway/backlog_items/NUXE-90) adds automatic sorting when generating a new kit.

__Testing:__ fetch this branch and try it out, I've already done the following with kit names:

- Top of the list
- Bottom of the list
- Middle of list


playbook/app/pb_kits/playbook/index.js
```diff
@@ -2,7 +2,7 @@ import './_playbook.scss'
 import 'lazysizes/plugins/attrchange/ls.attrchange'
 import 'lazysizes'
 
-// React Component JSX Imports from the React Kits
+// vvv React Component JSX Imports from the React Kits vvv
```
bh247484, line 5: Might be worth leaving another comment here saying `// The comment below is used by the kit generator to alphabetize imports, moving or removing it will break functionality` or something to that effect.  

Also might be good to name it more explicitly, something like `start-flag` and `end-flag`.  I like the concept though, clever solution.

andrew-k9: I was just airing on the side of caution by keeping the same wording but adding arrows. During dailies today we can see what the rest of the team would think is a good idea. Something like an extra line above and below the comments like DO NOT REMOVE BELOW/ ABOVE?

As for adding a method for the kits that seems easy enough 👷🏽‍♀️

thestephenmarshall: Generally speaking, it's probably a good idea to just add something to the effect of 

> This file is autogenerated. Please do not edit.

playbook/lib/generators/kit/kit_generator.rb
```diff
@@ -74,13 +74,35 @@ def create_templates
...
         template "kit_js.erb", "#{full_kit_directory}/docs/index.js"
 
-        # Import kit examples  ===========================
-        append_to_file("app/pb_kits/playbook/packs/react-examples.js") do
-          "import * as #{@kit_name_pascal} from 'pb_#{@kit_name_underscore}/docs'\nWebpackerReact.setup(#{@kit_name_pascal})\n"
-        end
-        append_to_file("app/pb_kits/playbook/index.js") do
-          "\nexport #{@kit_name_pascal} from './pb_#{@kit_name_underscore}/_#{@kit_name_underscore}.jsx'"
-        end
+        re_array = File.readlines("app/pb_kits/playbook/packs/react-examples.js")
```
thestephenmarshall, line 77: We should be interpolating the path here instead of hard-coding it.

playbook/lib/generators/kit/kit_generator.rb
```diff
@@ -74,13 +74,35 @@ def create_templates
...
+        webpack_components.sort!
+
+        sorted_file_array = re_array[0..(re_array.index("// KIT EXAMPLES\n") + 1)]
+        sorted_file_array += example_components
+        sorted_file_array << "\n"
+        sorted_file_array << "WebpackerReact.setup({\n"
+        sorted_file_array += webpack_components
+        sorted_file_array << "})\n"
+
+        File.open("app/pb_kits/playbook/packs/react-examples.js", "w+") { |f| f.write(sorted_file_array.join) }
```
thestephenmarshall, line 94: Interpolate the path

playbook/lib/generators/kit/kit_generator.rb
```diff
@@ -74,13 +74,35 @@ def create_templates
...
+
+        file_array = File.readlines("app/pb_kits/playbook/index.js")
```
thestephenmarshall, line 96: Interpolate

playbook/lib/generators/kit/kit_generator.rb
```diff
@@ -74,13 +74,35 @@ def create_templates
...
+        start = file_array.index("// vvv React Component JSX Imports from the React Kits vvv\n")
+        finish = file_array.index("// ^^^ React Component JSX Imports from the React Kits ^^^\n")
+        components = file_array[(start + 1)..(finish - 1)]
+
+        components << "export #{@kit_name_pascal} from './pb_#{@kit_name_underscore}/_#{@kit_name_underscore}.jsx'\n"
+        components.sort!
+        file_array = file_array[0..start] + components + file_array[finish..-1]
+
+        File.open("app/pb_kits/playbook/index.js", "w+") { |f| f.write(file_array.join) }
```
thestephenmarshall, line 105: Interpolate

thestephenmarshall (changes requested):

thestephenmarshall (approved):

PR #1308, opened 2021-01-11
[NUXE-98]  Selectable card border bug (using border)

patmccler:
Fixes border issue.
Cause: Padding props overwrite the css which accounts for changing border size. Fixed by adding a 1 px buffer div around border that shrinks to 0 when border grows.

Other fixes:
Fixes an issue in the react implementation where padding and margin props were being applied twice.
Fixes an issue where margin and padding props both applied to outer container div, making padding not work as expected (inside border)
 - I only see one place in Nitro using padding prop, the "Window Card" in home tour. This PR fixes the jumping likely seen there when selecting and deselecting a card

#### Screens
No visual changes.
Screen Shot 2021-01-11 at 5 54 10 PM

#### Breaking Changes
None
#### Runway Ticket URL
[NUXE-98](https://nitro.powerhrg.com/runway/backlog_items/NUXE-98)
#### How to test this
/kits/selectable_card/rails
/kits/selectable_card/react

May need to pull down locally and add padding props to the exampes,  as the bug was not visible until padding props are used

#### Checklist:


playbook/app/pb_kits/playbook/pb_selectable_card/docs/_selectable_card_single_select.jsx
```diff
@@ -23,7 +23,7 @@ class SelectableCardSingleSelect extends React.Component {
             name="gender"
             onChange={this.handleSelect.bind(this)}
             value="male"
-            {...props}
+            {...this.props}
```
coleerikson, line 26: What's going on here with the swap from `{...props}` to `{...this.props}`?

playbook/app/pb_kits/playbook/pb_selectable_card/_selectable_card.jsx
```diff
@@ -55,7 +57,9 @@ const SelectableCard = ({
     { 'checked': checked,
       'disabled': disabled,
       'enabled': !disabled }),
-  globalProps(props), className)
```
coleerikson, line 58: Is dropping `globalProps` cool here?

playbook/app/pb_kits/playbook/pb_selectable_card/_selectable_card.jsx
```diff
@@ -55,7 +57,9 @@ const SelectableCard = ({
...
+  dark && 'dark' || '',
```
coleerikson, line 60: Is this line essentially `dark ? 'dark' : ""`..if so, let's use this one since it's a more common pattern.

playbook/app/pb_kits/playbook/pb_selectable_card/docs/_selectable_card_single_select.jsx
coleerikson: Is this the "double padding/margin" fix?

playbook/app/pb_kits/playbook/pb_selectable_card/_selectable_card.jsx
bh247484: I see it's invoked below on line 98 but I think it should be here as well, otherwise global props won't make it into the `classes` variable which gets passed to the outer div on line 84.

playbook/app/pb_kits/playbook/pb_selectable_card/docs/_selectable_card_single_select.jsx
bh247484: This is an old school class based React component (instead of new functional comp) so I think you've gotta have `this.props` as opposed to just `props`.  Recent PR went through and added prop passes to all the React examples.  They likely misapplied props here without the `this.`.

patmccler: props aren't passed to render function on a class component, so they weren't doing anything as `...props`
Pairing with Jasper on this now

playbook/app/pb_kits/playbook/pb_selectable_card/_selectable_card.jsx
patmccler: Putting global props in two places is what was causing the padding and margin props to apply twice, both to the outer div and to the label that actually displays.  This means that you get twice as much space as you intended.

Global props looks like it includes dark props and spacing props. I accounted for dark manually since I didn't want to put spacing on the outer element.

 Open to better ways to handle, let me know what you guys think

playbook/app/pb_kits/playbook/pb_selectable_card/_selectable_card.jsx
patmccler: Exactly yes, Ill change it over

playbook/app/pb_kits/playbook/pb_selectable_card/docs/_selectable_card_single_select.jsx
coleerikson: Ok, my mistake, I didn't even notice the class structure. So many of our components are functional it just caught my eye. `this.props` is totally cool if we stick with that class component, but if it's possible to use hooks that would be cool too.

playbook/app/pb_kits/playbook/pb_selectable_card/_selectable_card.jsx
bh247484: Would it create problems to leave `globalProps` off the inner `label` element and leave it only on the outermost div?

patmccler: @bh247484 The reason for putting global props on the inner `label` is that is the element you actually see when the kit is rendered. If you put global props only on the outermost `div`, then any padding applies outside the border and not inside, as you'd likely expect when using the kit

playbook/app/pb_kits/playbook/pb_selectable_card/docs/_selectable_card_single_select.jsx
patmccler: Changed this over to functional, matching the rest of the react examples

jasperfurniss (approved):

PR #1314, opened 2021-01-14
Clear partials use in kits

xjunior:


playbook/app/controllers/playbook/application_controller.rb
```diff
@@ -6,7 +6,6 @@ class ApplicationController < ActionController::Base
     helper Playbook::ApplicationHelper
     helper Playbook::PbDocHelper
     helper Playbook::PbSampleHelper
-    append_view_path Playbook::Engine.root + "app/pb_kits"
```
jasoncypret, line 9: Does this change our Getting started docs? https://playbook.powerapp.cloud/getting_started

jasoncypret (changes requested):
None of the kits views load in Milano. I'll test locally next

image

playbook/app/controllers/playbook/application_controller.rb
xjunior: It does, I'll update it

xjunior: Actually, no. Our `getting_started` has no mention to the view paths. But it should not be necessary because even today, `Playbook::Engine` configures the view paths already.

jasoncypret (approved):

PR #1401, opened 2021-03-22
Revert "Revert "Allow url to be empty so form_with can generate a url based on the model""

jasperfurniss:
Reverts powerhome/playbook#1400


jasperfurniss:
@xjunior

PR #1411, opened 2021-03-29
More flexible dependency versions

benlangfeld:
Allows consuming applications to upgrade all dependencies within the specified major version, so as not to block feature and bug fix upgrades on playbook releases.

Since this does not permit major version upgrades, and thus backward incompatible upgrades, it should not be problematic.


playbook/package.json
```diff
@@ -73,21 +73,21 @@
...
     "flatpickr": "^4.6.6",
-    "flow-bin": "0.137.0",
-    "flow-runtime": "0.14.0",
-    "highcharts": "7.2.2",
+    "flow-bin": "^0.137.0",
+    "flow-runtime": "^0.14.0",
+    "highcharts": "^7.2.2",
     "jest-axe": "^4.1.0",
     "jsx-control-statements": "^3.2.8",
     "lazysizes": "^5.2.2",
```
thestephenmarshall, line 93: Realistically, most of these should be `devDependencies` which means you shouldn't care at all about the version. The only thing we ship non-transpiled is our SCSS tokens which are pretty standard and shouldn't really care about the version of `node-sass` either.

playbook/package.json
```diff
@@ -98,24 +98,24 @@
...
     "webpack-cli": "^3.3.12",
     "webpacker-react": "^0.3.2"
   },
   "peerDependencies": {
-    "react": "16.8.6",
-    "react-dom": "16.8.6",
-    "trix": "1.3.0",
+    "react": "^16.8.6",
+    "react-dom": "^16.8.6",
+    "trix": "^1.3.0",
```
thestephenmarshall, line 118: We should be careful here as what happens here is we tell the transpiler *not* to include this version, but instead require that the consumer of the library provide the version specified in their application. We are fairly confident that much later versions of React would still be backwards compatible with the transpiled JavaScript, but this is definitely a place that needs much more testing. `react` and `react-dom` as you can imagine are the main libs of concern here.

thestephenmarshall (changes requested):
This needs alpha testing before we can move any further with this PR. We cannot ship this before that has been done. Also, I'll defer to the opinions of the rest of @powerhome/ux-team  with regards to the severity of the changes here and whether or not this indicates the need for a major release or not. It is my opinion that this could possibly be shipped as a regular release. Again, this depends on alpha testing against this PR alone first.

playbook/package.json
benlangfeld: I'm all for moving things to devDependencies. If you can list which ones can move and which have to stay, I'm totally up for that.

benlangfeld:
If this whole change would be of concern, but there similar relaxation of dependencies (including the above suggestion to make many of these `devDependencies`) that the UX team can provide, @powerhome/eternals would be very grateful; they have a list of open PRs which are marked as blocked on playbook for this reason.

playbook/package.json
thestephenmarshall: I would say everything here in `package.json` can be moved to devDependencies as the only ones we are truly concerned about are in peerDependencies currently. These being in the wrong place has been a matter of contributors not using the `-D` flag when installing a library and/or not understanding what goes where and why.

thestephenmarshall:
I can add a commit here to move things to where we see fit. Is that okay @benlangfeld ?

benlangfeld:
@thestephenmarshall All yours :)

benlangfeld:
Hey @thestephenmarshall , it seems we have a Nitro PR open to test this. Are we going this way, or moving those deps to `devDependencies`?

thestephenmarshall:
@benlangfeld that is correct. I have moved most everything to devDependencies in my last commit and fixed other items that make our dev docs in Nitro work for the React side as well.

thestephenmarshall (approved):

benlangfeld:
Thanks @thestephenmarshall , I'm looking forward to this.

thestephenmarshall:
@benlangfeld the only dependency that I had to end up specifying was `webpack` because [it is a webpack-dev-server peerDependency](https://github.com/webpack/webpack-dev-server/blob/master/package.json#L116)

benlangfeld:
Thanks for getting this tested and merged. Please let me know when it makes it into a release so we can upgrade Tempo.

thestephenmarshall:
@benlangfeld it has been yanked

thestephenmarshall:
we need to see why CI is being a PITA

thestephenmarshall:
[New PR](https://github.com/powerhome/playbook/pull/1440)

PR #1435, opened 2021-04-05
Updated user css file

christinaatai:
#### Screens
*Before*
Screen Shot 2021-04-05 at 12 11 54 PM

*After*
Screen Shot 2021-04-05 at 12 22 15 PM

*Old NavItem Block  doc example*
Screen Shot 2021-04-05 at 2 04 19 PM

*New NavItem Block doc example*
Screen Shot 2021-04-05 at 2 03 56 PM

#### Breaking Changes

No –– this PR only affects the CSS class in User Kit.

#### Runway Ticket URL

https://nitro.powerhrg.com/runway/backlog_items/NUXE-584

#### How to test this

I tested this bug by updating the doc example so that a User Kit is in Nav/NavItem to confirm that the styling is behaving as expected.

#### Checklist:

### DIFF
--- a/playbook/app/pb_kits/playbook/pb_user/_user.scss
+++ b/playbook/app/pb_kits/playbook/pb_user/_user.scss
@@ -38,9 +38,6 @@
 
   &[class*=_horizontal]  {
     align-items: center;
-    .content_wrapper > * {
-      padding: 0 0 0 0;
-    }
 
     &[class*=_center]  {
       justify-content: center;
--- a/playbook/app/pb_kits/playbook/pb_user/_user.scss
+++ b/playbook/app/pb_kits/playbook/pb_user/_user.scss
@@ -38,6 +38,9 @@
 
   &[class*=_horizontal]  {
     align-items: center;
+    .content_wrapper > * {
+      padding: 0 0 0 0;
+    }
 
     &[class*=_center]  {
       justify-content: center;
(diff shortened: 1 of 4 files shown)


playbook/app/pb_kits/playbook/pb_user/_user.scss
```diff
@@ -38,6 +38,9 @@
 
   &[class*=_horizontal]  {
     align-items: center;
+    .content_wrapper > * {
+      padding: 0 0 0 0;
+    }
```
bh247484, line 43: Initially felt like the fix should be making the List kit more specific/limited in how it applies padding to children (currently it's applying it to `[class*=_title]` to any child elements, see screenshot below) but now not so sure.  Seems safe enough to apply the above padding rules since the User kit will likely never want padding on those elements but again feels like the long term fix is actually in the List kit.

Curious to hear what others think.

Here's the styles as applied in the browser for reference:
list_padding

I guess scoping the styles in the List kit will be difficult since there's no telling how many/what children will be passed.  Makes me think what Christina's got here makes the most since near term.

christinaatai: @bh247484 Good call! It totally agree with you. The weird padding that affected the name in the User kit was supposed to only affect the title of the Nav list title, so I updated the affected class name in the CSS. That way, CSS in User kit does not need be be touched.

bh247484 (approved):
Looks good to me!

jasoncypret (approved):

jasoncypret:
@powerhome/ux-team This can go right into a master. No need for an alpha

PR #1452, opened 2021-04-14
[NUXE-605] Decouple website from library

xjunior:
**Runway Ticket:** [NUXE-605](https://nitro.powerhrg.com/runway/backlog_items/NUXE-605)
**Nitro PR:** [nitro#20491](https://github.com/powerhome/nitro-web/pull/20491)

### Breaking Changes

The breaking changes focused on the development cycle. While both projects lived inside the same space in the previous arrangement, everything would naturally happen there. This PR introduces changes proposed by [RFC-0056](https://github.com/powerhome/rfcs/blob/main/0056-playbook-project-organization.md) and splits the library code and the [playbook website](https://playbook.powerapp.cloud/) app code into two separate projects.

Now commands to *release a new library version* (`rake pb_release:version`), *generate a new library kit* (`rails g kit`), and *update the library changelog* (`rake changelog`) should, naturally, run from the library directory (`playbook/`). And commands such as *generate a new sample*, which aims to generate content to the website, should run from the website directory.

#### Docker Environment

Another breaking change is the removal of docker from the development environment. This is because docker added more complexity to the mix while providing no value in the context. `playbook-website` has no service dependencies, and it's a lot easier/faster/simpler to use plain rails development environment.

The library code is also a lot simpler to manage/develop/test in a local environment. Docker-compose provided no value in this scenario and was replaced by foreman.

Now, to go from 0 to develop, you have to:

```bash
git clone git@github.com:powerhome/playbook.git
cd playbook
asdf install
yarn install
(cd playbook-website && bundle)
yarn start-dev
```

### How to test this

Playbook website should _just work ™_. A PR for nitro will also be created as soon as possible and linked here.

### TODO

- [x] Changelog generator
- [x] Husky / Overcommit

#### Javascript
- [x] Warning on mounting WebpackerReact components
- [x] Development environment

#### Ruby issues
- [x] pb_release
- [x] kits generators
- [x] sample generator
- [x] Figure out title_mixin


playbook-website/config/initializers/global_variables.rb
```diff
@@ -0,0 +1,2 @@
+MENU = YAML.load_file(Playbook::Engine.root.join("app/pb_kits/playbook/data/menu.yml"))
+SAMPLES = YAML.load_file(Playbook::Engine.root.join("app/pb_kits/playbook/data/samples.yml"))
```
xjunior, line 2: One day…

playbook-website/app/javascript/site_styles/docs/_markdown.scss
```diff
@@ -1,25 +1,25 @@
-@import "../../../pb_title/title_mixin";
-@import "../../../tokens/spacing";
-@import "../../../tokens/colors";
-
+// @import "playbook-ui/dist/pb_title/title_mixin";
```
drborges, line 1: Are these commented-out lines temporary?

benlangfeld:
I dropped the Milano tag because the deployment is not functional and is aggressively paying Guardians.

playbook-website/app/javascript/site_styles/docs/_markdown.scss
xjunior: Yeah, until I figure this out. Cause this `title_mixin` isn't actually exposed in playbook's public API, so I'm thinking how to tackle this issue, I can either expose with `tokens` or inline/duplicate these mixins in this file.

xjunior:
@benlangfeld I'm trying to make it functional, but I'll try it from another angle first, then.

playbook-website/app/javascript/packs/main.scss
```diff
@@ -0,0 +1,30 @@
+// SCAFFOLD
+@import "playbook-ui/dist/reset";
```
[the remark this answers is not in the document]
xjunior: I have to find a way to remove `main.js` and continue to produce the final `main.css`

indiebrain:
This PR's review stack is failing to deploy and causing a torrent of PagerDuty alerts for the Guardians team. It seems like the platform is receiving a `404` HTTP response when pinging the Readiness Probe / Liveness Probe endpoint of the container deployed there. Has this change moved [the readiness probe path](https://github.com/powerhome/playbook/blob/66b69be7a89f46ac2e97501e66aeeba877cad77c/playbook/config/deploy/templates/deployment.yaml.erb#L45-L56) of the playbook deployment pods?

I'm removing the Milano tag to reduce the amount of PagerDuty noise.

xjunior:
@indiebrain thanks for clarification!

README.md
```diff
@@ -3,32 +3,29 @@
...
-1. run `bin/rails s`
-1. run `./bin/webpack-dev-server`
-3. open [http://localhost:3000](http://localhost:3000)
+### Getting started
 
-<details><summary>Making changes to the Gemfile:</summary>
-<p>
+1. Install commit hooks `yarn prepare`
+1. Install all projects dependencies: `yarn install`
+1. Install website ruby dependencies: `cd playbook-website && bundle`
```
jasoncypret, line 19: The location these commands are run is a bit strange, as they jump around a bit. For example, here I need to be inside the website folder. But to run the website I would need to be in root. Not sure if anything can be done about that in this version. Perhaps we just need to make the readme a bit more clear about location as I ran `yarn start-dev` from the wrong folder.

jasoncypret (changes requested):
The sample pages (REACT ONLY) are not displaying

jasoncypret (changes requested):
The`/getting_started` page is not displaying markdown. Also the `/changelog` doesn't load

jasoncypret:
@xjunior Can you confirm a couple of commands for me?

1. Generate kit: `bin/rails generate kit "Buttonzzz"` in `/playbook`
2. Release: `bin/rails pb_release:version` in `/playbook`
3. Generate Sample: `bin/rails sample "Buttonzz" ` in `/playbook-website`

I'm not getting any of these to work.

xjunior:
@jasoncypret the third command should be `bin/rails g sample …`, but everything else is right. I fixed the generators in my last commit.

jasoncypret:
@xjunior I was able to run all the commands now. The sample page generated a broken jsx path...but that is likely related to the other issues so I'll just report for now

README.md
thestephenmarshall:
```
➜  playbook-website git:(decouple-website) bundle
Ignoring nokogiri-1.10.9 because its extensions are not built. Try: gem pristine nokogiri --version 1.10.9
Ignoring redcarpet-3.4.0 because its extensions are not built. Try: gem pristine redcarpet --version 3.4.0
Warning: the running version of Bundler (1.17.2) is older than the version that created the lockfile (1.17.3). We suggest you upgrade to the latest version of Bundler by running `gem install bundler`.
Your Ruby version is 2.6.3, but your Gemfile specified 2.5.0
```

I ran `bundle` as described above and this is what I received. 

```
➜  playbook-website git:(decouple-website) ruby -v
ruby 2.5.0p0 (2017-12-25 revision 61468) [x86_64-darwin20]
```

thestephenmarshall: Fixed by running `gem install bundler -v 1.17.3` so that might be good to add to the docs @xjunior

thestephenmarshall, line 19:
```suggestion
1. Install website ruby dependencies: `cd playbook-website && bundle`
  1. If you receive a bundle(r) related error, be sure that `bundler` is installed first. See the `BUNDLED WITH` section of `playbook-website/Gemfile.lock` for the exact version to install.
```

thestephenmarshall (changes requested):
I checked out the release creation process and just want to confirm:

- Distribution files are created when the website is loaded, which means every PR will contain distribution files unless we discontinue checking those files in. Thoughts @jasoncypret 
- The process seems to incorrectly name two files:
  1. `reset.css.css`
  2. `reset.css.js`
  
  I don't think this is correct ^. There should only be one `reset.css` ever. 

Screen Shot 2021-05-17 at 4 45 43 PM

xjunior:
@thestephenmarshall we must update the dist files to be able to compile the website against it. Many PR's from now we'll be able to get rid of this process, but the directory/files structure in `src` must mirror dist, so we can point either way, and our packs actually contain library code baked in. For now, we should ignore those files that are a result of a simpler webpack.config.js that I couldn't get rid of. I didn't want to spend too much time tunning webpack.config.js because my goal is to move to rollup, or at least evaluate other options.

xjunior:
Also, I don't see a problem in the PR's updating the dist. They'll certainly conflict if people start to push them to their PR's, but they always have the option to not do that. If we really don't want that, we can [ignore those files](https://stackoverflow.com/questions/4348590/how-can-i-make-git-ignore-future-revisions-to-a-file) and only update them during a release.

xjunior:
> The`/getting_started` page is not displaying markdown. Also the `/changelog` doesn't load

Fixed these

jasoncypret (approved):

PR #1462, opened 2021-04-19
[CHMN-60] HaveIBeenPwned passphrase

patmccler:
#### Screens
No new visuals.
#### Breaking Changes
No, additive changes to the passphrase kit, with some refactoring. Other kits not affected. Passphrase kit not used yet (but will be soon)

Passphrase kit
  - add a new prop to enable use of https://haveibeenpwned.com/Passwords . This is a free api that you pass the first 5 characters of a potential passphrase sha-1 hash to, and it responds with ~500 sha-1 hashes that have the same first 5. If your hash is in the response, the passphrase you want to use is breached and shouldn't be used.

 - Add strength to the data attribute of the kit so a rails page can access that information and act on it. Currently, its just a callback in react so its hard to work with in rails.
 
- refactor use of "zxcvbn" into a hook

#### Runway Ticket URL

[CHMN-60](https://nitro.powerhrg.com/runway/backlog_items/CHMN-60)
#### How to test this

/kits/passphrase/react
/kits/passphrase/

See that the data attribute changes as you type
See that the API is hit when typing in the last example on each page.
#### Checklist:


playbook/app/pb_kits/playbook/pb_passphrase/useHaveIBeenPwned.js
```diff
@@ -0,0 +1,51 @@
...
+    () => {
+      // only check the API for passphrases above the minimum size
+      if (passphrase.length < minLength) {
+        setIsPwned(false)
+        return
+      }
+
+      const handler = setTimeout(() => {
+        checkHaveIBeenPwned(passphrase)
+          .then((pwned) => setIsPwned(pwned))
```
thestephenmarshall, line 40: Since this hits an external web API, I think it makes sense to have some failover here incase the service is down or returning something other than 2xx code with a `.catch` and `.finally` which would set `pwned` to `false`. We will not know for sure whether to set the value to `true` in this case, so best to default this way.

playbook/app/pb_kits/playbook/pb_passphrase/useZxcvbn.js
```diff
@@ -0,0 +1,54 @@
...
+
+  const [percent, setPercent] = useState('0')
+  const [variant, setVariant] = useState('negative')
+  const [text, setText] = useState('\u00A0') //nbsp to keep height constant
+  const [result, setResult] = useState({})
+
+  useEffect(() => {
+    if (confirmation) {
+      return
+    }
```
thestephenmarshall, line 19:
```suggestion
    if (confirmation) return
```

playbook/app/pb_kits/playbook/pb_passphrase/useZxcvbn.js
```diff
@@ -0,0 +1,54 @@
...
+      setText('Too weak')
+    } else if (str < strongThreshold){
+      setPercent('50')
+      setVariant('warning')
+      setText('Almost there, keep going!')
+    } else if (str >= strongThreshold) {
+      setPercent('100')
+      setVariant('positive')
+      setText('Success! Strong passphrase')
+    }
```
thestephenmarshall, line 44: Not critical, but this area seems a little difficult to parse the logic. You could simplify it by taking the contents of the conditional statements and converting to human-readable variables. Take the two one for example:

```js
if (passphrase.length <= 0) {
...
} else if (common || isPwned) {
```

This could be:

```js
const noPassword = passphrase.length <= 0

if (noPassword) {
...
} else if (common || isPwned) {
...
```

and so on. For me, this makes reading the logic quick and decisive.

thestephenmarshall (commented):
@patmccler great job on this one so far. I only have one real nit on the async `fetch` which I mentioned down below.

playbook/app/pb_kits/playbook/pb_passphrase/useZxcvbn.js
thestephenmarshall:
```suggestion
    if (noPassword) {
      setPercent('0')
      setVariant('negative')
      setText('\u00A0') //nbsp to keep height constant
    } else if (commonPassword) {
      setPercent('25')
      setVariant('negative')
      setText('This passphrase is too common')
    } else if (weakPassword) {
      setPercent('25')
      setVariant('negative')
      setText('Too weak')
    } else if (averagePassword){
      setPercent('50')
      setVariant('warning')
      setText('Almost there, keep going!')
    } else if (strongPassword) {
      setPercent('100')
      setVariant('positive')
      setText('Success! Strong passphrase')
    }
```

Just an example from above

playbook/app/pb_kits/playbook/pb_passphrase/useHaveIBeenPwned.js
patmccler: Great call. I was thinking "If the api is down, it can never be set to true" but that doesn't hold if it falls over after any request is successful. I'll set a default!

playbook/app/pb_kits/playbook/pb_passphrase/useZxcvbn.js
patmccler: Updated as suggested. Definitely easier to parse at a glance

thestephenmarshall (approved):
Lovely! Great work 🚀

PR #1481, opened 2021-05-04
[GOG-421] Use UX AWS role for SOPS decryption

benlangfeld:
This causes SOPS to assume the UX role when accessing these secrets, which is necessary based on the new permissions model which grants selective access to KMS keys. Use of the key without a role is also retained for Milano's access.

Previously, all members of the Developers group had access to all KMS keys. We're moving toward access only to those keys which protect secrets that are within the remit of specific groups.

Only UX need access to Playbook secrets.

https://nitro.powerhrg.com/runway/backlog_items/GOG-421

### DIFF
--- a/playbook/.sops.yaml
+++ b/playbook/.sops.yaml
@@ -1,13 +1,19 @@
 creation_rules:
-  - filename_regex: production
+  - path_regex: production
     key_groups:
     - kms:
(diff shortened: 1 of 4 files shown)


indiebrain (approved):
Works for me.

stephenagreer (approved):

thestephenmarshall (approved):

batistaDev1113 (approved):

PR #1543, opened 2021-06-28
WIP: initial startup test

thestephenmarshall:
@xjunior I created this PR just from running the initial setup instructions and noticed these changes. I'm mostly concerned with the `yarn.lock` changes, but even the unintentional changes to the distribution files seems a little worrisome for me. WDYT?


xjunior:
@thestephenmarshall IDK how yarn.lock got out of sync in master, but it seems like the yarn.lock file in master is Yarn v1, and therefore wrong. The changes to the files in dist are normal, since while you build the files for the website locally it's also regenerating the files in dist, and the files there are not always up to date with the latest commit in master.

PR #1606, opened 2021-08-31
Flexible dependencies for clients

jasoncypret:
I forked Playbook recently so I could update the app I run to rails 6. Thought I would share back my fork. This makes the versions more flexible to versions greater than or equal to 5.2.4.5 only. Before it used to check that but put an upper limit on rails 6.0. This restriction is not needed to enforce. as clients can use any new version of rails to test.

#### Breaking Changes

I don't think so. This just changes the dependency to be more flexible. I DO NOT BELIEVE a version increase will be needed for Nitro.

#### Runway Ticket URL

[INSERT URL]

#### How to test this

1. Does the site run and deploy? Does it run locally?
2. Does it affect Nitro in anyway? Should we cut a release to double check.


jasoncypret:
@thestephenmarshall  @xjunior I forked Playbook recently so I could update the app I run to rails 6. Thought I would share back my fork. Any concerns for Nitro?

thestephenmarshall (commented):
Need to test this out pretty thoroughly in website docs and Nitro as an alpha.

jasoncypret:
Closed in favor of #1615

PR #1615, opened 2021-09-14
Flexible dependencies for clients

jasoncypret:
I forked Playbook recently so I could update the app I run to rails 6. Thought I would share back my fork. This makes the versions more flexible to versions greater than or equal to 5.2.4.5 only. Before it used to check that but put an upper limit on rails 6.0. This restriction is not needed to enforce. as clients can use any new version of rails to test.

#### Breaking Changes

I don't think so. This just changes the dependency to be more flexible. I DO NOT BELIEVE a version increase will be needed for Nitro.

#### Runway Ticket URL

[INSERT URL]

#### How to test this

1. Does the site run and deploy? Does it run locally?
2. Does it affect Nitro in any way? We Should cut a release to double-check.


jasoncypret:
@thestephenmarshall Did this get is tested in your alpha?

kre8sions (approved):

PR #1725, opened 2022-01-17
[PLAY-74] Allow Legend Kit to use our Status Colors

JelenaD1:
#### Screens
Rails:

Playbook_Design_System

React:

Playbook_Design_System_and_legend_spec_rb_—_playbook

#### Breaking Changes
No. Default color remain the same. Developer can now pass a color prop and use any of the data, status, product and category colors list.

[Link to runway ticket](https://nitro.powerhrg.com/runway/backlog_items/PLAY-74)

#### How to test this
Open a browser and inspect legend kit rails and react code. Add a color prop and pass any desired value from available color list.

#### Checklist:


playbook/app/pb_kits/playbook/pb_legend/_legend.jsx
```diff
@@ -18,7 +18,13 @@ type LegendProps = {
...
     | "data_5"
     | "data_6"
-    | "data_7",
+    | "data_7"
+    | "success"
+    | "warning"
+    | "error"
+    | "info"
+    | "neutral"
+    | "primary",
```
that-jill, line 27: can we alphabetize this list (here and in the ruby file) so that it's more readable?

playbook/app/pb_kits/playbook/pb_legend/legend.rb
```diff
@@ -4,7 +4,7 @@ module Playbook
   module PbLegend
     class Legend < Playbook::KitBase
       prop :color, type: Playbook::Props::Enum,
-                   values: (1..7).map { |n| "data_#{n}" },
+                   values: %w[data_1 data_2 data_3 data_4 data_5 data_6 data_7 success warning error info neutral primary],
```
that-jill, line 7: there is a strong argument here that these values are [magical](https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad#:~:text=In%20programming%2C%20a%20%22magic%20number,in%20more%20than%20one%20place). It may be prudent to extract them into a constant, which (at least on the rails side), can then be referenced in the two places we use it. On the react side, I believe there is a way to define custom datatypes, though I'm not sure whether that precedent has been set in Playbook.

playbook/app/pb_kits/playbook/pb_legend/docs/example.yml
```diff
@@ -1,11 +1,13 @@
...
   rails:
   - legend_default: Default
   - legend_prefix: With Prefix Text
- 
-  
-  
+  - legend_status_colors: With Status Colors
+
+
+
```
that-jill, line 9: I understand that this was preexisting, but I'm not sure why all this extra whitespace is here, and I don't think we should keep it.

playbook/app/pb_kits/playbook/pb_legend/docs/example.yml
```diff
@@ -1,11 +1,13 @@
 examples:
-  
+
   rails:
   - legend_default: Default
   - legend_prefix: With Prefix Text
- 
-  
-  
+  - legend_status_colors: With Status Colors
```
that-jill, line 6: I would love to see more documentation around the allowed colors that can be passed to this component, since it seems a little unintuitive about what they accept

playbook/spec/pb_kits/playbook/kits/legend_spec.rb
```diff
@@ -23,8 +23,9 @@
 
   describe "#classname" do
     it "returns namespaced class name", :aggregate_failures do
-      color = "data_1"
-      expect(subject.new(color: color, dark: true, text: "Text").classname).to eq "pb_legend_kit_data_1 dark"
+      expect(subject.new(text: "Test default").classname).to eq "pb_legend_kit_data_1"
+      expect(subject.new(color: "data_5", dark: true, text: "Text").classname).to eq "pb_legend_kit_data_5 dark"
+      expect(subject.new(color: "success", text: "Test").classname).to eq "pb_legend_kit_success"
```
that-jill, line 28: this expectation could probably use some love / clarification. We're not necessarily checking that the correct classname is applied in every single instance, but instead in named instances -- so it may be prudent to clarify that in the specs and check them singly.

that-jill (commented):
for what it's worth, this kit is also leaving out the `data_8` color. Also, I understand the written intent on the story is to add status colors -- and this will unblock some work for Nova -- but [there are even more colors](https://playbook.powerapp.cloud/visual_guidelines) that playbook defines and this kit will not allow us to use. It may be worth asking @kre8sions whether these colors (especially product/category colors) are also worth adding as well.

playbook/app/pb_kits/playbook/pb_legend/_legend.jsx
JelenaD1: Sure we can. As far as color goes I see data_8 is missing. Do we want to add product and category colors as well? @that-jill @kre8sions .

playbook/app/pb_kits/playbook/pb_legend/docs/example.yml
JelenaD1: We can do that.

playbook/app/pb_kits/playbook/pb_legend/legend.rb
JelenaD1: So I followed the pattern for defining prop values in Playbook. I can extract it and in a constant if that's what you meant.

playbook/app/pb_kits/playbook/pb_legend/_legend.jsx
kre8sions: @JelenaD1 yes

kre8sions (approved):

PR #1742, opened 2022-01-25
[PLAY-104] Remove dist folder tracking

thestephenmarshall:


jasoncypret (approved):
The deployment looks correct to me. Do you know if dist is regenerated with the deployment?

thestephenmarshall:
@jasoncypret indeed it does! [This line](https://github.com/powerhome/playbook/blob/dist-update/Dockerfile#L52) compiles those assets during the "build image" step in the pipeline which is then deployed in a later step to the destination.

PR #1763, opened 2022-02-02
[PLAY-88] Currency kit new abbreviate prop

kyleschechter:
#### Screens

[INSERT SCREENSHOT]

#### Breaking Changes

[Yes/No (Explain)]

#### Runway Ticket URL

[https://nitro.powerhrg.com/runway/backlog_items/PLAY-88](url)

#### How to test this

[INSERT TESTING DETAILS]

#### Checklist:

- [ ] **DEPLOY** Please add the `Milano` label when you are ready for a review.
- [ ] **SCREENSHOT** Please add a screen shot or two.
- [ ] **READ DOCS** Please make sure you have read and understand the [Playbook Release Process](https://github.com/powerhome/playbook/wiki/Playbook-Releases)


playbook/app/pb_kits/playbook/pb_currency/currency.rb
```diff
@@ -84,10 +87,17 @@ def whole_value
...
+        number_to_human(value, units: { thousand: "K", million: "M", billion: "B", trillion: "T" })
+      end
+
       def units_element
         _, decimal_part = amount.split(".")
         if unit.nil?
-          decimal_part.nil? ? ".00" : ".#{decimal_part}"
+          if abbreviate == false
+            decimal_part.nil? ? ".00" : ".#{decimal_part}"
+          end
```
JelenaD1, line 100: I'm wondering if we can optimize this and say `if unit.nil? && abbreviate == false` ?

playbook/spec/pb_kits/playbook/pb_currency/currency_spec.rb
```diff
@@ -13,6 +13,7 @@
   it { is_expected.to define_enum_prop(:size).with_default("md").with_values("sm", "md", "lg") }
   it { is_expected.to define_prop(:symbol).with_default("$").of_type(Playbook::Props::String) }
   it { is_expected.to define_enum_prop(:variant).with_default("default").with_values("default", "light", "bold") }
+  it { is_expected.to define_prop(:abbreviate).of_type(Playbook::Props::Boolean) }
```
JelenaD1, line 16: Maybe will worth it to test the default as well in case someone change it in future.

playbook/app/pb_kits/playbook/pb_currency/currency.rb
kyleschechter: I'll try it out

playbook/spec/pb_kits/playbook/pb_currency/currency_spec.rb
kyleschechter: I'll try that too

playbook/app/pb_kits/playbook/pb_currency/currency.rb
kyleschechter: done

playbook/spec/pb_kits/playbook/pb_currency/currency_spec.rb
kyleschechter: done

JelenaD1 (approved):

playbook/app/pb_kits/playbook/pb_currency/_currency.jsx
```diff
@@ -69,6 +71,20 @@ const Currency = (props: CurrencyProps) => {
...
+      maximumFractionDigits: 1,
+    }).format(whole.split(',').join('')).toString()
+    let output
+    if (arg === 'amount') {
+      output = num.slice(0, -1)
+    } else if (arg === 'unit') {
+      output = num.slice(-1)
+    }
+    return output
+  }
```
thestephenmarshall, line 86:
```suggestion
  const getFormattedNumber = (input) => new Intl.NumberFormat('en-US', {
      notation: 'compact',
      maximumFractionDigits: 1,
    }).format(input)
  
  type AbbrType = 'amount' | 'unit'
  
  const getAbbreviatedValue = (abbrType: AbbrType) => {
    const num = `${getFormattedNumber(whole.split(',').join(''))}`,
          isAmount = abbrType === 'amount',
          isUnit =  abbrType === 'unit'
               
    return isAmount ? num.slice(0, -1) : isUnit ? num.slice(-1) : ''
  }
```

A couple of things about the original code here:

- Lower readability 
- The number formatter should be abstracted, especially to allow future re-use. Doing so also helps readability
  - Functions should strive to do one thing well and preferably only one
- `arg` is way to generic for *any* use case. Name your arguments contextually and don't forget to specify types. Now we can tell exactly the values that can be accepted. 
- The function was not returning any value if neither enumerated value was provided. It would just be `undefined` which would be bad

playbook/app/pb_kits/playbook/pb_currency/_currency.jsx
```diff
@@ -92,18 +108,19 @@ const Currency = (props: CurrencyProps) => {
             dark={dark}
             size={sizes[size]}
         >
-          {`${whole}`}
+          {abbreviate ? getAbbreviatedValue('amount') : whole}
```
thestephenmarshall, line 111: We should steer towards leaving logic out of the templating area. Could you please move this to a local variable?

playbook/app/pb_kits/playbook/pb_currency/_currency.jsx
```diff
@@ -92,18 +108,19 @@ const Currency = (props: CurrencyProps) => {
...
             className="unit"
             color="light"
             dark={dark}
         >
+          {abbreviate ? getAbbreviatedValue('unit') : null}
           <If condition={unit}>
             {unit}
             <Else />
-            {`.${decimal}`}
+            {abbreviate ? '' : `.${decimal}`}
```
thestephenmarshall, line 123: Same as the above comment ^

thestephenmarshall (commented):
Coming along. Got some nits to be addressed.

playbook/app/pb_kits/playbook/pb_currency/_currency.jsx
```diff
@@ -69,6 +71,24 @@ const Currency = (props: CurrencyProps) => {
...
+  const getAbbreviatedValue = (abbrType: AbbrType) => {
+    const num = `${getFormattedNumber(whole.split(',').join(''))}`,
+      isAmount = abbrType === 'amount',
+      isUnit =  abbrType === 'unit'
+    return isAmount ? num.slice(0, -1) : isUnit ? num.slice(-1) : ''
+  }
+
+  const getAmount = abbreviate ? getAbbreviatedValue('amount') : whole,
+    getAbbreviation = abbreviate ? getAbbreviatedValue('unit') : null,
+    getDecimalValue = abbreviate ? '' : `.${decimal}`
```
JelenaD1, line 90: I'm curious why you don't  define a variable with a `const` keyword for last two examples?

playbook/spec/pb_kits/playbook/pb_currency/currency_spec.rb
```diff
@@ -13,6 +13,7 @@
   it { is_expected.to define_enum_prop(:size).with_default("md").with_values("sm", "md", "lg") }
   it { is_expected.to define_prop(:symbol).with_default("$").of_type(Playbook::Props::String) }
   it { is_expected.to define_enum_prop(:variant).with_default("default").with_values("default", "light", "bold") }
+  it { is_expected.to define_prop(:abbreviate).with_default(false).of_type(Playbook::Props::Boolean) }
```
JelenaD1, line 16: Do we expect any change in a classname when `abbreviate` prop is being passed?

playbook/app/pb_kits/playbook/pb_currency/_currency.jsx
kyleschechter: I actually saw this done a couple times in other files. When you have a bunch of `const`s one after the other, you can separate them by commas and old put one `const` before the first one @JelenaD1

kyleschechter: *only

playbook/spec/pb_kits/playbook/pb_currency/currency_spec.rb
kyleschechter: good question. No, because the only thing that is changing is the values of the variables being displayed in the return. The class name will not change

playbook/app/pb_kits/playbook/pb_currency/_currency.jsx
JelenaD1: I did not know that. Thanks for explaining.

playbook/spec/pb_kits/playbook/pb_currency/currency_spec.rb
kyleschechter: Let me rephrase that 'no', I don't think we can expect any change

playbook/app/pb_kits/playbook/pb_currency/_currency.jsx
thestephenmarshall: Mostly just to keep word trash down in the code. Seeing `const` a bunch of times (unnecessarily) is going to reduce readability and make the code appear more monotonous. What is best about this approach is you can now visually group variables of concern together.

playbook/spec/pb_kits/playbook/pb_currency/currency_spec.rb
thestephenmarshall: I am also looking at the Jest test in place. I think there are some necessary additions.

thestephenmarshall (changes requested):
@kyleschechter I see that `currency_spec.rb` and `currency.test.js` are testing two different scenarios. We want to ensure that both the Ruby and the React versions of each test are testing the same cases respectively.

PR #1772, opened 2022-02-09
Lightbox Component

jasperfurniss:
React Kit Coverage
Screen Shot 2022-02-14 at 2 07 18 PM

https://user-images.githubusercontent.com/84343331/167418307-595857ac-60dd-44cb-80f8-9d0c605d5c3e.mov


thestephenmarshall (changes requested):
This one is going to be a little tricky. Need to determine the circular dependency situation out.

playbook/app/pb_kits/playbook/pb_lightbox/_lightbox_header.jsx
```diff
@@ -0,0 +1,103 @@
...
+                </Flex>
+              </Flex>
+            </FlexItem>
+          </If>
+          <FlexItem flex={1}>
+            <Flex justify="end">
+              <Title
+                  dark
+                  size={4}
+                  text="All Photos"
```
Ivancandido11, line 92: Im assuming at some point this will be used for something, instead of just text. I can remove it though if we don't want it here for now.

thestephenmarshall (commented):
This seems to be missing some pretty basic interactivity:

- ESC to close
- Left/Right Arrow navigation

I think these should be added at the very least before this PR will be ready.

playbook/app/pb_kits/playbook/pb_lightbox/Carousel/Slides.jsx
```diff
@@ -0,0 +1,36 @@
...
+  urls: Array<string>,
+  current: number,
+  onChange: (index: number) => void,
+  onClick: (index: number) => void,
+}
+
+export default function Slides({
+  urls = [],
+  current = 0,
+  // onClick = noop,
```
thestephenmarshall, line 18: Remove

thestephenmarshall (approved):
Approving so the hold is released. Still more work to do.

thestephenmarshall:
@jasperfurniss have a look, but this one is good to go imo!

PR #1866, opened 2022-05-02
[PLAY-162] Add ISC license

thestephenmarshall:
React Kit Coverage
https://nitro.powerhrg.com/runway/backlog_items/PLAY-162


LICENSE
```diff
@@ -0,0 +1,15 @@
+ISC License
+
+Copyright (c) {{ year }}, {{ author }}
```
jasoncypret, line 3: Don't we need to fill this out?

thestephenmarshall: `2022, Power Home Remodeling` how should that last part be?

PR #1883, opened 2022-05-06
Migration to the new registry (Harbor)

gruberdev:
React Kit Coverage
#### Breaking Changes

No. Only adds the new registry to the repository.

#### Runway Ticket URL

[GOG-594](https://nitro.powerhrg.com/runway/backlog_items/GOG-594)


thestephenmarshall:
@gruberdev this makes sense to me, but could you get this to deploy please?

Jenkinsfile
```diff
@@ -19,7 +19,7 @@ app.build(
 
   stage('Code Checkout') {
     scmVars = checkout scm
-    appImage = "quay.io/powerhome/playbook:${git.triggeringCommit(scmVars)}"
+    appImage = "image-registry.powerapp.cloud/playbook/playbook:${git.triggeringCommit(scmVars)}"
```
jasperfurniss, line 22: We're going to need to tweak 3 more files i believe in order for this to work. @gruberdev 
Please look at the following 3: 

1) https://github.com/powerhome/playbook/blob/284aea6632b3bada43394f062453376367475771/playbook-website/bin/deployer#L10

2) https://github.com/powerhome/playbook/blob/be09fa5bdc35556407d8f720acd99ed847c73f16/playbook-website/config/deploy/templates/deployment.yaml.erb#L23

3) https://github.com/powerhome/playbook/blob/be09fa5bdc35556407d8f720acd99ed847c73f16/Jenkinsfile#L22

4) https://github.com/powerhome/playbook/blob/be09fa5bdc35556407d8f720acd99ed847c73f16/Makefile#L8

thestephenmarshall (changes requested):
Not sure what the scope on this PR is. Was it the intent to change multiple files? Is this PR still WIP? I think we just need a little more clarification as to what is happening here.

gruberdev:
> @gruberdev this makes sense to me, but could you get this to deploy please?

Indeed, Milano uses a missing reference in a template to deploy images from the registry; I believe it should work after https://github.com/powerhome/playbook/pull/1883/commits/1323715f61fd7c188d04e0a06a786692f0d0ae8b.

> Would we need to tweak 3 more files i believe in order for this to work? @gruberdev Please look at the following files and see if they are relevant for this Pull Request. Also, we definitely need a full description on this work. We don't know the intention behind this pull request.
> 
> 1. https://github.com/powerhome/playbook/blob/284aea6632b3bada43394f062453376367475771/playbook-website/bin/deployer#L10
> 2. https://github.com/powerhome/playbook/blob/be09fa5bdc35556407d8f720acd99ed847c73f16/playbook-website/config/deploy/templates/deployment.yaml.erb#L23
> 3. https://github.com/powerhome/playbook/blob/be09fa5bdc35556407d8f720acd99ed847c73f16/Jenkinsfile#L22
> 4. https://github.com/powerhome/playbook/blob/be09fa5bdc35556407d8f720acd99ed847c73f16/Makefile#L8

The reference to the master image used seems to be of a specific version, as to where I don't have enough context to attest to whether it is adequate to use the latest master build to replace it. However, I've added the image generated from last build in its place; let me know if this change works.

> Not sure what the scope on this PR is. Was it the intent to change multiple files? Is this PR still WIP? I think we just need a little more clarification as to what is happening here.

This PR aims to migrate the image building and storage from `Quay.io` to our newer internal registry that uses `Harbor` as the base registry management platform. 

The PR's original purpose worked as intended since the image built was successfully uploaded to the newer internal container registry. It is worth noting I was not aware at the time of the additional changes that would have to be made, so it could be considered a WIP. Using the feedback provided until now, I made a few further changes, and any feedback is welcome.

Makefile
```diff
@@ -5,7 +5,7 @@ ifdef TTY
 	AWS_CREDS_MOUNT = --mount type=bind,source=$(HOME)/.aws/credentials,destination=/root/.aws/credentials,readonly
 endif
 
-DEPLOYER_IMAGE = quay.io/powerhome/deployer:master-c94bf553840b07335fbb8904d5a9963dd5ffce00-336
+DEPLOYER_IMAGE = image-registry.powerapp.cloud/app/deployer@sha256:349dfeda19a069bfe1167f7369acbd91f313bc459cd73ed6082a868c2a3e626b
```
benlangfeld, line 8: Why not continue to use an image tag that allows us to compare with the APP repo?

benlangfeld (approved):

thestephenmarshall (approved):

PR #1903, opened 2022-05-13
[Play-136] add ability to select weeks instead of days date picker kit

dmcindoe:
#### Screens

[INSERT SCREENSHOT]

#### Breaking Changes

No, adding new Prop to kit with a default value of false. 

#### Runway Ticket URL

(https://nitro.powerhrg.com/runway/backlog_items/PLAY-136)

#### How to test this

Implement Kit with Prop active to make sure work

#### Checklist:

- [ ] **DEPLOY** Please add the `Milano` label when you are ready for a review.
- [ ] **SCREENSHOT** Please add a screen shot or two.
- [ ] **READ DOCS** Please make sure you have read and understand the [Playbook Release Process](https://github.com/powerhome/playbook/wiki/Playbook-Releases)


playbook/app/pb_kits/playbook/pb_date_picker/_date_picker.jsx
```diff
@@ -40,7 +40,7 @@ type DatePickerProps = {
   onChange: (String) => void,
   pickerId?: String,
   placeholder?: String,
-  plugins?: Boolean,
+  plugins?: "ms" | "ws",
```
thestephenmarshall, line 43: Apparently this `plugins` prop was not implemented in a very sensible way from my perspective. The way it currently stands, this boolean simply [disables one plugin that is being configured inside the helper itself](https://github.com/powerhome/playbook/blob/master/playbook/app/pb_kits/playbook/pb_date_picker/date_picker_helper.js#L56). I am having trouble understanding the direction of the original code you are replacing as it seem a lot more reasonable, from a Playbook kit perspective, to instead have props that do a better job of explaining what their purpose is compared to just `plugins`. These props would simply map to each individual feature. Let's your change here as an example:

```tsx
// instead of this plurally named enumerable
plugins?: "ms" | "ws"
```
```tsx
// add individual props, one for each feature
monthSelect?: boolean,
weekSelect?: boolean,
```

This would still achieve the same effect, but be a lot less vague than the original implementation. If you chose to do this as described above, you still have to fool-proof it a little bit incase someone enables two or more features that do not work together, so perhaps you could then do some consolidation:

```tsx
// add a selection type feature
selectionType?: "month" | "week"
```

In summary, this would be the best option which can both better explain what is going in to the kit and what will be produced without introducing any crossing feature paths like month vs. day selection.

playbook/app/pb_kits/playbook/pb_date_picker/_date_picker.jsx
```diff
@@ -74,7 +74,7 @@ const DatePicker = (props: DatePickerProps) => {
     onChange = () => {},
     pickerId,
     placeholder = 'Select Date',
-    plugins = false,
+    plugins = "",
```
thestephenmarshall, line 77:
```suggestion
    plugins = false,
```

thestephenmarshall: Keeping this false would be best as we will still check the value of this prop before enabling any plugins.

playbook/app/pb_kits/playbook/pb_date_picker/date_picker_helper.js
```diff
@@ -52,9 +53,15 @@ const datePickerHelper = (config) => {
     }
   }
 
-  const setMonthAndYearPlugin = () => (
-    plugins ? [ monthSelectPlugin({ shorthand: true, dateFormat: 'F Y', altFormat: 'F Y' }) ] : []
-  )
+  const setPlugin = () => {
+    let p 
+    if (plugins === "ms") {
```
thestephenmarshall, line 58:
```tsx
    if (selectionType === "month" || plugins === true) {
```

In this very particular case, we would be requiring that a new major release of Playbook be created as now the `plugins` feature has been changed from how it originally worked. I don't think we will want to do this within this unit of work. Instead, what would work best is to introduce the new feature atop the existing feature as you can see in the logic above. Here we are still checking for `plugins === true` thereby keeping the boolean toggle in place to enable only month selection just as it was before.

playbook/app/pb_kits/playbook/pb_date_picker/date_picker_helper.js
```diff
@@ -52,9 +53,15 @@ const datePickerHelper = (config) => {
...
+     p = [ monthSelectPlugin({ shorthand: true, dateFormat: 'F Y', altFormat: 'F Y' }) ] 
+    } else if ( plugins === "ws") {
+      p = [ weekSelect({})]
```
thestephenmarshall, line 61: This part is fine since it is introducing a brand new feature 👍🏼

playbook/app/pb_kits/playbook/pb_date_picker/date_picker_helper.js
```diff
@@ -98,12 +105,12 @@ const datePickerHelper = (config) => {
...
     }],
     onChange: [(selectedDates, dateStr) => {
-      onChange(dateStr, selectedDates)
+      onChange(dateStr, selectedDates) 
     }],
     onYearChange: [() => {
       yearChangeHook()
     }],
-    plugins: setMonthAndYearPlugin(),
+    plugins: setPlugin(),
```
thestephenmarshall, line 113: 👍🏼

thestephenmarshall (approved):

playbook/app/pb_kits/playbook/pb_date_picker/date_picker.rb
```diff
@@ -42,7 +42,11 @@ class DatePicker < Playbook::KitBase
       prop :placeholder, type: Playbook::Props::String,
                          default: "Select Date"
       prop :plugins, type: Playbook::Props::Boolean,
-                     default: false
+                     default: false,
+                     deprecated: true
+      prop :selection_type, type: Playbook::Props::Enum,
+                            values: %w[week month],
```
thestephenmarshall, line 48: See below

```suggestion
                            values: %w[week month none],
```

playbook/spec/pb_kits/playbook/kits/date_picker_spec.rb
```diff
@@ -26,6 +26,7 @@
   it { is_expected.to define_prop(:picker_id).of_type(Playbook::Props::String).that_is_required }
   it { is_expected.to define_prop(:placeholder).of_type(Playbook::Props::String) }
   it { is_expected.to define_prop(:plugins).of_type(Playbook::Props::Boolean).with_default(false) }
+  it { is_expected.to define_prop(:selection_type).of_type(Playbook::Props::Enum).with_default("month") }
```
thestephenmarshall, line 29: This has to be updated as well.

```suggestion
  it { is_expected.to define_prop(:selection_type).of_type(Playbook::Props::Enum).with_default("none") }
```

thestephenmarshall (commented):
@dmcindoe See my latest comments. I found one thing that will conflict with [time selection](https://github.com/powerhome/playbook/pull/1906)

thestephenmarshall (changes requested):
I just noticed one thing that is missing based on the HiFi mocks:
Screen Shot 2022-05-24 at 2 25 49 PM

When you hover over a row, there is a shadow and background color that should appear. Also, I noticed that the HiFi mock doesn't explicitly show what a selected week looks like.

thestephenmarshall (approved):

PR #2071, opened 2022-09-14
[PLAY-219] Typescript Conversion: Date Picker

augustomallmann:
#### Screens

Screen Shot 2022-09-14 at 2 14 39 PM

#### Breaking Changes

No breaking changes, typescript conversion of the Date Picker kit

#### Runway Ticket URL

[[PLAY-219]](https://nitro.powerhrg.com/runway/backlog_items/PLAY-219)

#### How to test this

- Check console for any errors
- Try reproducing the kit with different props 
- Run `./test.sh` 

#### Checklist:


playbook/app/pb_kits/playbook/pb_date_picker/date_picker_helper.ts
```diff
@@ -54,7 +73,7 @@ const datePickerHelper = (config, scrollContainer) => {
     }
   }
   const calendarResizer = () => {
-    const cal = document.querySelector(`#cal-${pickerId}.open`)
+    const cal = document.querySelector(`#cal-${pickerId}.open`) as HTMLElement | null
```
thestephenmarshall, line 76:
```suggestion
    const cal = document.querySelector(`#cal-${pickerId}.open`) as HTMLElement
```

I would almost want to remove `null` more here then..

playbook/app/pb_kits/playbook/pb_date_picker/date_picker_helper.ts
```diff
@@ -54,7 +73,7 @@ const datePickerHelper = (config, scrollContainer) => {
...
     const parentInput = cal.parentElement
     if (cal.getBoundingClientRect().right > window.innerWidth) {
```
thestephenmarshall, line 78: ... use the safe operator in this conditional

```suggestion
    if (cal?.getBoundingClientRect().right > window.innerWidth) {
```

playbook/app/pb_kits/playbook/pb_date_picker/date_picker_helper.ts
```diff
@@ -70,10 +89,10 @@ const datePickerHelper = (config, scrollContainer) => {
     let pluginList = []
 
     // month and week selection
-    if (selectionType === "month" || plugins === true) {
+    if (selectionType === "month" || plugins.length > 0) {
```
thestephenmarshall, line 92: This part confuses me a little bit, but I think what is going on here is the result of converting to TS which now expects an array _always_. Does that sound right?

augustomallmann: yes, as plugins is an array, it starts throwing the error `This condition will always return 'false' since the types 'Plugin<{}>[]' and 'boolean' have no overlap.`. The type comes from `flatpickr`, so the way I fixed it was by checking if the array is bigger than 0. Does it make sense?

thestephenmarshall (approved):

PR #2271, opened 2023-01-20
Pigeon Map POC

nidaqg:
### POC on using [Pigeon Maps](https://pigeon-maps.js.org/docs/) as an option.

- Much easier to use, does provide many options through their API, Different maptiles can also be used, have not spend time exploring it yet. 

#### Screens

Screenshot 2023-01-20 at 8 41 49 AM


nidaqg:
Closing this PR as we are not using Pigeon Map for the Map kit.

PR #2281, opened 2023-01-25
[Play-554] button focus state

Israel-Molestina:
#### Screens

Screen Shot 2023-01-25 at 9 56 33 AM

#### Breaking Changes

NO

#### Runway Ticket URL

[Runway Ticket](https://nitro.powerhrg.com/runway/backlog_items/PLAY-554)

#### How to test this

Tab through buttons state make sure they all have focus state

#### Checklist:


playbook/app/pb_kits/playbook/pb_button/_button_mixins.scss
```diff
@@ -5,6 +5,7 @@
 @import "../tokens/spacing";
 @import "../tokens/transition";
 @import "../tokens/typography";
+@import "../tokens/accessibility";
```
jasoncypret, line 8: This is not really a token. Whats our strategy for re-useable mixins? @jasperfurniss

playbook/app/pb_kits/playbook/pb_button/_button_mixins.scss
```diff
@@ -86,6 +88,10 @@ $pb_button_border_width: 0px;
     &:hover {
       @include pb_button_hover;
     }
+    &:active {
+      transition: none;
```
jasoncypret, line 92: @Israel-Molestina Can we not add transitions? I was going to suggest adding transitions as the different state switched are jarring. See how smooth the states (including focus) is on the [checkbox kit](https://playbook.powerapp.cloud/kits/checkbox/react)

playbook/app/pb_kits/playbook/pb_button/_button_mixins.scss
jasperfurniss: Yeah I was not expecting this to be a token. Only a mixin. Because we will indeed be using similar patterns across multiple kits.

Israel-Molestina: Do we have a mixins folder? Or should one be created? We also have titles.scss in the tokens folder which is a mixin folder. So we could move both to a new mixin folder?

jasperfurniss: Ideally, the mixins would live close to their corresponding kit. For example, button mixin should live close to the button component. 

In our case, however, we are attempting to create a mixin that is not kit specific and will span across multiple kits. 

I think making a new mixin folder for now is better structure for us, and clears up confusion.  
So, yes.

playbook/app/pb_kits/playbook/pb_button/_button_mixins.scss
Israel-Molestina: @jasoncypret I've added transition to the focus states. But when I talked to Ramon he did not want transitions on the active states. This way it is very obvious that you just clicked the button

jasperfurniss (approved):

PR #2313, opened 2023-02-13
Moved focus SCSS to utilities folder

nidaqg:
Patch fix for error in Nitro due to not having the focus styles in dist


playbook/app/pb_kits/playbook/pb_button/_button_mixins.scss
```diff
@@ -5,7 +5,7 @@
 @import "../tokens/spacing";
 @import "../tokens/transition";
 @import "../tokens/typography";
-@import "../mixins/accessibility";
+@import "../utilities/focus";
```
jasoncypret, line 8: This is not a utility. Utilities can be thought of as class helpers like tail winds. This has no output

jasperfurniss: We can definitely put it somewhere else. 👍 
The goal was to put it in central spot that wouldn't cause issues with the current release, and then later move to a better location.

Right now, we don't have a sort of global place for generic mixins (not kit-specific) that we use across multiple kits.

The `focus` mixin will almost definitely be used for other kits, so that's why i was thinking it shouldn't live next to the button kit itself. 

Where would you like it to live?

jasperfurniss (approved):

PR #2631, opened 2023-07-07
[PLAY-802] Implementing Responsive Title

carloslimasd:
**What does this PR do?**
It implements responsive Title sizes so that we can dynamically change Title sizes based on screen size.

image

1. Go to Title Kit on Playbook Website
2. Scroll down to **Responsive** example

#### Checklist:


playbook/app/pb_kits/playbook/pb_title/title.rb
```diff
@@ -6,9 +6,7 @@ class Title < Playbook::KitBase
       prop :color, type: Playbook::Props::Enum,
                    values: [nil, "default", "light", "lighter", "success", "error", "link"],
                    default: nil
-      prop :size, type: Playbook::Props::Enum,
-                  values: [1, 2, 3, 4],
-                  default: 3
+      prop :size, default: 3
```
markdoeswork, line 9: Should it still have a type?

```suggestion
            prop :size, type: Playbook::Props::Number,
                               default: 3
```

playbook/app/pb_kits/playbook/pb_title/_title.tsx
```diff
@@ -36,9 +39,25 @@ const Title = (props: TitleProps): React.ReactElement => {
   const ariaProps: {[key: string]: string | number} = buildAriaProps(aria)
   const dataProps: {[key: string]: string | number} = buildDataProps(data)
   const getBold = bold ? '' : 'thin'
+
+  const generateSizeClasses = () => {
+    let css = ''
+    
+    if (typeof size == "number" || typeof size == "string") {
```
markdoeswork, line 46: I think I would just make it a number, but I might be missing some kind of context

playbook/app/pb_kits/playbook/pb_title/title.rb
markdoeswork: I see the prop is an object

playbook/app/pb_kits/playbook/pb_title/docs/_title_responsive.html.erb
```diff
@@ -0,0 +1 @@
+<%= pb_rails("title", props: { text: "Responsive Title", tag: "h1", size: {xs: "3", sm: "2", md: "1"} }) %>
```
markdoeswork, line 1: Same string vs number concept I would use `{xs: 3, sm: 2, md: 1}` if it works

playbook/app/pb_kits/playbook/pb_title/_title.tsx
nidaqg: Having it as just number would definitely make more sense @markdoeswork , but the kit right now allows both number or string as the size and we have to provide for both so we don't break current implementations

playbook/app/pb_kits/playbook/pb_title/title.rb
nidaqg: good catch! This definitely needs a type, I would say keep it Enum so we can have the values array as well, what do you think @markdoeswork ?

carloslimasd: I am talking to @markdoeswork about this because we would have to have mixed types here. It would be exactly like the React version, allowing a string, number, or object that would have a key as string and the value as string or number.

carloslimasd: We talked and we would have to create a new prop type just for this one. In this case, we preferred to leave it without the type.

markdoeswork (dismissed):
LG2M

playbook/app/pb_kits/playbook/pb_title/docs/_title_responsive.jsx
```diff
@@ -0,0 +1,16 @@
...
+      <Title
+          size={{xs: "3", sm: "2", md: "1"}}
+          text="Responsive Title"
+          {...props}
+      />
+    </>
+  )
+}
+
+export default TitleResponsive
```
nidaqg, line 16: looks good @carloslimasd ! Great work on this one! I think we should definitely add in some copy for the doc examples though, we can ask Colin to provide some.

jasperfurniss (approved):

PR #2633, opened 2023-07-07
[PLAY-903] Circle Icon with Emojis

nidaqg:

- Creates new 'emoji' prop for IconCircle kit so it can accept HTML emojis
- Allows kit to take emojis + unicode for emojis
- validates that text passed in is an emoji before rendering to page
-  Only render icon OR emoji so both are not rendered if user tries to pass in both props
- Works for react and rails

Screenshot 2023-07-07 at 2 44 25 PM


playbook/app/pb_kits/playbook/pb_icon_circle/_icon_circle.tsx
```diff
@@ -26,23 +27,41 @@ type IconCircleProps = {
...
   const ariaProps = buildAriaProps(aria)
   const dataProps = buildDataProps(data)
   const classes = classnames(buildCss('pb_icon_circle_kit', size, variant), globalProps(props), className)
 
+  const isValidEmoji = (emoji: string) => {
+    // Using regular expression to check if the string is a valid emoji/emoji Unicode
+    const emojiRegex = /^(\p{Emoji}|\uFE0F)+$/u;
+    return emojiRegex.test(emoji);
+  };
+
```
nidaqg, line 41: Added validation for emojis. The emoji prop takes a string so added this validation to make sure that a random string being passed in would not just render that string to the page. Tested this and if anything is passed to the prop that is not a valid emoji or emoji unicode, nothing renders to the circle. This is the same as for the icon prop, if anything is passed to icon prop that is not a valid icon, it renders nothing to the circle.

playbook/app/pb_kits/playbook/pb_icon_circle/_icon_circle.tsx
```diff
@@ -26,23 +27,41 @@ type IconCircleProps = {
...
         {...dataProps}
         className={classes}
         id={id}
     >
-      <Icon
-          dark={dark}
-          icon={icon}
-      />
+      {
+        icon && !emoji && (
```
nidaqg, line 50: Not sure on this, but added in this check because without it, if a dev passes in icon AND emoji prop, both get rendered to the circle

playbook/app/pb_kits/playbook/pb_icon_circle/_icon_circle.tsx
```diff
@@ -26,23 +27,41 @@ type IconCircleProps = {
...
+          <Icon
+            dark={dark}
+            icon={icon}
+          />
+        )
+      }
+      {
+        emoji && !icon && isValidEmoji(emoji) && (
```
nidaqg, line 58: same as above, + using the emoji validation

playbook/app/pb_kits/playbook/pb_icon_circle/icon_circle.rb
```diff
@@ -3,14 +3,21 @@
...
                   values: %w[xs sm md base lg xl],
                   default: "md"
       prop :variant, type: Playbook::Props::Enum,
                      values: %w[default royal blue orange purple teal red yellow green],
                      default: "default"
 
+      def valid_emoji(emoji)
+        emoji_regex = /\p{Emoji}/
+        emoji_regex.match?(emoji)
+      end
```
nidaqg, line 19: similar validation for emojis as the one on the react side

playbook/app/pb_kits/playbook/pb_icon_circle/icon_circle.rb
```diff
@@ -3,14 +3,21 @@
 module Playbook
   module PbIconCircle
     class IconCircle < Playbook::KitBase
-      prop :icon, required: true
+      prop :icon
```
nidaqg, line 6: Took out 'required' since we can now use 'icon' OR 'emoji' prop.

playbook/spec/pb_kits/playbook/kits/icon_circle_spec.rb
```diff
@@ -26,9 +30,5 @@
       expect(subject.new(icon: icon, variant: "purple", size: "lg").classname).to eq "pb_icon_circle_kit_lg_purple"
       expect(subject.new(icon: icon, classname: "additional_class").classname).to eq "pb_icon_circle_kit_md_default additional_class"
     end
-
-    it "raises an error when not given an icon" do
-      expect { subject.new({}) }.to raise_error(Playbook::Props::Error)
-    end
```
nidaqg, line 32: Since we removed 'required' from icon, had to remove the test for it as well

playbook/app/pb_kits/playbook/pb_icon_circle/_icon_circle.scss
```diff
@@ -21,6 +21,10 @@ $pb_icon_circle_sizes: (
   overflow: hidden;
   position: relative;
 
+  .icon_circle_emoji {
+    font-family: sans-serif;
```
nidaqg, line 25: This was a tricky one. I noticed that the emojis were not aligning correctly, they were off by 2ish pixels, even with flex, align, etc. After some research I found that different fonts render emojis differently and some fonts can cause misalignment. sans-serif seemed to be the consensus on most basic font that would render emojis without misalignment and it did in fact fix the issue.

jasperfurniss (approved):

jasperfurniss (approved):

PR #2650, opened 2023-07-17
Refactor Dialog Kit Close on Outside Click

gavhuang:
**What does this PR do?**
- Refactors the Dialog Rails kit's "click outside to close" feature to not be dependent on the `data-open-dialog` attribute
- If not improved, this could block the SweetAlert replacement ([see POC: PBNTR-62](https://github.com/powerhome/nitro-web/pull/33256))

**How to test?**
1. Go to the Dialog Rails kit
2. Open a dialog, then click outside to close
3. Go to the "Overlay Click" example. You should NOT be able to close.

#### Checklist:
- [X] **LABELS** Add a label: `enhancement`, `bug`, `improvement`, `new kit`, `deprecated`, or `breaking`. See [Changelog & Labels](https://github.com/powerhome/playbook/wiki/Changelog-&-Labels) for details.
- [X] **DEPLOY** I have added the `milano` label to show I'm ready for a review.

### DIFF
--- a/playbook/app/pb_kits/playbook/pb_dialog/dialogHelper.js
+++ b/playbook/app/pb_kits/playbook/pb_dialog/dialogHelper.js
@@ -26,10 +26,9 @@ const dialogHelper = () => {
       if (dialogParentDataset.overlayClick === "overlay_close") return
 
       const clickedOutsideDialogBox = event.target.classList.contains("pb_dialog_rails")
-      const closeDialog = () => dialogElement.close()
 
       if (clickedOutsideDialogBox) {
-        closeDialog()
+        dialogElement.close()
       }
     })
   })
--- a/playbook/app/pb_kits/playbook/pb_dialog/dialogHelper.js
+++ b/playbook/app/pb_kits/playbook/pb_dialog/dialogHelper.js
@@ -29,6 +29,7 @@ const dialogHelper = () => {
 
       if (clickedOutsideDialogBox) {
         dialogElement.close()
+        event.stopPropagation()
       }
     })
   })
--- a/playbook/app/pb_kits/playbook/pb_dialog/dialogHelper.js
+++ b/playbook/app/pb_kits/playbook/pb_dialog/dialogHelper.js
@@ -1,6 +1,7 @@
 const dialogHelper = () => {
   const openTrigger = document.querySelectorAll("[data-open-dialog]");
   const closeTrigger = document.querySelectorAll("[data-close-dialog]");
+  const dialogs = document.querySelectorAll(".pb_dialog_rails")
 
   openTrigger.forEach((open) => {
     open.addEventListener("click", () => {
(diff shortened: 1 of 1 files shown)


playbook/app/pb_kits/playbook/pb_dialog/dialogHelper.js
```diff
@@ -32,6 +18,20 @@ const dialogHelper = () => {
...
     });
   });
+
+  // Close dialog box on outside click
+  dialogs.forEach((dialogElement) => {
+    dialogElement.addEventListener("click", (event) => {
+      const dialogParentDataset = dialogElement.parentElement.dataset
+      if (dialogParentDataset.overlayClick === "overlay_close") return
+
+      const clickedOutsideDialogBox = event.target.classList.contains("pb_dialog_rails")
```
gavhuang, line 28: @markdoeswork would we ever have a case where 2 dialogs would be open at once? If not, this is fine.

markdoeswork: Yes, Just as the sales mentor form has a specific dialog for "blank mentor" I would expect nitro devs to want dialogs for multiple reasons.

Or use cases where they're using some number of the 6 different status alert props

gavhuang: @markdoeswork yeah that's fine but would they be open at the exact same time? If so, then on any click outside, both would close. In my mind, we would never have a case where two or more would be open at once right?

jasperfurniss: @gavhuang That might be a little more of a product question for @co1in-1ewis I'd say...
I think there could be places where we want two to stay open. _But_ i'm also fine with making that harder to do... because it feels like a very rare case. @nidaqg @co1in-1ewis and I just looked at something similar for popover with filter. But let's check in with Colin and see.

nidaqg: It may have changed, but there were some places in Nitro where the code was set up so that a dialog was nested within a dialog, so a full height dialog with another dialog trigger inside it. Would this refactor effect that sort of usecase @gavhuang ?

gavhuang: @nidaqg good point! To address those situations (assuming this is an edge case), we could prevent both dialogs from closing on outside click with the prop.

jasperfurniss: Screenshot 2023-07-18 at 1 37 16 PM
Yes. I think it's good. I would just resolve the weird spacing on that button first. @gavhuang

gavhuang: @jasperfurniss this was only a local test of multiple dialog boxes. So not included in this PR :)

jasperfurniss (dismissed):

nidaqg (approved):

Israel-Molestina (approved):

jasperfurniss (approved):

PR #2673, opened 2023-07-31
Use Internal npm Registry

dhhuynh2:
We have a new internal npm proxy registry that utilizes Verdaccio to cache packages. This PR updates the necessary files to utilize the registry.

https://github.com/powerhome/npm-registry

https://portal.powerapp.cloud/docs/default/system/npm-registry


benlangfeld:
What should we do about the following bits?

* https://github.com/powerhome/playbook/blob/350784ebd4f38e37ba549ada3a9c21cf1f6f09a3/renovate.json#L30-L42
* https://github.com/powerhome/playbook/blob/350784ebd4f38e37ba549ada3a9c21cf1f6f09a3/yarn.secrets.env#L1
* https://github.com/powerhome/playbook/blob/350784ebd4f38e37ba549ada3a9c21cf1f6f09a3/.npmrc#L1
* https://github.com/powerhome/playbook/blob/350784ebd4f38e37ba549ada3a9c21cf1f6f09a3/.github/workflows/github-actions-alpha-version.yml#L16
* https://github.com/powerhome/playbook/blob/master/Jenkinsfile#L19-L28
* https://github.com/powerhome/playbook/blob/master/Dockerfile#L75-L76
* https://github.com/powerhome/playbook/blob/master/docker-compose.yml#L5-L7

As for Renovate, that will no longer be able to propose dependency upgrades, right?

dhhuynh2:
> 

We should be able to remove all of these references. As long as it calls the registry, it'll handle auth with the token. Given that it appears to be the only secret in `yarn.secrets.env`, I suppose we can just remove it altogether? I'll flip it into draft while I do that.

For Renovate, just noticed some issues with [guardians owned applications](https://developer.mend.io/github/powerhome/milano). Assumed it'd use `.yarnrc` or `.yarnrc.yml` configurations for some reason, but seems we have to tweak `hostRules` in the renovate configs. Will focus on fixing that before we go any further.

benlangfeld:
> Edit: Ah - just realized the registry isn't public, meaning Renovate can't hit it. HAProxy won't redirect that either. Seems obvious in hindsight. Will brainstorm solutions.

Socratic method strikes again 😄

benlangfeld:
> @dhhuynh2 @benlangfeld @ryanmrodriguez
> 
> Is there anything I can do to keep this one moving forward? Could we get a PR summary on the benefits of what this Pull Request is doing/adding?
> 
> Thanks so much all!

This is in draft status. Please wait for it to be finalised and come out of draft.

dhhuynh2:
> @dhhuynh2 @benlangfeld @ryanmrodriguez
> 
> Is there anything I can do to keep this one moving forward? Could we get a PR summary on the benefits of what this Pull Request is doing/adding?
> 
> Thanks so much all!

Hey Jasper, this PR is still actively being worked on. We've been working through authentication issues with renovate and dependabot that were blocking it from being merged, which we think we've resolved. Will be sure to let you know when it's ready to open back up for review.

As far as a summary on what the PR is doing, you could skim over this PR description: https://github.com/powerhome/npm-registry/pull/1

This PR updates `yarn.lock` and `yarnrc.yml` to correctly use the registry, thus allowing it to use the registry and get the benefits listed in that description (reducing fontawesome registry calls, more cache hits, etc).

dhhuynh2:
Merging this will probably cause yarn.lock conflicts in all PRs, which will need to be resolved via:

1. `git fetch` to pull from remote
2. `git checkout origin/master yarn.lock` to take master's yarn.lock
3. `yarn` to update it with any (potential) dependency changes)

jasperfurniss:
@dhhuynh2 @benlangfeld Adding a milano, and possibly more importantly, an alpha label, so i can make sure our release scripts work as expected.

renovate.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": [
-    "config:base"
+    "config:base",
```
benlangfeld, line 3:
```suggestion
```

This is already included in https://github.com/powerhome/renovate-config/blob/main/default.json#L3

thestephenmarshall (approved):

thestephenmarshall (dismissed):

.github/workflows/github-actions-alpha-version.yml
```diff
@@ -13,15 +13,18 @@ jobs:
...
       BRANCH_NAME: ${{ github.head_ref }}
-      FONTAWESOME_NPM_AUTH_TOKEN: ${{ secrets.FONTAWESOME_NPM_AUTH_TOKEN }}
       GEM_HOST_API_KEY: ${{ secrets.RUBYGEMS_API_KEY }}
       NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
     steps:
       - uses: actions/checkout@v3
+      - name: 'Enable registry auth'
+        shell: bash
+        run: |
+          echo "_auth= ${{ secrets.POWERHOME_NPM_REGISTRY_TOKEN }}" >> ~/.npmrc
```
ryanmrodriguez, line 23: Shouldn't we scope this to npm.powerapp.cloud?

```
registry=https://npm.powerapp.cloud
//npm.powerapp.cloud/:always-auth=true
//npm.powerapp.cloud/:_auth=$REDACTED
````

**Comment by dhhuynh2** [comment]
Thought I'd post a summary of where I think this PR is at, since this turned out to be a little more complicated than I originally thought. Also included ideas of how we can proceed:

### Context 
Playbook [uses fontawesome-pro](https://github.com/powerhome/playbook/blob/66d2c75c97f4b815a47833d6e737eaadc922ad7f/playbook-website/package.json#L11), which we would like to use our new npm-registry to cache. It is utilized in playbook-website.

### What this PR does 
Removes all references to the old fontawesome secrets. Switches registry to use new npm-registry instead of npmjs. Keeps publishing workflow the same.

### Problem 
The PR would solve the original problem of switching us over to cache fontawesome, but as playbook is a public repo, people outside Power would be unable to run yarn.

### Solution A

Migrate playbook-website over to a new repository. In the new repo, we could utilize our new registry, and in this current repository, we would leave it using npmjs/yarn.

Since fontawesome-pro would be gone in this scenario, we could even remove the fontawesome secrets in this repo, which would enable non-Power developers to run yarn.

This seems like the ideal solution, as it resolves an existing problem (enabling non-Power devs to work on the repo), while fixing our fontawesome issue. Problem with this is that per @jasperfurniss, playbook seems to reference playbook-website, meaning that this would be more complicated than it appears.

### Solution B

Fix Github Actions by adding in token per https://github.com/powerhome/npm-registry/pull/12, afterwards merge this PR.

This would mean that non-Power devs would not be able to run yarn, as the registry is internal. However, they are currently unable to do this because of secrets. This would essentially just maintain the status quo while allowing us to cache fontawesome packages. In the long-term, we could then explore Solution A.

I think B makes sense for now in the interim, then potentially moving playbook-website out would be the long term solution.

CC @jasperfurniss @ryanmrodriguez

**Comment by benlangfeld** [comment]
Let's go forward with option B.

File: .github/workflows/github-actions-alpha-version.yml [hunk 3]
```diff
@@ -13,15 +13,18 @@ jobs:
...
       BRANCH_NAME: ${{ github.head_ref }}
-      FONTAWESOME_NPM_AUTH_TOKEN: ${{ secrets.FONTAWESOME_NPM_AUTH_TOKEN }}
       GEM_HOST_API_KEY: ${{ secrets.RUBYGEMS_API_KEY }}
       NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
     steps:
       - uses: actions/checkout@v3
+      - name: 'Enable registry auth'
+        shell: bash
+        run: |
+          echo "//npm.powerapp.cloud/:_auth="$(echo -n 'gh-actions:${{ secrets.POWERHOME_NPM_REGISTRY_PASSWORD }}' | base64)"" >> ~/.npmrc
```
benlangfeld, line 23: We must still allow auth with the public registry for publishing.

```suggestion
          echo "//registry.npmjs.org/:_authToken="$(echo -n '${{ secrets.NPM_TOKEN }}')"" >> ~/.npmrc
          echo "//npm.powerapp.cloud/:_auth="$(echo -n 'gh-actions:${{ secrets.POWERHOME_NPM_REGISTRY_PASSWORD }}' | base64)"" >> ~/.npmrc
```

jasperfurniss (approved):

PR #2723, opened 2023-09-01
[PBNTR-85] Col and Colgroup table kits

markdoeswork:
Story: https://nitro.powerhrg.com/runway/backlog_items/PBNTR-85

adds `thead` and `col` tags for our background kit so we can highlight table columns

screenshot-127 0 0 1_3000-2023 09 13-10_00_34

1. Go to /kits/table/rails#table-column-subkit https://pr2723.playbook.beta.gm.powerapp.cloud/kits/table/rails#table-column-subkit
2. See The new kit

#### Checklist:


playbook/colors.yml
```diff
@@ -0,0 +1,62 @@
+background_colors:
```
markdoeswork, line 1: this file might be better placed

Israel-Molestina (dismissed):
LGTM!

playbook/app/pb_kits/playbook/pb_table/table_col.html.erb
```diff
@@ -0,0 +1,9 @@
+<%= content_tag(:col,
+    span: object.span,
+    aria: object.aria,
+    class: object.classname,
+    data: object.data,
+    style: "background-color: #{object.background_color};",
```
nidaqg, line 6: PR looks great @markdoeswork ! Just had a question on this piece, any particular reason we're having to use a style tag here instead of applying the color via scss? I do see that you're adding classnames based on the chosen background color in the .rb file so wanted to know what the reasoning was here

cc: @jasperfurniss

playbook/colors.yml
nidaqg: Would make sense to have this file more towards the root of the playbook folder. We do have a types folder where we store files with all the TS types for tokens/props. We could use the same folder or start a different one for rails? What do you think? I like the idea of what you're doing here!

playbook/app/pb_kits/playbook/pb_table/docs/_table_alignment_column.md
```diff
@@ -1,2 +1,2 @@
-You can individually align a pieace of table data, but a more practical use would be applied to align a column.
+You can individually align a piece of table data, but a more practical use would be applied to align a column.
```
markdoeswork, line 1: spelling mistake

playbook/colors.yml
markdoeswork: @nidaqg I would make a "labels" or "tokens" dir in the rails root and then add these yml files to it. We could also have a `sizes.yml` I see things like this all the time `%w[xxs xs sm md base lg xl]`

nidaqg: I like the idea of having more yml files for types just like we do for TS. It'll definitely clean up a lot of our .rb files!

I'll defer to @jasperfurniss on where he thinks these files should live, we need to make sure it makes sense for file structure as a whole

playbook/app/pb_kits/playbook/pb_table/table_colgroup.rb
```diff
@@ -0,0 +1,11 @@
...
+
+module Playbook
+  module PbTable
+    class TableColgroup < Playbook::KitBase
+      def classname
+        generate_classname("pb_table_colgroup_kit")
+      end
+    end
+  end
+end
```
jasoncypret, line 11: Whats the point of making this a kit...It doesn't do anything?

playbook/app/pb_kits/playbook/pb_table/table_col.html.erb
jasoncypret: ++

playbook/colors.yml
markdoeswork: This has been removed from this PR as its out of scope

playbook/app/pb_kits/playbook/pb_table/table_col.html.erb
markdoeswork: background color is being handled by a method in the controller `def classname` that follows the same class naming convention as our background kit

playbook/app/pb_kits/playbook/pb_table/table_colgroup.rb
markdoeswork: has been removed

jasperfurniss (approved):

PR #2756, opened 2023-09-21
[PLAY-845]  All Kits by type

nickamantia:
[PLAY-845](https://nitro.powerhrg.com/runway/backlog_items/PLAY-845)

This PR:
Adds typings to the `menu.yml` file.
Adds new swift section ont he all ktis page. This is local env only.

Screenshot 2023-09-22 at 10 57 18 AM

1. Go to `/kits` and check to see if all type sections have the correct kits.

#### Checklist:


playbook-website/lib/pb_doc_helper.rb
```diff
@@ -27,33 +27,58 @@ def nav_hash_array(link)
...
-    def get_kits
+    def get_kits(type = "rails")
       menu = YAML.load_file(Playbook::Engine.root.join("dist/menu.yml"))
-      menu["kits"]
+      menu["kits"][type]
     end
 
-    def get_kits_pb_website
-      menu = YAML.load_file(Rails.root.join("config/menu.yml"))
-      menu["kits"]
```
nickamantia, line 54: Is thais okay to remove? i believe that our new `aggregate_kits` does what this this method did. Also, it doesn't seem like we call this method anywhere.

jasperfurniss: Yes. You can remove it! Good catch.

playbook-website/lib/pb_doc_helper.rb
```diff
@@ -27,33 +27,58 @@ def nav_hash_array(link)
...
+    def aggregate_kits
+      menu = YAML.load_file(Playbook::Engine.root.join("dist/menu.yml"))
```
jasoncypret, line 51: @jasperfurniss  Any idea why we are not using this? Might be the wrong path but this is called once and available everywhere. For some reason we stopped referencing this. I'll bet this would save some server processing if we wen't back to it.

https://github.com/powerhome/playbook/blob/master/playbook-website/config/initializers/global_variables.rb#L3-L4

playbook-website/lib/pb_doc_helper.rb
```diff
@@ -27,33 +27,58 @@ def nav_hash_array(link)
 
     # Deal with lists of kits, used in Playbook doc and Externally
     # rubocop:disable Style/StringConcatenation
-    def pb_kits(type: "rails", limit_examples: false, dark_mode: false, method: get_kits)
```
jasoncypret, line 30: Isn't `method` used by the docs in Nitio? If so I suggest testing with an alpha.

playbook-website/app/helpers/application_helper.rb
```diff
@@ -137,7 +137,7 @@ def format_search_hash(kit)
   def search_list
     all_kits = []
     formatted_kits = []
-    MENU["kits"].each do |kit|
```
jasoncypret, line 140: `MENU["kits"]` is cached. Why are we doing this?

jasoncypret (changes requested):

playbook-website/lib/pb_doc_helper.rb
jasperfurniss: @nickamantia  Yeah let's go ahead and reference those cached versions that @jasoncypret pointed out from the global_variables.rb. Let me know if you want syntax help change these lines.

playbook-website/lib/pb_doc_helper.rb
jasperfurniss: Yes. @nickamantia We'll want to test the 2nd & 4th search results[ here](https://github.com/search?q=repo%3Apowerhome%2Fnitro-web%20pb_kits&type=code) . Those (the dev docs page) depend on this method.

jasperfurniss: @jasoncypret Clarification: we are only removing the get_kits method that is being called on the pb_kits method here. Nick is testing the alpha now, but it shouldn't be affected.

playbook-website/app/helpers/application_helper.rb
jasperfurniss: @jasoncypret Clarification: Actually Nick is pushing up a commit momentarily that will utilize the MENU (cached one) under the hood of our aggregate_kits method. We did need to create a new method that does the correct grouping to support the sidebar use cases for rails, react, and swift since the structure of the menu.yml is now different. But again, inside of that new method, we will reference the cached version. So check again soon on this one!

playbook-website/config/menu.yml
```diff
@@ -1,111 +1,236 @@
...
-    - person
-    - person_contact
-    - source
-    - stat_change
-    - stat_value
-    - title_count
-    - title_detail
-  - user_badge
-  - walkthrough
+  rails:
```
jasoncypret, line 2: [What about a different structure to indicate platforms](https://gist.github.com/jasoncypret/9f99cf83f2d508d4744612b5bea38c05)? I'm afraid this method it will be very hard to keep the categories in sync.

playbook-website/app/controllers/pages_controller.rb
```diff
@@ -5,6 +5,7 @@
 require "will_paginate/array"
 
 class PagesController < ApplicationController
+  include PlaybookWebsite::PbDocHelper
```
jasoncypret, line 8: Should this be included in the application controller instead? https://github.com/powerhome/playbook/blob/master/playbook-website/app/controllers/application_controller.rb

playbook-website/config/menu.yml
jasperfurniss: I explored something very similar with him already...
kits:
  - avatars:
      platforms: ['react', 'rails']
      components:
        - avatar
        - avatar_action_button
        - multiple_users
        - multiple_users_stacked
        - user
  - background:
      platforms: ['react']
  - bread_crumbs:
      platforms: ['rails']
  - buttons:
      platforms: ['react', 'rails']
      components:
        - button
        - button_toolbar
        - circle_icon_button
  - card:
..........

Ultimately, our current method has the following benefits, which is why we ultimately chose it:
1) clearly separated: our current approach offers a clear, top-level separation between Rails and React components, making it easy to look up components for a specific platform. (especially in his aggregate_kits method that already has been written for sorting the sidebar)
2) simplicity: It's straightforward and doesn't require nesting keys like platforms under each kit, which can be simpler to parse. (visually for our sidebar and programmatically in Nick's current method)
3) reduced redundancy: If a kit is exclusive to one platform, we don't have to explicitly label it as such—it's obvious by its placement... so one less thing to do as well.

jasperfurniss: Can you explain why this method it will be very hard to keep the categories in sync? Nick already has it working well. And both our structure and the one you proposed would use similar syntax to update and change (make new additions to) the menu.yml file with the generator.

PR #3289, opened 2024-03-26
[PBNTR-243] Global Positioning Props to Allow Inset Value

nidaqg:

Adds the option to pass in an inset to the global positioning prop to provide for edge cases like the Avatar kit

Screenshot 2024-03-27 at 8 12 19 AM

### DIFF
--- a/playbook/app/pb_kits/playbook/utilities/globalProps.ts
+++ b/playbook/app/pb_kits/playbook/utilities/globalProps.ts
@@ -119,19 +119,19 @@ type Position = {
 }
 
 type Top = {
-  top?: Sizes,
+  top?: Sizes | {value: string, inset: boolean},
 }
 
 type Right = {
-  right?: Sizes,
+  right?: Sizes | {value: string, inset: boolean},
 }
 
 type Bottom = {
-  bottom?: Sizes,
+  bottom?: Sizes | {value: string, inset: boolean},
 }
 
 type Left = {
-  left?: Sizes,
+  left?: Sizes | {value: string, inset: boolean},
 }
 
 type Shadow = {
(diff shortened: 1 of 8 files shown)


playbook/app/pb_kits/playbook/utilities/globalProps.ts
```diff
@@ -441,11 +441,54 @@ const PROP_CATEGORIES: {[key:string]: (props: {[key: string]: any}) => string} =
...
+    if (typeof left === 'string') {
+      css += `left_${left}`;
+    } else if (typeof left === 'object' && left.inset) {
+      css += `left_${left.value}_inset`;
+    }else if (typeof left === 'object' && !left.inset) {
+      css += `left_${left.value}`;
+    }
+    return css;
+  },
+
```
jasperfurniss, line 491: I know when we chatted about this, you were thinking about abstracting to a smaller function to get things a little more DRY and have a single point of truth.
 
I think you can feel that freedom in this file. 

Do you want to test this: 
```tsx

const generateCss = (prop, value) => {
  let css = "";
  if (typeof value === 'string') {
    css = `${prop}_${value}`;
  } else if (typeof value === 'object' && value.inset) {
    css = `${prop}_${value.value}_inset`;
  } else if (typeof value === 'object' && !value.inset) {
    css = `${prop}_${value.value}`;
  }
  return css;
};

const cssProps = {
  topProps: ({ top }) => generateCss('top', top),
  rightProps: ({ right }) => generateCss('right', right),
  bottomProps: ({ bottom }) => generateCss('bottom', bottom),
  leftProps: ({ left }) => generateCss('left', left),
};
```

playbook/lib/playbook/bottom.rb
```diff
@@ -24,9 +24,12 @@ def bottom_props
...
         value = send(k)
-        return nil unless bottom_values.include? value
-
-        "bottom_#{value}"
+        css = if value.is_a?(Hash) && value[:inset] && bottom_values.include?(value[:value])
+                "bottom_#{value[:value]}_inset"
+              elsif bottom_values.include?(value)
+                "bottom_#{value}"
+              end
+        css
```
thestephenmarshall, line 32: Akin to what @jasperfurniss mentioned above, I think this can be abstracted away to `playbook/position.rb` as a single method for top, left, bottom, and right.

playbook/app/pb_kits/playbook/utilities/globalProps.ts
nidaqg: Done! Slightly different from your above suggestion but same principles! @jasperfurniss

playbook/lib/playbook/bottom.rb
nidaqg: Done! Thank you for the feedback, definitely a lot better than the repetition that was there before @thestephenmarshall

jasperfurniss (approved):

thestephenmarshall (approved):

PR #3410, opened 2024-05-14
[PBNTR-306] Collapsible Kit with Custom Collapsible.Main

nidaqg:
**What does this PR do?**

[Runway Story](https://nitro.powerhrg.com/runway/backlog_items/PBNTR-306)

- ✅ The ability to pass in a custom component in place of Collapsible.Main.
- ✅ When custom Collapsible.Main used, kit does not add any of the flex/flexItem wrappers so devs have full control on structure of that Main
- ✅ Devs must use useCollapsible hook to attach toggling click event to the custom Main (if we wrap it in a div ourselves, it will prevent things like position:sticky from working which will defeat the purpose
- ✅ Ability to NOT use any Main and the kit to still work with external control
- ✅ All the above for React as well as Rails
  - ✅ Click event for rails side without needing custom JS
  - ✅ Backwards compatible: No change for current implementations of the kit.
- ✅ Make the icon button available as a subcomponent to use for custom interfaces (react and rails)
- ✅ Alpha testing for collapsibles in Nitro before shipping

**Screenshots:** 

Screenshot 2024-05-15 at 12 51 49 PM


jasperfurniss (approved):

thestephenmarshall (approved):
I think this one looks great and appears to maintain backwards compatibility.

PR #3440, opened 2024-05-29
[PLAY-1343] fix Tooltip oversize and misalignment issue

ElisaShapiro:
**What does this PR do?** A clear and concise description with your runway ticket url.
[PLAY-1343](https://nitro.powerhrg.com/runway/backlog_items/PLAY-1343) -  as of right now this PR is for making an alpha that reverts change from [PLAY-1249](https://github.com/powerhome/playbook/pull/3380) in order to truly verify that this change isn't the cause of the runway tooltip misalignment issue.

**Screenshots:** Screenshots to visualize your addition/change

**How to test?** Steps to confirm the desired behavior:
4. See addition/change

#### Checklist:
- [ ] **LABELS** Add a label: `enhancement`, `bug`, `improvement`, `new kit`, `deprecated`, or `breaking`. See [Changelog & Labels](https://github.com/powerhome/playbook/wiki/Changelog-&-Labels) for details.
- [ ] **DEPLOY** I have added the `milano` label to show I'm ready for a review.
- [ ] **TESTS** I have added test coverage to my code.

PR #3445, opened 2024-06-03
Revert "[PLAY-1343] fix Tooltip oversize and misalignment issue"

thestephenmarshall:
Reverts powerhome/playbook#3440


ElisaShapiro (approved):
Thanks for reverting!

PR #3514, opened 2024-07-09
[PLAY-1416] Add alias map entrypoint for rails

markdoeswork:
Runway https://runway.powerhrg.com/backlog_items/PLAY-1416
Nitro Alpha  https://github.com/powerhome/nitro-web/pull/41052
Alias Logic https://github.com/powerhome/playbook-icons/pull/36/files

The point of this story is to remove the alias logic from `plabyook` an put it in `plabyook-icons`
Removes [icon_aliases.json](https://github.com/powerhome/playbook/pull/3514/files#diff-6d5097953fcb84f6a2036e7ab0c0c25720cd2f64ca434A387b5dff241d01c55dd) and moves it to Playbook Icons Repo.

I added a "feature" of the rails icon kit where you could pass an alias map to it. But its not required to be used for the icon kit.

More details about the alias logic here https://github.com/powerhome/playbook-icons/pull/36


Dockerfile
```diff
@@ -90,3 +90,7 @@ COPY --link --from=release /home/app/src/playbook-website/public /home/app/src/p
 COPY --link --from=release /home/app/src/node_modules/@powerhome/playbook-icons/icons /home/app/src/temp-icons
 RUN cp -r /home/app/src/temp-icons/* /home/app/src/playbook-website/app/javascript/images/
 RUN rm -rf /home/app/src/temp-icons
+
+COPY --link --from=release /home/app/src/node_modules/@powerhome/playbook-icons/aliases.json /home/app/src/aliases.json
```
jasperfurniss, line 94: We should avoid using docker here. 
Could we just initialize the file in an initializer? 

```.rb
# config/initializers/load_json.rb

require 'json'

module_path = Rails.root.join('node_modules', 'playbook-icons', 'aliases.json')
JSON_DATA = JSON.parse(File.read(module_path))
```

playbook/app/pb_kits/playbook/pb_icon/_icon.tsx
```diff
@@ -41,19 +40,6 @@ type IconProps = {
   spin?: boolean,
 } & GlobalProps
 
-type AliasType = string | string[];
```
jasperfurniss, line 44: I'm a bit confused? Why did we remove all of the logic from the icon alias work. It still needs to live inside the playbook icon kit, even if the actual alias list lives in the `playbook-icons` repo.

jasperfurniss: I see now that a lot was moved to the playbook-icons library. I'll check that out now.

Dockerfile
jasperfurniss: And then the importing could look like this (if we need the import in a JS file)
```
// app/javascript/packs/application.js

import jsonData from 'my-module/data.json';

console.log(jsonData);
```

jasperfurniss (approved):

PR #3629, opened 2024-08-23
[PBNTR-462] Adding default value prop to Rails Star Rating kit

carloslimasd:
**What does this PR do?**
Adding default value prop to Rails Star Rating kit

image

1. Go to the Rails Star Rating kit page.

#### Checklist:


nidaqg (approved):

ElisaShapiro (commented):
Looks great and works great! 
I do have a producty suggestion:  just to make it explicitly clear that this default value is for the interactive variant and expressly added so that form/filters can function correctly, we should add some markdown stating so. It could even actually be included it in the same example as Interactive. The markdown can say something like: "Pass `default_value` to the interactive variant for form submission to function. If you do not want to suggest an initial rating to the user, `default_value=0` works."
I worry that not explicitly stating this will cause nitro devs to encounter the same issue we ran into last week - hopefully this can help head off any questions or issues down the line.
(Additionally - the two interactive star rating kits on the [Form page's doc examples](https://playbook.powerapp.cloud/kits/form) should probably have default_value set in them as well because of this).

PR #3685, opened 2024-09-11
[PBNTR-480] Implementing Radio custom children prop

carloslimasd:
**What does this PR do?**
Implement radio custom children prop for use kits instead of a text.

**Screenshots:**
image

1. Go to the Radio kit doc page.
2. Scroll down to the Custom Children doc example.

#### Checklist:


jasperfurniss (dismissed):

playbook/app/pb_kits/playbook/pb_radio/radio.html.erb
```diff
@@ -1,18 +1,42 @@
...
-    id: object.id,
-    value: object.value,
-    **combined_html_options) do %>
-
-    <% if content.present? %>
-      <%= content %>
-    <% else %>
-      <%= radio_button_tag object.name, object.value, object.selected, object.input_options %>
+<% if object.custom_children %>
+  <div data-pb-radio-children="true">
```
nidaqg, line 2: Everything works correctly but a few things: if we use custom children, the top level div now is this, which means the wrapping div is no longer 'pb_radio_kit' . Top level div in our kits should have that classname, we do do this in one other place i believe, the CircleChart, where we add a wrapper. If you want to NOT put pb_radio_kit on the top level div, it should at least have a classname to the effect of pb_radio_kit_wrapper, does that make sense?

I'm assuming moving the pb_radio_kit classname up was effecting the way scss was applying here? Was that why we put the pb_radio_kit class inside the flex?

playbook/app/pb_kits/playbook/pb_radio/radio.html.erb
```diff
@@ -1,18 +1,42 @@
...
+      cursor: "pointer",
+      **combined_html_options
+    }) do %>
+      <%= content_tag(:label,
+        checked: object.checked,
+        class: object.classname,
+        data: object.data,
+        id: object.id,
+        value: object.value,
+        **combined_html_options) do %>
```
nidaqg, line 16: Do we want to keep the data, aria and html_options props on the label and not the top level div? With this set up applying global props to this kit is applying the global prop only to the input, not the entire kit which is not what we want. We want global props to apply to the top level div for the kit.
Screenshot 2024-09-16 at 11 57 09 AM

playbook/app/pb_kits/playbook/pb_radio/radio.html.erb
carloslimasd: @nidaqg I didn't want to change much because I only had to add a div because the Flex kit does not accept data attributes. If we fix this in the future, we won't need this div anymore.

I think passing an empty class is weird since it wouldn't apply any CSS.

I suggest passing the data attribute to the label content tag. This way, we could remove the div and not use an empty class, which would solve your second comment as well (passing global props to the higher wrapper).

What do you think?

nidaqg: I'm good with trying that @carloslimasd ! We want to make sure global props are applying to the right div basically

nidaqg (approved):

PR #3767, opened 2024-10-04
More storage for CI image builds

benlangfeld:
https://ci.powerapp.cloud/blue/organizations/jenkins/powerhome%2Fplaybook/detail/master/996/pipeline/

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -26,7 +26,7 @@ app.build(
             ]
         ],
         storageConfig: [
-          size: '10Gi',
+          size: '15Gi',
         ],
       ])
     }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -25,6 +25,9 @@ app.build(
                 memory: '10Gi',
             ]
         ],
+        storageConfig: [
+          size: '10Gi',
+        ],
       ])
     }
 


jasperfurniss (dismissed):

Jenkinsfile
```diff
@@ -25,6 +25,9 @@ app.build(
                 memory: '10Gi',
             ]
         ],
+        storageConfig: [
+          size: '10Gi',
```
mfilipe, line 29: It is already 10Gi ([default value in pac-buildkit](https://github.com/powerhome/pac-buildkit/blob/bc114c6cb82291ab400b880e250313330f2d9e03/operator/constants.py#L18-L20)), so we should increase this value.

TeamTeaTime, line 29:
```suggestion
          size: '15Gi',
```

default is already 10Gi. Before we merge this we'll want to clean up the old caches

benlangfeld: The docs say default is 5Gi.

How can I find utilisation metrics, for CPU and memory also? Inline is easy, remote builder is hidden and affecting other projects like NitroID.

benlangfeld: Actually, default seems to be 10Mi. Typo?

https://github.com/powerhome/pac-buildkit/blob/main/charts/operator/templates/configmap.yaml#L16-L18

TeamTeaTime: The 10Mi ephemeral storage limit so far has only come into play on the multi-arch builds. This is using [this default](https://github.com/powerhome/pac-buildkit/blob/main/charts/operator/templates/configmap.yaml#L9) in question

benlangfeld:
```
#39 [release 4/6] RUN --mount=id=playbook-yarncache,type=cache,target=/home/app/.cache/yarn,uid=9999,gid=9999,sharing=locked     cd playbook; NODE_ENV=production; yarn release

#39 ERROR: open /var/lib/buildkit/runc-overlayfs/snapshots/snapshots/124/fs/usr/local/share/.cache/yarn/v6/npm-prosemirror-model-1.19.0-d7ad9a65ada0bb12196f64fe0dd4fc392c841c29-integrity/node_modules/prosemirror-model/src/index.ts: no space left on device
```

Is this ephemeral storage exhaustion, or the cache volume?

TeamTeaTime: As for utilization metrics, it's on the todo list to update the links to also include remote metrics, but you can hunt it down via the runner pod name: https://metrics.powerapp.cloud/d/s9c8D_K7k/pod-resource-usage?orgId=1&refresh=10s&var-datasource=000000003&var-namespace=pac-buildkit&var-filter=.%2B&var-pod=buildkit-runner-mssns-tydiw&from=now-1h&to=now

nidaqg (approved):

PR #3782, opened 2024-10-09
Fix staging and production deployments

indiebrain:
Deployments to non-pr environments are broken.

- https://milano.powerapp.cloud/powerhome/playbook/staging/deploys/9880200#

```
key not found: "open_ai"
```

First, https://github.com/powerhome/playbook/pull/3776 only added
values for the `open_ai` value to the `prs` environment - IE review
stacks. The project still needs valaues for staging and production.

Second, even if these values were present, the env var name the
application is looking for and the env var name used by the deployment
are different.

https://github.com/powerhome/playbook/blob/ddf4db1da92383ac58e8c6f17669a0d38629c7c2/playbook-website/config/initializers/openai.rb#L3


indiebrain:
@jasperfurniss , staging and production `secrets.yaml` also need the `open_ai` value set in order to unblock the deployment pipeline on `main`. Please feel free to add them here.

jasperfurniss (dismissed):

jasperfurniss:
Thanks @indiebrain - makes sense!

jasperfurniss (approved):

PR #3807, opened 2024-10-16
[PLAY-1485] Fix overflow container bug by using padding instead of outline

kangaree:
**What does this PR do?**

In [PLAY-829](https://runway.powerhrg.com/backlog_items/PLAY-829), we removed the buffer to fix content alignment issues and added an outline so the content would not 'jump' (which was handled originally by the buffer).

- Fix overflow container bug by using padding instead of outline
- Subtract 1px of padding in the selected label container to account for the extra 1px border of selected label
- Do this for all padding sizes xxs to xl
- For "selectable cards", keep an outline as the content is different (a padding 0 div wrapped around flex and cards)
- Fix the dark mode for rails and apply it to the card

Screenshot 2024-10-30 at 10 25 07 AM

Screenshot 2024-10-23 at 9 15 54 AM

1. Go to the Playbook Website
2. Go to the Selectable Card kit
3. Click the cards
4. Make sure the content does not "jump" or shift around and stays in place
5. Go to the Nitro test env
6. Use the Impersonation Flyout
7. Search by title so multiple results show (e.g., "Nitro Quality Ninja")
8. The selected card should not be cut off
9. If you select a card, the content should not "jump" or shift around
10. Since there are multiple employees, everything should fit in the overlay- the impersonate and cancel button should fit on the overlay and not be below a scroll.

#### Checklist:


playbook/app/pb_kits/playbook/pb_selectable_card/selectable_card.rb
```diff
@@ -25,7 +25,7 @@ class SelectableCard < Playbook::KitBase
 
       def classname
         [
-          generate_classname_without_spacing("pb_selectable_card_kit", checked_class, enable_disabled_class),
+          generate_classname_without_spacing("pb_selectable_card_kit", checked_class, enable_disabled_class, display_input_class),
```
jasperfurniss, line 28: As a rule of thumb, we don't change the default classname on a kit/variant... because it can affect custom css overrides in nitro-web, especially kits that are a part of forms. Example: https://github.com/search?q=repo%3Apowerhome%2Fnitro-web%20pb_selectable_card_kit_checked_enabled&type=code



Here's the way i "might" handle it (test and see, and look at other kit .rb files): 

```suggestion
          generate_classname_without_spacing("pb_selectable_card_kit", checked_class, enable_disabled_class) + display_input_class,
```

kangaree: I've changed it from:
`pb_selectable_card_kit_checked_enabled_display_input`
to
`pb_selectable_card_kit_checked_enabled display_input`

nidaqg (approved):

PR #3874, opened 2024-11-01
[PLAY-1603] Add color hover docs and fix background hover

markdoeswork:
runway https://runway.powerhrg.com/backlog_items/PLAY-1603

Add a "Text Color" hover prop option is the ask in this story 

because color is already a hover option i just used that. The Background and color hover props weren't working because a lot of our kits are listening for classnames that are "_error", "_success" or "_#{any_color_token}"

so i changed the classname to not trigger the pre existing css by making the syntax "hover_color-#{any_color_token}". Doing this to the background kit fixes it too

Checkout the page here https://pr3874.playbook.beta.gm.powerapp.cloud/visual_guidelines/hover

screenshot-127_0_0_1_3000-2024_11_01-09_29_40

### DIFF
--- a/playbook/app/pb_kits/playbook/utilities/_hover.scss
+++ b/playbook/app/pb_kits/playbook/utilities/_hover.scss
@@ -2,11 +2,11 @@
 
 @mixin hover-color-classes($colors-list) {
     @each $name, $color in $colors-list {
-      .hover_background_#{"" + $name}:hover {
+      .hover_background-#{"" + $name}:hover {
         background-color: $color !important;
         transition: background-color $transition-speed ease;
       }
-      .hover_color_#{"" + $name}:hover {
+      .hover_color-#{"" + $name}:hover {
         color: $color !important;
         transition: color $transition-speed ease;
       }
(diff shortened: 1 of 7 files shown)


skduncan (dismissed):
lgtm

playbook/app/pb_kits/playbook/utilities/_hover.scss
```diff
@@ -2,11 +2,11 @@
 
 @mixin hover-color-classes($colors-list) {
     @each $name, $color in $colors-list {
-      .hover_background_#{"" + $name}:hover {
+      .hover_background-#{"" + $name}:hover {
```
jasperfurniss, line 5: Is there a reason we changed the underscore for all those separators on these?

I'd prefer to stick to what we had, unless there's an issue you encountered with some scss bleed occurring. 

Preference:
```suggestion
      .hover_background_#{"" + $name}:hover {
```

It looks like you changed it to a `-` instead of underscore for background and color:

.hover_background-#{"" + $name}:hover {
.hover_color-#{"" + $name}:hover {

But, you left it unchanged for scale and shadow. 
If possible, we want color & background to match the way we have it for scale and shadow, with _all_  underscores for consistency. 

Thanks dude!

markdoeswork: @jasperfurniss 

The card kit has css to that "listens" for our color tokens and changes the background of the kit if they are present.

Using a different pattern, in this case `-` avoids that query 

this behavior in our card kit is the same behavior in our typography kits  which is why i use the same patter for hover

Screenshot 2024-11-13 at 2 06 14 PM

skduncan (approved):
lgtm

jasperfurniss (approved):

PR #3893, opened 2024-11-07
[PBNTR-667] Rails Typeahead Form Integration

ElisaShapiro:
[PBNTR-667](https://runway.powerhrg.com/backlog_items/PBNTR-667) addresses missing functionality from the playbook typeahead used within forms via the `form.typeahead` or `f.typeahead` construction. Specifically, this ticket is in response to an SLA request for accessible label, validation, and reset functionality. 

The results of my investigation were the following: 
- Label click-to-focus functionality was not present in the kit as currently structured. The added label functionality utilizes the recommended Rails form helper `@template.label(@object_name, name)`. It depends on a label's for property matching the text input's id.
- Validation was not present in the kit as currently structured. Validation now works in two different ways. A boolean prop `required: true` taps into the browser's default validation properties. A user can also use the additional validation prop (in conjunction with `required: true` to add a custom validation message.
- The resetting issue I believe is just a matter of implementation because the `is_multi: false` in the provided code snippet turns it a "react-rendered-rails" typeahead, not a "pure rails" one, and the reset functionality does not work as expected in the form context.

I also cleaned up the [form kit page](https://playbook.powerapp.cloud/kits/form) some (having distinct labels between doc examples so that all label-clicks do not redirect to the "default' doc example). I would suggest a follow up story to add a "Form + Typeahead" doc example or clarifying text in the description on the Typeahead kit page as well to avoid the react-rendered-rails vs. rails issue.

Clicking the label puts the focus on the input (see blue highlight around input box and cursor in input)
label focus
Adding `required: true` provides validation (using the browser's generic "Please fill out this field"
for PR typeahead required validation
Using a validation prop with a custom message will change the validation message (used along with `required: true`).
for PR new custom validation error

1. Go to [form kit page](https://pr3893.playbook.beta.gm.powerapp.cloud/kits/form). A typeahead has been added as the first entry in the Default + Validation doc example.
2. When you click the "example typeahead" label, the focus should redirect to the typeahead.
3. When you've selected a user in the typeahead, scroll down to the bottom of the doc example and hit the "cancel" button. The form field for typeahead should reset (just like the rest of them do). 
4. Without anything entered in the typeahead field, try to submit the form by pressing the "Save Example" button. The red error indicator should appear for the typeahead example (just like it does for the rest of them).
5. Clicking the labels for any other option in the Default + Validation doc ex should direct focus to the field directly below it, NOT to the corresponding field in the Default doc example like it does currently in [production](https://playbook.powerapp.cloud/kits/form#default-validation). 

#### Checklist:
~- [ ] **TESTS** I have added test coverage to my code.~

### DIFF
--- a/playbook/app/pb_kits/playbook/pb_typeahead/typeahead.rb
+++ b/playbook/app/pb_kits/playbook/pb_typeahead/typeahead.rb
@@ -40,6 +40,8 @@ module Playbook
       prop :pill_color, type: Playbook::Props::Enum,
                         values: %w[primary neutral success warning error info data_1 data_2 data_3 data_4 data_5 data_6 data_7 data_8 windows siding roofing doors gutters solar insulation accessories],
                         default: "primary"
+      prop :required, type: Playbook::Props::Boolean,
+                      default: false
 
       def classname
         default_margin_bottom = margin_bottom.present? ? "" : " mb_sm"
--- a/playbook/app/pb_kits/playbook/pb_typeahead/typeahead.rb
+++ b/playbook/app/pb_kits/playbook/pb_typeahead/typeahead.rb
@@ -42,6 +42,8 @@ module Playbook
                         default: "primary"
       prop :required, type: Playbook::Props::Boolean,
                       default: false
+      prop :validation, type: Playbook::Props::HashProp,
+                        default: {}
 
       def classname
         default_margin_bottom = margin_bottom.present? ? "" : " mb_sm"
--- a/playbook/lib/playbook/forms/builder/typeahead_field.rb
+++ b/playbook/lib/playbook/forms/builder/typeahead_field.rb
@@ -5,6 +5,7 @@ module Playbook
     class Builder
       def typeahead(name, _options = {}, _html_options = {}, props: {})
         props[:name] = name
+        props[:label] = props[:label] == true ? name.to_s.underscore.humanize : props[:label]
         @template.pb_rails("typeahead", props: props)
       end
     end
--- a/playbook/lib/playbook/forms/builder/typeahead_field.rb
+++ b/playbook/lib/playbook/forms/builder/typeahead_field.rb
@@ -4,15 +4,38 @@ module Playbook
   module Forms
     class Builder
       def typeahead(name, _options = {}, _html_options = {}, props: {})
+        # props[:name] = name
+        # props[:label] = props[:label] == true ? name.to_s.underscore.humanize : props[:label]
+        # props[:label] = @template.label(@object_name, name) if props[:label] == true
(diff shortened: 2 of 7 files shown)


playbook/lib/playbook/forms/builder/typeahead_field.rb
```diff
@@ -5,6 +5,7 @@ module Forms
     class Builder
       def typeahead(name, _options = {}, _html_options = {}, props: {})
         props[:name] = name
+        props[:label] = props[:label] == true ? name.to_s.underscore.humanize : props[:label]
```
terryfinn, line 8: If we did something similar to this:

https://github.com/powerhome/playbook/blob/244697cbb9c2de700a3fc7422723415cec66f73c/playbook/lib/playbook/forms/builder/form_field_builder.rb#L9

Then we would get the benefit of Rails inflections and translations for the files name. That is a lot of features that can greatly improve the usability of this kit.

playbook/app/pb_kits/playbook/pb_typeahead/typeahead.rb
```diff
@@ -40,6 +40,8 @@ class Typeahead < Playbook::KitBase
       prop :pill_color, type: Playbook::Props::Enum,
                         values: %w[primary neutral success warning error info data_1 data_2 data_3 data_4 data_5 data_6 data_7 data_8 windows siding roofing doors gutters solar insulation accessories],
                         default: "primary"
+      prop :required, type: Playbook::Props::Boolean,
```
terryfinn, line 43: Does this allow both Boolean values and arbitrary strings? I think we would need to support both. True for when you want rails to use the attribute name / translations to generate a string, false when for whatever reason you do not want a label, and a string for when you have a specific value you would like to use for the label.

playbook/lib/playbook/forms/builder/typeahead_field.rb
ElisaShapiro: I had tried to get this construction to work for a while but was not able to get label click-to-focus functionality to work with typeahead I think because of the nature of the embedded/under the kit input differing from how other kits have it. Can definitely give it another shot (may require a bit more kit refactoring).

playbook/app/pb_kits/playbook/pb_typeahead/typeahead.rb
ElisaShapiro: I believe only the text input and text area kits (all the _field variations) support validation strings at this time. Going to add validation prop here and suggest adding that to other kits as well in follow up.

nidaqg (approved):

jasperfurniss (approved):

PR #3910, opened 2024-11-14
[PLAY-1648] Height Global Props

markdoeswork:
Runway https://runway.powerhrg.com/backlog_items/PLAY-1648

I added tokens and doc pages for the height, min height and max height global props

screenshot-localhost_3000-2024_11_15-14_32_14


jasperfurniss:
This looks great! 

I did find an issue though that I'd love for you to resolve before we merge this work!
Screenshot 2024-11-19 at 9 00 05 AM

I tested with the Card, which is one of the kits i did with my dynamic height prop in a previous story. 

We need to add logic so that when the user/dev adds a height that is NOT one of our tokens, the kit should NOT add a classname that gets thrown on:

Screenshot 2024-11-19 at 9 00 16 AM

Otherwise, we'll just have random classnames on there.

My dynamic height is still working correctly though with your tokens as well-- it's just that extraneous classname should not be added.

playbook/lib/playbook/kit_base.rb
```diff
@@ -172,8 +172,17 @@ def data_attributes
     end
 
     def dynamic_inline_props
-      styles = global_inline_props.map { |key, value| "#{key.to_s.gsub('_', '-')}: #{value}" if value.present? }.compact
+      styles = global_inline_props.map { |key, value| "#{key.to_s.gsub('_', '-')}: #{value}" if inline_validator(key, value) }.compact
       styles.join("; ").presence
     end
+
+    def inline_validator(key, value)
```
markdoeswork, line 179: @jasperfurniss i added this method to conditionally skip

jasperfurniss (approved):

PR #4123, opened 2025-01-14
Scss fixes for Responsive Scroll for Table

nidaqg:
Testing a fix for for the templates for [PBNTR-778](https://runway.powerhrg.com/backlog_items/PBNTR-778)

This PR:
- Removes display block from responsive scroll class as it was breaking the table when it did not have enough columns to take up full width
- Makes fixes so that box-shadow on table-responsive-scroll does not appear if the container=false prop is used


skduncan (approved):
lgtm

ElisaShapiro (dismissed):

nickamantia (dismissed):

jasperfurniss (approved):

PR #4135, opened 2025-01-17
[PBNTR-768] Adding sticky right column and refactoring PBTable enhanced element

carloslimasd:
**What does this PR do?**
Adding sticky right column and refactoring PBTable enhanced element.

**What was done:**
1. Using the class selector (".table-responsive-collapse"), we weren't running the PBTable Enhanced Element for all the tables. What made this work for the other tables was the "document.querySelectorAll".
2. Using "document.querySelectorAll" caused unexpected behaviors (as mentioned above) and performance issues. For each table, we would run the same logic for ALL the tables on the page. This is an On2 code. If we had 10 tables on the page, we would have run it 100 times.
3. I removed the use of ref functions and binds since using arrow functions, we don't need it.
4. The React version was executing PBTable Enhanced Element causing all the issues mentioned above. This code was added 5 years ago so we could add the data atttribute "data-title" to cells. I removed this connection and I am proposing on "Next steps" we refactor the way we do it.

1. Go to Table doc page for Rails and React

#### Checklist:


nidaqg (approved):
Awesome work here @carloslimasd !!!

PR #4193, opened 2025-01-29
[PLAY-1865] React Date Picker Reinitialization with Other Inputs Bug 

kangaree:
**What does this PR do?**
- Add empty dependency array to React Date Picker kit so flatpickr init runs only on first render.
- If a form had other inputs and date picker(s), any update to the parent form's state would trigger a reinitialization to the date picker(s). Now date pickers should initialize only once.

1. Go to a page with a React form that has the date picker and other inputs in Safari
2. Type and make sure the React Date Picker does not "restart" on every event / slows down other inputs.

#### Checklist:

### DIFF
--- a/playbook/app/pb_kits/playbook/pb_date_picker/_date_picker.tsx
+++ b/playbook/app/pb_kits/playbook/pb_date_picker/_date_picker.tsx
@@ -134,7 +134,7 @@ useEffect(() => {
     yearRange,
     required: false,
   }, scrollContainer)
-})
+}, [])
   const filteredProps = {...props}
   if (filteredProps.marginBottom === undefined) {
     filteredProps.marginBottom = "sm"


nidaqg (approved):

nidaqg:
This is great @kangaree ! This only effects the React side, correct? When we alpha test this, lets make sure we test locations that use this kit inside ReactHookForms, ReactHookForms do depend on some re-render logic, should be fine, but we should test just in case! I would suggest the tenure analysis page as a high traffic page to test!

kangaree:
> This is great @kangaree ! This only effects the React side, correct? When we alpha test this, lets make sure we test locations that use this kit inside ReactHookForms, ReactHookForms do depend on some re-render logic, should be fine, but we should test just in case! I would suggest the tenure analysis page as a high traffic page to test!

Correct- it only affects the React side!

Thanks! I will make sure we test the tenure analysis page as well as any other date picker kits inside ReactHookForms.

kangaree:
Closing because we're going to work on this in https://runway.powerhrg.com/backlog_items/PLAY-1888

PR #4201, opened 2025-01-30
[PLAY-1866] Mask Use PBenahcnedElement

markdoeswork:
Runway [PLAY-1866](https://runway.powerhrg.com/backlog_items/PLAY-1866)
Alpha https://github.com/powerhome/nitro-web/pull/45790

The input mask isn't working in Nitro because Nitro wasn't expecting the JS module to be the new pattern Mark made in

I changed it to use our PbEnhancedElement kit

### DIFF
--- a/playbook/app/pb_kits/playbook/pb_text_input/docs/_text_input_mask.html.erb
+++ b/playbook/app/pb_kits/playbook/pb_text_input/docs/_text_input_mask.html.erb
@@ -38,7 +38,7 @@
     margin_bottom: "md",
     name: "currency_name",
     id: "example-currency",
-    placeholder: "$0.00",
+    value: "$99.99",
   }) %>
 
 <style>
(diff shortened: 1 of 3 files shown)


kangaree:
Why don't we use the `playbook/app/pb_kits/playbook/pb_text_input/inputMask.ts` file so we can keep the formatting in one place?

For example:
```js
import { INPUTMASKS } from './inputMask'

formattedValue = INPUTMASKS.currency.format(rawValue)
```

nidaqg:
@markdoeswork is there a location on that alpha where we can tests if the input mask is working as expected? If not, could you create an example for testing?

Additional question: Does this account for the usecase where a default value is used for the text input? On the React side we had the issue where if a default value was there the mask did not apply when the page loaded + when you interacted with the kit, the mask was incorrect. Can we test for this as well?

markdoeswork:
@nidaqg its not being used any where in nitro yet so a place to test would be the "all doc examples page"
https://pr45790.nitro-web.beta.gm.powerapp.cloud/dev_docs/playbook

Not sure about the default value, I'll check it out now

PR #4303, opened 2025-02-21
[PBNTR-500] Currency Kit: Handling of `null` values

nickamantia:
This PR adds a `nullDisplay` and `null_display` prop. The user can now choose how they want to display null value.

Screenshot 2025-02-24 at 11 33 21 AM

1. Go to Currency kit
2. Scroll to last doc example

#### Checklist:


playbook/app/pb_kits/playbook/pb_currency/currency.html.erb
```diff
@@ -1,17 +1,24 @@
...
+      <% else %>
+        <div><%= object.negative_sign %><%= object.symbol %></div>
+        <div><%= object.title_props[:text] %></div>
+        <div><%= object.body_props[:text] %></div>
+      <% end %>
     <% else %>
-      <%= pb_rails("body", props: object.currency_wrapper_props) do %>
-        <%= object.negative_sign %><%= object.symbol %>
+     <% if object.null_display && object.amount.blank? %>
+        <%= pb_rails("title", props: {text: object.null_display}) %>
```
nidaqg, line 14: Using title kit here is causing the null value to always be displayed as a title, which may not be what we want!

In addition to this, do you think we need to account for emphasized variant, light and bold variants? If they are using this null as a value within a table, my assumption is they would want consistency in the display, so if the rest of the values are light, they'll want this to be light? 

We can also always only account for whatever the default is for this kit and add follow ups to account for different variants. I would also reach out to a designer with these questions!

nidaqg:
General question @nickamantia , do you think we need to account for emphasized, light and bold variants? If they are using this null as a value within a table, my assumption is they would want consistency in the display, so if the rest of the values are light, they'll want this to be light as well? 

We can also always only account for whatever the default is for this kit and add follow ups to account for different variants. I would also reach out to a designer with these questions!

playbook/app/pb_kits/playbook/pb_currency/currency.html.erb
nickamantia: @nidaqg great call!! What should we have here instead?

nickamantia:
> General question @nickamantia , do you think we need to account for emphasized, light and bold variants? If they are using this null as a value within a table, my assumption is they would want consistency in the display, so if the rest of the values are light, they'll want this to be light as well?
> 
> We can also always only account for whatever the default is for this kit and add follow ups to account for different variants. I would also reach out to a designer with these questions!

I 100% think we should make sure this works with all variants. That is a excellent point.

nidaqg (dismissed):

skduncan (approved):
lgtm

nidaqg (approved):

PR #4311, opened 2025-02-24
[PLAY-1923] Add tooltip delay props

markdoeswork:
_old_

> Runway https://runway.powerhrg.com/backlog_items/PLAY-1923
> 
> adds `delay_open` and `delay_close` props to the rails tooltip, because we have this functionality in react, but it was missing from rails
> 
> You can check it out here https://pr4311.playbook.beta.px.powerapp.cloud/kits/tooltip/rails#delay_
> 
> The popper js library introduced "flashing" to the tool tip when i tried to implement dynamic delay logic

We are using a new library and this work in this story has been moved to https://github.com/powerhome/playbook/pull/4322


nidaqg:
Everything looks good @markdoeswork , but the 're-using tooltips' example seems to be acting strange, not seeing this behavior in prod.


https://github.com/user-attachments/assets/b2761ea3-018c-41c8-a0ab-d3a5c0f1410e

playbook/app/pb_kits/playbook/pb_tooltip/index.js
```diff
@@ -89,7 +89,7 @@ export default class PbTooltip extends PbEnhancedElement {
       clearTimeout(this.autoHideTimeout)
       this.autoHideTimeout = setTimeout(() => {
         this.hideTooltip()
-      }, 1000)
+      }, this.delayClose)
```
nidaqg, line 92: could this replacing of the hardcoded 1000 value with delayClose be causing the issue with the re-usable tooltip  @markdoeswork ?

markdoeswork: ah yes i need to keep that hard coded

PR #4327, opened 2025-02-26
[PBNTR-888] Adding Select Validity Message

carloslimasd:
**What does this PR do?**
Adding Select Validity Message

4. See addition/change

#### Checklist:

### DIFF
--- a/playbook/app/pb_kits/playbook/pb_select/select.rb
+++ b/playbook/app/pb_kits/playbook/pb_select/select.rb
@@ -21,6 +21,7 @@ module Playbook
       prop :options, type: Playbook::Props::HashArray, required: false, default: []
       prop :show_arrow, type: Playbook::Props::Boolean, default: false
       prop :required, type: Playbook::Props::Boolean, default: false
+      prop :validity_message, type: Playbook::Props::String, default: ""
 
       def classnames
         classname + inline_class + compact_class + show_arrow_class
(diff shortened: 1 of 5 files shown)


playbook/app/pb_kits/playbook/pb_select/select.rb
```diff
@@ -21,6 +21,7 @@ class Select < Playbook::KitBase
       prop :options, type: Playbook::Props::HashArray, required: false, default: []
       prop :show_arrow, type: Playbook::Props::Boolean, default: false
       prop :required, type: Playbook::Props::Boolean, default: false
+      prop :validity_message, type: Playbook::Props::String, default: ""
```
ElisaShapiro, line 24: It may make sense to keep the prop name consistent with the other form field builder items with this feature. `validation` is the prop name for form_field.rb/text_input and typeahead_field.rb/typeahead (can see in the [doc ex](https://github.com/powerhome/playbook/blob/e015a05d5f7d0e9d447c97f6fda9601e9abcfe06/playbook/app/pb_kits/playbook/pb_form/docs/_form_form_with_validate.html.erb#L25) and their .rb and _field.rb files).

carloslimasd: Since the one you mentioned is an object and we need just a string for this case, would it make sense to use "validation_message", @ElisaShapiro ?

ElisaShapiro: Great point on that! I agree that `validation_message` makes sense as the string version of the validation: { message: "" } } object! Keeps the language "almost-the-same" for devs using it while preserving the string vs. object distinction.

nidaqg (approved):

PR #4373, opened 2025-03-10
[PLAY-1871] Speed Up Local Vite Builds

kangaree:
**What does this PR do?**
- Use relative imports in docs files instead of "barrel files"
  - https://vite.dev/guide/performance#avoid-barrel-files
  - https://www.reddit.com/r/reactjs/comments/1ggpyms/vite_and_barrel_files_performance_issues/
- Rely on Vite's built-in typescript support instead of a separate plugin
- In website vite build, warm up heavily used library files, playbook.css and playbook-doc.js
- Fix "huge hamburger" icon before stylesheets are loaded by setting in html.erb
- Correct the sanitization of code snippets:
  - `import Button from "../../pb_button/_button"` becomes `import { Button } from 'playbook-ui'`
  - Combine multiple import statements into one (`import { X, Y, Z } from 'playbook-ui'`)

**Screenshots:**
Screenshot 2025-03-10 at 11 39 00 AM
Screenshot 2025-03-10 at 11 24 20 AM (2)

1. Regression test the website
2. Regression test Nitro

#### Checklist:


nidaqg:
Looks great @kangaree ! Small thing with the regex piece: the aliased imports are showing up a little strange, you can see it here for the Date kit:

Screenshot 2025-03-12 at 10 45 36 AM

I believe that is the only kit we use that syntax for.

Secondly, how would this effect in flight stories that are adding doc examples with the 'wrong' import syntax? Will those doc examples stop working once this PR ships or will they still work and we can make those changes later?

I would also make a note to update the Kit Generator in the future so the scaffolding generated adds the correct syntax for the imports.

kangaree:
@nidaqg 

Thanks! I'll fix the regex for Date.

This will be fine for in flight stories- we still have the barrel files in place and the doc examples will work for both cases and we can make changes later.

I'll see if I can make a quick change to the Kit Generator, otherwise I'll add a note to update it in the future!

nidaqg:
> @nidaqg
> 
> Thanks! I'll fix the regex for Date.
> 
> This will be fine for in flight stories- we still have the barrel files in place and the doc examples will work for both cases and we can make changes later.
> 
> I'll see if I can make a quick change to the Kit Generator, otherwise I'll add a note to update it in the future!

Awesome! Thanks @kangaree !

nidaqg (dismissed):

nidaqg (approved):

PR #4402, opened 2025-03-18
[PLAY-1714] Add Values to Icon Kit Color Prop

gmorris1242:
Adding the ability to pass the same text color values as the body kit including light and lighter. 

This seemed like the most straightforward and least disruptive way to do it. If more color options are required, there may be a better way. Let me know what you think.

[PLAY-1714](https://runway.powerhrg.com/backlog_items/PLAY-1714)
[Review ENV](https://pr4402.playbook.beta.gm.powerapp.cloud/kits/icon/rails)

### DIFF
--- a/playbook/app/pb_kits/playbook/pb_icon/_icon.scss
+++ b/playbook/app/pb_kits/playbook/pb_icon/_icon.scss
@@ -1,10 +1,18 @@
 @import "../tokens/colors";
 
+$additional_colors: (
+  "default": $text_lt_default,
+  "light": $text_lt_light,
+  "lighter": $text_lt_lighter,
+  "link": $primary
+);
+
 // All the merges below create $icon_colors, a map of all color tokens in colors.scss
 $merge_kits1: map-merge($status_colors, $category_colors);
 $merge_kits2: map-merge($merge_kits1, $product_colors);
 $merge_kits3: map-merge($merge_kits2, $text_colors);
-$icon_colors: map-merge($merge_kits3, $data_colors);
+$merge_kits4: map-merge($merge_kits3, $data_colors);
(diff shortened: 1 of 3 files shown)


playbook/app/pb_kits/playbook/pb_icon/_icon.scss
```diff
@@ -1,10 +1,18 @@
 @import "../tokens/colors";
 
+$additional_colors: (
+  "default": $text_lt_default,
```
nidaqg, line 4: I think we can remove default here since this is the same color as the default for the icon kit in general, so doesn't seem needed! The rest is good!

playbook/app/pb_kits/playbook/pb_icon/docs/_icon_color.html.erb
```diff
@@ -1,5 +1,11 @@
...
     <%= pb_rails("icon", props: { icon: "roofing", fixed_width: true, color: "product_5_background", size: "2x" }) %>
-<% end %>
\ No newline at end of file
+<% end %>
+
+<%= pb_rails("flex", props: {}) do %>
+    <%= pb_rails("icon", props: { icon: "user", fixed_width: true, color: "light", size: "2x" }) %>
+    <%= pb_rails("icon", props: { icon: "recycle", fixed_width: true, color: "lighter", size: "2x" }) %>
+    <%= pb_rails("icon", props: { icon: "roofing", fixed_width: true, color: "default", size: "2x" }) %>
+<% end %>
```
nidaqg, line 11: For these docs, looks like the runway story is asking for a separate doc for 'Text Color Values' so lets confirm that with Ashlee and then separate it into its own doc, that would make the most sense.

nidaqg (approved):

PR #4406, opened 2025-03-19
[PLAY-1952] Nav Kit: Set Default Cursor to Pointer for All Nav Items

gmorris1242:
This makes all nav items have the cursor hover state of pointer even if the link prop isn't set.

[PLAY-1952](https://runway.powerhrg.com/backlog_items/PLAY-1952)

[Review ENV](https://pr4406.playbook.beta.px.powerapp.cloud/kits/nav/react)

https://github.com/user-attachments/assets/8e0e986a-9ec4-4aa9-9b9e-81765dd99901

### DIFF
--- a/playbook/app/pb_kits/playbook/pb_nav/_nav.scss
+++ b/playbook/app/pb_kits/playbook/pb_nav/_nav.scss
@@ -43,3 +43,8 @@
     }
   }
 }
+
+[class*="pb_nav_list_kit_item"],
+[class*="pb_nav_list_item"] {
+  &:hover { cursor: pointer; }
+}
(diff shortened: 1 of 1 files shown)


playbook/app/pb_kits/playbook/pb_nav/_nav.scss
```diff
@@ -43,3 +43,8 @@
     }
   }
 }
+
+[class*="pb_nav_list_kit_item"],
+[class*="pb_nav_list_item"] {
+  &:hover { cursor: pointer; }
+}
```
nidaqg, line 50: This is great @gmorris1242 ! But can we scope this within the [class^="pb_nav_list"] above? We generally try to scope all scss within a parent selector for our kits and avoid having scss outside of the parent selector unless necessary

gmorris1242: Thanks yeah, that makes sense.

nidaqg (approved):

PR #4416, opened 2025-03-21
[PLAY-1957] Fix Inline Date Picker Icon Logic

gmorris1242:
I'm now realizing I may have taken this too far based on the stories expected behavior. I made it so the + shows back up after removing the selected date instead of staying a chevron. If we want to just have it be a + all the time, I can stash this and I think it's a quick fix for that. Let me know.

I also noticed that the sass was targeting these by id, which was fine when only one input was on the page, but with multiple inputs I was running into issues, so I changed that up a bit.
 
[PLAY-1957](https://runway.powerhrg.com/backlog_items/PLAY-1957)
[ENV](https://pr4416.playbook.beta.px.powerapp.cloud/kits/date_picker/react)


playbook/app/pb_kits/playbook/pb_date_picker/_date_picker.tsx
```diff
@@ -106,41 +106,60 @@ const DatePicker = (props: DatePickerProps): React.ReactElement => {
...
+      disableRange,
+      disableWeekdays,
+      enableTime,
+      format,
+      hideIcon,
+      inLine,
+      maxDate,
+      minDate,
+      mode,
+      onChange: handleDatePickerChange,
```
nidaqg, line 143: Instead of passing the onChange here in this way, did you consider moving the above handleDatePicker function to the date_picker_helper file instead and passing it in the config in a similar way to how we handle the yeahChangeHook [here](https://github.com/powerhome/playbook/blob/master/playbook/app/pb_kits/playbook/pb_date_picker/date_picker_helper.ts#L207)?

If we can move this logic to that file, I'm hoping we won't need separate logic for rails and react since that helper file is used by both. 

It may not work, so I would timebox it and see if its doable! Let me know if you have any issues/questions!

The secondary question here is, is the ask to move this logic to the kit itself so it behaves this way out of the box or to make changes that will allow for the code that currently lives in that doc example to work? I'm going to post that on the story and lets hold off applying my suggestions above till we get a response on that!

gmorris1242: Good call on the date_picker_helper file. That's so much better!

nidaqg (approved):

PR #4501, opened 2025-04-14
[PLAY-1906] - Overlay Kit: Add Opacity & Gradient prop - React

gmorris1242:
[PLAY-1906](https://runway.powerhrg.com/backlog_items/PLAY-1906)
[CSB](https://codesandbox.io/p/devbox/keen-murdock-dkpvkv?workspaceId=ws_Ni9WCwV5jk56bH746ZU8Ee)
[Review ENV](https://pr4501.playbook.beta.gm.powerapp.cloud/kits/overlay/react)

**EDITS: I added new example docs and split up the copy in the existing default example.**

This PR adds support for a `gradient={false}` prop to the Overlay kit, as well as an `opacity` prop for rendering a solid color overlay. Support for `black`, `white`, `success`, and `error` colors has also been added.

The original request called for a limited set of opacity values, but it was easier to allow the full range.

I added copy to the example docs—let me know if a new example section would be better. I was on the fence about that.

Tests for the new props are also included.
Screenshot 2025-04-14 at 3 10 01 PM
Screenshot 2025-04-14 at 3 10 07 PM


playbook/app/pb_kits/playbook/pb_overlay/_overlay.tsx
```diff
@@ -18,14 +18,16 @@ type OverlayProps = {
...
     children: React.ReactNode[] | React.ReactNode,
-    color: "card_light" | "bg_light" | "card_dark" | "bg_dark",
+    color: "card_light" | "bg_light" | "card_dark" | "bg_dark" | "black" | "white" | "success" | "error",
     data?: { [key: string]: string },
     dynamic?: false,
+    gradient?: boolean,
     htmlOptions?: { [key: string]: string | number | boolean | (() => void) },
     id?: string,
     layout: { [key: string]: string },
+    opacity: 'opacity_1' | 'opacity_2' | 'opacity_3' | 'opacity_4' | 'opacity_5' | 'opacity_6' | 'opacity_7' | 'opacity_8' | 'opacity_9' | 'opacity_10',
```
markdoeswork, line 28: Sometimes i wish we didn't have css tokens as prop values. But we don't encourage people to inject the classnames directly 

But i like that it works `<Overlay className="opacity_5"/>`

markdoeswork (dismissed):
LG2M

playbook/app/pb_kits/playbook/pb_overlay/_overlay.tsx
gmorris1242: Oh yeah, good point. Opacity almost seems like it could be global too.

nidaqg (approved):
Great work on this one @gmorris1242 ! Love the docs!

PR #4515, opened 2025-04-17
[PLAY-1889] Add hook and doc example

markdoeswork:
Runway https://runway.powerhrg.com/backlog_items/PLAY-1889

Adds a hook for the copy javascript from our copy button kit

kinda like how we expose the collapsible logic for the collapsible kit

You can check it out here https://pr4515.playbook.beta.gm.powerapp.cloud/kits/copy_button/react#copy-hook

### DIFF
--- a/playbook/app/pb_kits/playbook/pb_copy_button/_copy_button.tsx
+++ b/playbook/app/pb_kits/playbook/pb_copy_button/_copy_button.tsx
@@ -21,23 +21,25 @@ type CopyButtonProps = {
   timeout?: number
 }
 
-const CopyButton = ({
-  aria = {},
-  className,
-  data = {},
-  from = '',
-  id,
-  text = 'Copy',
-  tooltipPlacement = 'bottom',
-  tooltipText = 'Copied!',
-  value = '',
-  timeout = 1000,
-}: CopyButtonProps) => {
+const CopyButton = (props: CopyButtonProps) => {
+  const {
+    aria = {},
+    className,
+    data = {},
+    from = '',
+    id,
+    text= 'Copy',
+    timeout = 1000,
+    tooltipPlacement= 'bottom',
+    tooltipText = 'Copied!',
+    value = '',
+  } = props
+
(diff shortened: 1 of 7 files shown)


playbook/app/pb_kits/playbook/pb_copy_button/docs/_copy_button_hook.md
```diff
@@ -0,0 +1 @@
+We provide a `usePBCopy` hook that you can import to your project. This hook will return a function that you can call to copy the text to the clipboard.
```
nidaqg, line 1: This is great! Can we add copy for the connection the 'from' and the id on the text we're copying? Just so it's clearer hoe to use this!

Do you think a click to copy may make more sense as a doc example instead of a hover to copy?

Finally, is there a way to get some feedback that shows that text has been copied? For the default version we have the tooltip so it's clear it's been copied, any way to do that with the hook as well? Even if it's just something that can be done in the doc?

playbook/app/pb_kits/playbook/pb_copy_button/usePBCopy.ts
```diff
@@ -0,0 +1,45 @@
...
+    }, timeout)
+  }
+
+  useEffect(() => {
+    return () => {
+      window.clearTimeout(timerRef.current)
+    }
+  }, [])
+
+  return [copied, copy] as const
```
nidaqg, line 44: this is awesome!

playbook/app/pb_kits/playbook/pb_copy_button/_copy_button.tsx
```diff
@@ -1,88 +1,60 @@
...
-      }
-    }
-
-    setCopied(true)
-
-    setTimeout(() => {
-      setCopied(false)
-    }, 1000);
-  }
+  const classes = classnames(buildCss('pb_copy_button_kit'), globalProps({ aria, className, data, id }), className)
```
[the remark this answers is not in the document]
markdoeswork: this is probably an artfiact of some in progress experiments, ima tinker with it

PR #4538, opened 2025-04-22
[PLAY-2038] Copy Button Kit Go Live and Rails Copy Hook

kangaree:
**What does this PR do?**
- Make the Copy Button Kit live (out of beta)
- In playbook-rails.js, run function, `addCopyEventListeners`, which adds event listeners to any elements with `data-external-copy-value` or `data-external-copy-from`
- Create a "Copy Hook" doc that dispatches an event with the `from` name

https://runway.powerhrg.com/backlog_items/PLAY-2038

Screenshot 2025-04-22 at 4 24 53 PM (2)

1. Go to Playbook
2. You should be able to search for Copy Button kit and also click on it from the menu
3. Go to /kits/copy_button/rails#copy-hook
4. The other Rails kits should not be affected

#### Checklist:

### DIFF
--- a/playbook/app/pb_kits/playbook/pb_copy_button/docs/_copy_button_hook_rails.html.erb
+++ b/playbook/app/pb_kits/playbook/pb_copy_button/docs/_copy_button_hook_rails.html.erb
@@ -1,0 +1,24 @@
+<%= pb_rails("copy_button", props: { display: "none", from: "hookbody" } ) %>
+
+<div onclick="usePBCopy('hookbody')">
+  <%= pb_rails("body", props: { cursor: "pointer" }) do %>
+    <span id="hookbody">I'm a custom copy hook! Click this body to copy this text!</span>
+  <% end %>
+</div>
+
+<%= pb_rails("tooltip", props: {
+  trigger_element_selector: "#hookbody",
+  tooltip_id: "hookbody_tooltip",
+  position: 'top',
+  trigger_method: "click"
+}) do %>
+  Copied!
+<% end %>
+
+<%= pb_rails("textarea", props: { margin_top: "xs", placeholder: "Paste here" }) %>
+
+<script>
+  const usePBCopy = (id) => {
+    document.dispatchEvent(new CustomEvent(id));
+  };
+</script>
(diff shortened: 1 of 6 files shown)


kangaree:
I ended up combining the react copy hook pattern with the collapsible external control pattern. 

I'm happy to change things, though. The doc admittedly looks a bit funky.

playbook/app/pb_kits/playbook/pb_copy_button/docs/_copy_button_hook_rails.html.erb
```diff
@@ -0,0 +1,24 @@
+<%= pb_rails("copy_button", props: { display: "none", from: "hookbody" } ) %>
```
nidaqg, line 1: Curious to know why we need to add a hidden button to the page here

kangaree: We need a way to load `playbook/app/pb_kits/playbook/pb_copy_button/index.js` and use `connect()`. This needs to be loaded somewhere, but only for this function.

It does look a bit off, though.

What do you think about either:

- If the `text` prop is empty, don't display a button
- Having a separate `hook` boolean prop

nidaqg: hmmm.. that makes sense that if the kit isn't there at all, the enhanced element won't get triggered. Can we create a data attribute the dev needs to attach to the same div they are attaching the onClick to and our kit can look for that data attribute and initialize the enhanced element if it is present? Does that make sense?

kangaree: I'll take a look!

playbook/app/pb_kits/playbook/pb_copy_button/docs/_copy_button_hook_rails.html.erb
```diff
@@ -0,0 +1,16 @@
+<div data-external-copy-from="hookbody">
```
nidaqg, line 1: Everything looks great! Minor thing: Did you try using the data prop to attach this data attribute to the Body kit itself? If that works, it'll be cleaner code to copy without raw divs!

nidaqg (approved):

PR #4581, opened 2025-05-06
[PLAY-2087] Empty State Kit: Kit Buildout - Rails

gmorris1242:
[PLAY-2087](https://runway.powerhrg.com/backlog_items/PLAY-2087)

This is the Rails side build of the Empty State Kit

[Review ENV](https://pr4581.playbook.beta.gm.powerapp.cloud/kits/empty_state)


skduncan (approved):

ElisaShapiro (dismissed):

playbook/app/pb_kits/playbook/pb_empty_state/empty_state.html.erb
```diff
@@ -0,0 +1,25 @@
+<%= pb_content_tag(:div,
+aria: aria,
+class: classname,
+data: data,
+id: id,
+html_options: html_options,
+**combined_html_options) do %>
```
nidaqg, line 7: Great work on this one! Minor thing, I think with the pb_content_tag, you don't need to explicitly pass all these in anymore, it is done under the hood with the kit base defaults [here](https://github.com/powerhome/playbook/blob/master/playbook/lib/playbook/kit_base.rb#L149). I would test that locally by adding id, or html_options, but I believe that is the case!
The default is `:div` as well so that deos not need to be stated unless using a tag other then div

gmorris1242: Yep, tested that out locally and it works, nice!

PR #4610, opened 2025-05-09
[PLAY-2125] Phone Number Input: Fix the Error Styles for Country Search - Rails & React

gmorris1242:
[PLAY-2125](https://runway.powerhrg.com/backlog_items/PLAY-2125)
[Review ENV](https://pr4610.playbook.beta.px.powerapp.cloud/kits/phone_number_input/react#country-search-1)

This fixes the error styling on the phone input for the country search. 

Here's an example of how it currently looks: [Production](https://playbook.powerapp.cloud/kits/phone_number_input/react)
and the fixed version: [Review ENV](https://pr4610.playbook.beta.px.powerapp.cloud/kits/phone_number_input/react#country-search-1)

### DIFF
--- a/playbook/app/pb_kits/playbook/pb_text_input/_text_input.scss
+++ b/playbook/app/pb_kits/playbook/pb_text_input/_text_input.scss
@@ -71,7 +71,7 @@
     }
     &.error {
       .text_input_wrapper {
-        input,
+        input:not(.iti__search-input),
         .text_input {
           border-color: $error_dark;
         }
@@ -102,7 +102,7 @@
       [class*="pb_body_kit"] {
         margin-top: $space_xs / 2;
       }
-      input,
+      input:not(.iti__search-input),
       .text_input {
         border-color: $error;
       }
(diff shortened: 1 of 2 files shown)


playbook/app/pb_kits/playbook/pb_text_input/_text_input.scss
```diff
@@ -71,7 +71,7 @@
     }
     &.error {
       .text_input_wrapper {
-        input,
+        input:not(.iti__search-input),
```
ElisaShapiro, line 74: Since this part of the fix lives in a different kit I'm wondering if it's worth leaving a commented out note above or next to current line 74 and 105 to indicate that this :not selector has been added to the Text Input specifically for the Phone Number Input Country Search Input which uses it under the hood. May not be necessary though, I might just be overthinking. Works just as expected!

ElisaShapiro (dismissed):

playbook/app/pb_kits/playbook/pb_text_input/_text_input.scss
gmorris1242: Yeah that's a good idea

nidaqg (approved):

PR #4619, opened 2025-05-13
[PLAY-2186] Filter Kit: Sidebar Variant - React

kangaree:
**What does this PR do?**
Make a "sidebar" React variant which is used by setting the `variant` prop to "sidebar". 
Create a <Filter.Section> component that adds collapsible functionality.
Follow the [handoff](https://huddle.powerapp.cloud/projects/824/handoffs/1926).

https://runway.powerhrg.com/backlog_items/PLAY-2020

Things to consider:
3. What will still need to be done after this POC?
Make a Rails variant (if necessary).
Match the handoff more- include the different examples (e.g., inline).

**Screenshots:**
Screenshot 2025-06-04 at 11 56 31 AM

1. Go to kits/filter/react#sidebar

#### Checklist:
- [x] **RC** I have added an `inactive RC` label if not an active RC.

### DIFF
--- a/playbook/app/pb_kits/playbook/pb_filter/docs/_filter_sidebar.jsx
+++ b/playbook/app/pb_kits/playbook/pb_filter/docs/_filter_sidebar.jsx
@@ -3,8 +3,13 @@ import React, { useState } from 'react'
 import Filter from '../../pb_filter/_filter'
 import Select from '../../pb_select/_select'
 import TextInput from '../../pb_text_input/_text_input'
-
-import { useCollapsible, SectionSeparator, Checkbox, Flex, Table, Caption, Icon } from 'playbook-ui'
+import useCollapsible from '../../pb_collapsible/useCollapsible'
+import SectionSeparator from '../../pb_section_separator/_section_separator'
+import Checkbox from '../../pb_checkbox/_checkbox'
+import Flex from '../../pb_flex/_flex'
+import Table from '../../pb_table/_table'
+import Caption from '../../pb_caption/_caption'
+import Icon from '../../pb_icon/_icon'
 
 const people = [
   { name: 'Hera Syndulla', territory: 'A galaxy far far away, like really far away...', products: ['Insulation'] },


playbook/app/pb_kits/playbook/pb_filter/docs/_filter_side.jsx
```diff
@@ -0,0 +1,166 @@
...
+          filters={{
+            'Full Name': 'John Wick',
+            'City': 'San Francisco',
+            'Products': ['Doors'],
+          }}
+          isCollapsed={isCollapsed}
+          onCollapse={() => setIsCollapsed(!isCollapsed)}
+          onSortChange={SortingChangeCallback}
+          results={116}
+          side
```
nidaqg, line 40: From the initial investigative story, we had decided to call the prop variant with this option being 'sidebar' and the separate 'inline' being the other option when we build out that version. So variant would have 3 options: 'sidebar' (this version), 'inline' (to be built) and 'default' which would be what we currently have

playbook/app/pb_kits/playbook/pb_filter/docs/_filter_side.jsx
```diff
@@ -0,0 +1,166 @@
...
+                variant="link"
+            />
+          </Collapsible.Content>
+        </Collapsible>
+      </Filter>
+    </>
+  )
+}
+
+export default FilterSide
```
nidaqg, line 166: This is great! But it does feel like a lot if work on the frontend from the devs with all the collapsible logic. How much of this can live in the kit itself? It may be nice to have a single subcomponent for the Filter, something like FilterSection or similar that takes all the props outlined by Edward in the handoff. 
So something like:
<Filter {filter props here}>
<Filter.Section
collapsible={true} <--- if present, make it a collapsible otherwise no collapsible
collapsed={true} <--- optional initial state
header={true} <---- optional pro for header as per handoff, if collapsible this would be true but without collapsible this can be true or false
>
// children which would be all form etc kits as you have them
</Filter.Section>
<Filter.Section>
</Filter.Section>
</Filter>


It's worth exploring, to reduce the amount of setup needed on the frontend. Similar to the collapsible nav and how it works

playbook/app/pb_kits/playbook/pb_filter/docs/_filter_side.jsx
kangaree: I've updated this to `sidebar`.

ElisaShapiro (dismissed):

playbook/app/pb_kits/playbook/pb_filter/_filter.scss
```diff
@@ -88,6 +88,10 @@
     z-index: 1;
   }
 
+  [class^=pb_collapsible_kit] .toggle-content.filter_section_collapsible {
+    padding-top: 8px !important;
```
nidaqg, line 92: Lets use our spacing token here $space_xs

playbook/app/pb_kits/playbook/pb_filter/Filter/index.tsx
```diff
@@ -1,18 +1,29 @@
...
   | (FilterDoubleProps & {
       double?: boolean,
     })
+  | (FilterSidebarProps & {
+    sidebar?: boolean,
+  })
 
 const Filter = ({
   double = false,
+  sidebar = false,
```
nidaqg, line 18: For the prop name, we wanted to do 'variant` with the default being `default`, this version being `sidebar` and then eventually also doing the `inline` as an option. So usage would be `variant = 'sidebar'`

playbook/app/pb_kits/playbook/pb_filter/docs/_filter_sidebar.jsx
```diff
@@ -0,0 +1,223 @@
...
+            }}
+            sortValue={[{ name: 'popularity', dir: 'desc' }]}
+            width="xs"
+            {...props}
+        >
+          <Filter.Section
+              collapsed={isCollapsed}
+              collapsible
+              header
+              headerText="Full Name"
```
nidaqg, line 87: Are header and headerText both required in order for the header text to show up if it is a collapsible? From the code above it feels like the header prop isn't needed in order to render the headerText within the collapsible, but is needed if it is not collapsible? I would maybe render a non-collapsible section here as well to showcase the difference?

nidaqg:
Love the doc example!

playbook/app/pb_kits/playbook/pb_filter/Filter/index.tsx
kangaree: I updated this to be `variant = 'sidebar'`

nidaqg (approved):
awesome work!

PR #4638, opened 2025-05-21
Revert "Revert "[PLAY-2033] Advanced Table: Action bar for Selectable Rows logic - Rails ""

nidaqg:
Reverts powerhome/playbook#4637


kangaree (dismissed):
Looks good to me. 

One improvement we can make (maybe in another story)- We talk about how a unique `id` is needed, but both [#default-required-props](https://pr4638.playbook.beta.gm.powerapp.cloud/kits/advanced_table/rails#default-required-props) and [#loading-state](https://pr4638.playbook.beta.gm.powerapp.cloud/kits/advanced_table/rails#loading-state), use the same id: `beta_table`. 

Also, a lot of these references are "beta", but advanced table has moved out of beta for a while. I think we can just switch these out to things like "advanced_table"

PR #5019, opened 2025-08-15
Testing upgrade to React 18.3

nidaqg:
Testing React upgrade to 18.3 in PB and PB website as a pre-cursor to an upgrade to React 19. Everything seems to work correctly, excepts tests that need a bit more work

Main take aways:
- Webpacker is not compatable at all (switched to using vite-rails now)
  - Added react and react-dom to dedupe in vite.config
  - Added <%= vite_react_refresh_tag %> tag to layouts
  - Nothing with ReactDOM.render is compatible so any libray using that under the hood will cause issues
  - Same for Jest tests: Had to update to latest for @testing-library/react (11.2.7 to 16.3.0 in PB) and @testing-library/jest-dom (5.16.4 to 6.7.0 in PB)


Jenkinsfile
```diff
@@ -32,7 +32,7 @@ app.build(
     }
 
     stage('Test') {
-      shell "docker compose run --workdir /home/app/src/playbook web ./test.sh"
+      shell "docker compose run --workdir /home/app/src/playbook web ./test.sh || true"
```
nidaqg, line 35: Temporary thing to not run tests to get a review env + alpha. MUST REMOVE before we ship

nidaqg:
closing for now, Will revisit later to separate the webpacker deprecation from the React upgrade.

PR #5124, opened 2025-09-04
[PLAY-2466] Dialog Kit: Fixes for global props and classname not working - React

nidaqg:
**What does this PR do?** 
[Runway story](https://runway.powerhrg.com/backlog_items/PLAY-2466)

- Fixes for global props and classname not working as expected for react it
- global props and classnames were being applied to the wrong div.

4. See addition/change

#### Checklist:

### DIFF
--- a/playbook/app/pb_kits/playbook/pb_dialog/_dialog.tsx
+++ b/playbook/app/pb_kits/playbook/pb_dialog/_dialog.tsx
@@ -74,7 +74,7 @@ const Dialog = (props: DialogProps): React.ReactElement => {
    const dataProps = buildDataProps(data)
    const htmlProps = buildHtmlProps(htmlOptions);
   const dialogClassNames = {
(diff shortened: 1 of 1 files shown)


kangaree (commented):
Looks good to me!

I took a look at Rails and it looks like global props and classname are also being applied to the wrapper instead of the dialog. Would it make sense to fix that in this story too?

nidaqg:
> Looks good to me!
> 
> I took a look at Rails and it looks like global props and classname are also being applied to the wrapper instead of the dialog. Would it make sense to fix that in this story too?

Thanks @kangaree ! So for the rails side, the global props are applied via the pb_content_tag that is on the dialog itself. I did test locally and the global props show up on the correct div in rails. For the react side, [heres a csb](https://codesandbox.io/p/devbox/musing-dubinsky-dd4s7p) showing the props working

PR #6141, opened 2026-05-06
[PLAY-2868] Input Kits: Expose `data-default-value` for smart filter reset

kangaree:
**What does this PR do?**
https://runway.powerhrg.com/backlog_items/PLAY-2868

**Rails** text input, select, date picker, and dropdown kits now emit `data-default-value` when a default is set, so Nitro can restore business defaults after form.reset() when smart reset is opted in.
**React** date picker and dropdown kits also use `data-default-value` because there is a default value prop that's used. The text input and select were not affected because they don't have a default value prop. 

Other notes:
- Dropdown: omit attribute when no default (was always present)
- Date picker React: fix buildDataProps key so DOM gets data-default-value

Screenshot 2026-05-07 at 4 22 52 PM

1. Test these inputs and make sure the inputs can still set value as normal, but it can also be reset. 

#### Checklist:

### DIFF
--- a/playbook/app/pb_kits/playbook/pb_dropdown/_dropdown.tsx
+++ b/playbook/app/pb_kits/playbook/pb_dropdown/_dropdown.tsx
@@ -30,6 +30,20 @@ function serializeDropdownFilterResetDefault(
     const optionList: GenericObject[] = Array.isArray(dropdownOptions)
         ? dropdownOptions
         : [];
+    const optionDefaultId = (option: GenericObject | undefined): string | undefined => {
+        if (!option) return undefined;
+
+        const id = option.id;
+        if (id != null && id !== "") return String(id);
+
+        const matched = optionList.find((listOption: GenericObject) => (
+            (option.value != null && listOption.value === option.value) ||
+            (option.label != null && listOption.label === option.label)
+        ));
+
+        if (matched?.id != null && matched.id !== "") return String(matched.id);
+        return undefined;
+    };
 
     if (variant === "quickpick") {
         if (typeof defaultValue === "string" && defaultValue) {
@@ -48,13 +62,13 @@ function serializeDropdownFilterResetDefault(
                 : [];
         if (!arr.length) return undefined;
         const ids = arr
-            .map((v) => (v as GenericObject)?.id)
+            .map((v) => optionDefaultId(v as GenericObject))
             .filter((id) => id != null && id !== "");
         return ids.length ? ids.join(",") : undefined;
     }
     if (defaultValue && typeof defaultValue === "object" && !Array.isArray(defaultValue)) {
-        const id = (defaultValue as GenericObject).id;
-        if (id != null && id !== "") return String(id);
+        const id = optionDefaultId(defaultValue as GenericObject);
+        if (id) return id;
     }
     return undefined;
 }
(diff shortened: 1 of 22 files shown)


playbook/app/pb_kits/playbook/pb_dropdown/_dropdown.tsx
```diff
@@ -21,6 +21,44 @@ import {
...
+    if (variant === "quickpick") {
+        if (typeof defaultValue === "string" && defaultValue) {
+            const matched = optionList.find(
+                (opt: GenericObject) => opt.label?.toLowerCase() === defaultValue.toLowerCase()
+            );
+            if (matched?.id != null && matched.id !== "") return String(matched.id);
+        }
+        return undefined;
+    }
+    if (multiSelect) {
```
nidaqg, line 57: I don't think this is applying @kangaree , I don't see it in the DOM for the 'MultiSelect Default Value' doc example. I see the data attribute for the default + the quick pick though!

playbook/app/pb_kits/playbook/pb_select/_select.tsx
```diff
@@ -100,6 +100,15 @@ const Select = ({
...
   const selectId = (inputOptions?.id as string) || name
 
+  const filterResetDefaultValue = (() => {
+    if (value === undefined || value === null || value === '') return undefined
+    if (Array.isArray(value)) {
+      const joined = value.filter((v) => v !== '' && v != null).join(',')
+      return joined || undefined
+    }
+    return String(value)
+  })()
```
nidaqg: I'm unsure on what this will do here. Doesn't look like we have a default value prop for the select, and looks like you're using the 'value' prop? If we actually use value as a prop, it generates that data attribute in the DOM but then no other selection can be made (clicking other options does not change the selection)

playbook/app/pb_kits/playbook/pb_text_input/_text_input.tsx
```diff
@@ -182,6 +182,9 @@ const TextInput = (props: TextInputProps, ref: React.LegacyRef<HTMLInputElement>
   // Filter out emojiMask from props passed to DOM element
   const { emojiMask: _emojiMask, ...domProps } = props
 
+  const filterResetDefaultValue =
+    value !== '' && value !== undefined && value !== null ? String(value) : undefined
```
nidaqg: For this one looks like when we edit the text in the input, the string passed to that data attribute updates in real time as well instead of retaining the initial default. Would this not mean that the reset to default would only ever 'revert' to what is currently in the input as opposed to the default?

playbook/app/pb_kits/playbook/pb_dropdown/_dropdown.tsx
kangaree: I'm updating this so it'll be `data-default-value="us,it"`

playbook/app/pb_kits/playbook/pb_select/_select.tsx
kangaree: Removing it! Thanks for flagging it, didn't think about how the React Select has controlled state / no default value prop.

playbook/app/pb_kits/playbook/pb_text_input/_text_input.tsx
kangaree: Also removing this React kit since it doesn't have a default value prop.

nidaqg (dismissed):

nickamantia (approved):

PR #6275, opened 2026-06-09
[PLAY-2933] POC: Typeahead (Multi-Select) – Drag-and-Drop Reordering of Selected Pills

kangaree:
**What does this PR do?**
https://runway.powerhrg.com/backlog_items/PLAY-2933

This is a POC for a drag and drop reordering of pills. It adds these props: `enable_pill_reorder`, `show_pill_index`, and `pill_drag_handle`. It uses the draggable kit. 

Adds opt-in drag-and-drop reordering to Typeahead multi-select pills via `enable_pill_reorder`, with optional `show_pill_index` (numbered prefixes) and `pill_drag_handle` (hide the grip and drag the whole pill). Pills render in a Draggable strip inside ValueContainer; order commits on drop and is available through onChange and the pb-typeahead-kit-{id}-result-option-reorder custom event, with keyboard support via Ctrl+Shift+Arrow. To make this work inside react-select, the Draggable kit gains opt-in pointer drag on Draggable.Item (pointer_drag / handle_pointer_drag) so pills can use mouse drag where HTML5 Drag and Drop breaks, without changing default Draggable behavior elsewhere.

Screenshot 2026-06-12 at 3 56 43 PM

1. Go to https://pr6275.playbook.wc.beta.hq.powerapp.cloud/kits/typeahead/react#typeahead_draggable_pills

#### Checklist:


nidaqg:
Hey @kangaree looks like we have a regression in the draggable kit: on prod you can drag items even if not holding the drag handle, but this makes it so you can only drag with the handle, does that make sense? Test with any doc that has the drag handle

kangaree:
> Hey @kangaree looks like we have a regression in the draggable kit: on prod you can drag items even if not holding the drag handle, but this makes it so you can only drag with the handle, does that make sense? Test with any doc that has the drag handle

Thanks for pointing this out @nidaqg ! I will fix.

playbook/app/pb_kits/playbook/pb_typeahead/docs/_typeahead_draggable_pills.html.erb
```diff
@@ -0,0 +1,56 @@
...
+}) %>
+
+<%= pb_rails("body", props: {
+  id: "typeahead-draggable-order",
+  color: "lighter",
+  margin_top: "xs",
+  text: "Current order: Year -> Territory -> Branch",
+}) %>
+
+<%= pb_rails("button", props: { id: "typeahead-draggable-reset", margin_top: "sm", text: "Reset", variant: "secondary" }) %>
```
ElisaShapiro: Reset is not working for the rails doc but I think that will just require an update to the javascript_tag (not any of the kit logic)

kangaree: You're right- the reset button was clearing the input.

I looked into it, and unfortunately a proper fix might be more involved since this is Rails- we need to add more logic to keep the "order" of pills, which feels like overkill for a doc example.

I went ahead and removed the button from the Rails doc!

playbook/app/pb_kits/playbook/pb_draggable/utilities/touchDrag.ts
```diff
@@ -171,3 +297,91 @@ export const bindTouchDrag = ({
...
+  itemElement.addEventListener('mousedown', handleMouseDown);
+  document.addEventListener('mousemove', handleMouseMove);
+  document.addEventListener('mouseup', handleMouseUp);
+
+  return () => {
+    itemElement.removeEventListener('mousedown', handleMouseDown);
+    document.removeEventListener('mousemove', handleMouseMove);
+    document.removeEventListener('mouseup', handleMouseUp);
+  };
+};
```
nidaqg, line 388: Looks like we're registering document-level mousemove/mouseup listeners for every draggable pill? Cleanup is present, but if we have many pills in a typeahead, this may mean that it makes every mousemove run through several inactive listeners?. It may be better to attach document listeners only after mousedown starts an active drag, then remove them on mouseup/cancel. thoughts?

kangaree: Thanks! Agreed- there were listeners for every pill at mount. I fixed it by attaching/coming off on mousedown/mouseup!

nidaqg (approved):

kangaree:
Closing this for now. Note to self when picking this up again: make sure the text isn't "selectable".

PR #6359, opened 2026-07-06
[PLAY-3123] Playbook Website: Rails Playgrounds

kangaree:
**What does this PR do?**
https://runway.powerhrg.com/backlog_items/PLAY-3123

Adds a Rails Playground tab for select kit pages that reuses the existing props panel but renders previews through a server-side pb_rails endpoint. Includes ERB codegen, live preview wiring, and translation for JSX/ERB children so compound kits (card, flex, dialog) behave like their React playground counterparts.

Kits:

- button
- badge
- caption
- icon
- title
- card
- flex
- checkbox
- text_input
- dialog
- dropdown

Screenshot 2026-07-06 at 2 35 27 PM

1. Go through kits and click playground in rails. 

#### Checklist:


playbook-website/app/services/playground/rails_renderer.rb
```diff
@@ -0,0 +1,308 @@
...
+    def render_compound_dialog_preview(merged_props, dialog_id)
+      simple_props = merged_props.except(:title, :text, :cancel_button, :confirm_button)
+
+      render_pb_kit("dialog", simple_props, safe_join([
+                                                        render_pb_kit("dialog/dialog_header", {
+                                                                        id: dialog_id,
+                                                                        title: "Header Title inside Dialog.Header",
+                                                                      }),
+                                                        @children.present? ? render_pb_kit("dialog/dialog_body") { @children.to_s.html_safe } : render_pb_kit("dialog/dialog_body", { text: "" }),
+                                                        render_pb_kit("dialog/dialog_footer", {
```
nidaqg, line 191: Can we avoid calling html_safe on playground-provided children unless the content came from a trusted renderer? The generic children fallback and dialog subcomponent body here look like they can return raw user-entered HTML. Since LiveExampleRails later executes inline scripts from the rendered HTML, this feels like a risky path for security issues, like we had for the previous playground.

playbook-website/app/javascript/components/Website/src/pages/KitShow/Tabs/Playground/hooks/useRailsPlaygroundPreview.ts
```diff
@@ -0,0 +1,93 @@
...
+            },
+            body: JSON.stringify(payload),
+            signal: controller.signal,
+          },
+        );
+
+        const data = await response.json();
+
+        if (!response.ok) {
+          setError(data.error || "Failed to render Rails preview");
```
nidaqg, line 73: Should we clear html here as well when a Rails preview request errors or when a new request starts? Right now a failed prop combo can show an error while still rendering the last successful preview

playbook-website/app/controllers/playground_controller.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+class PlaygroundController < ApplicationController
+  skip_before_action :verify_authenticity_token, only: [:preview]
```
nidaqg: Curious to know why this is needed here? Since this accepts arbitrary preview payloads and renders HTML, it would be safer to keep normal CSRF protections unless there’s a specific reason the React app cannot send the token

playbook-website/app/javascript/components/Website/src/components/LiveExamples/LiveExampleRails.tsx
```diff
@@ -136,6 +137,19 @@ const removePortaledPopovers = (
...
+  container.querySelectorAll("dialog[open]").forEach((dialog) => {
+    if (typeof dialog.close === "function") {
+      dialog.close();
+    }
+  });
+};
+
+const reinitializeRailsKits = (container: HTMLElement): void => {
+  PbKitRegistry.rescan(container);
+  window.dispatchEvent(new Event("turbo:frame-load"));
```
nidaqg: Feels like we're doing a lot of init passes per preview update. Do we know if kits with direct DOM listeners, especially dropdown/dialog/date-picker style behavior, don’t double-bind across playground updates? A smaller single initialization path would be safer if possible

playbook-website/app/controllers/playground_controller.rb
kangaree: Fair point. This wasn't wired up yet to use the CSRF from the React app. 

I removed `skip_before_action :verify_authenticity_token` and updated `useRailsPlaygroundPreview` to send X-CSRF-Token from the existing `csrf_meta_tags`.

playbook-website/app/javascript/components/Website/src/components/LiveExamples/LiveExampleRails.tsx
kangaree: Good question. I don't think it would double-bind as we have logic to counter it (`element._pbEnhanced || this.elements.has(element)`)

I agree this is a heavier path than ideal. I'm taking a look at a smaller single init path!

playbook-website/app/javascript/components/Website/src/pages/KitShow/Tabs/Playground/hooks/useRailsPlaygroundPreview.ts
kangaree: Updated when a new request starts!

playbook-website/app/services/playground/rails_renderer.rb
kangaree: Agreed. I updated this to escape the playground-provided children.

kangaree (commented):
Thanks so much for your review! I'm still working on your feedback, but wanted to provide some updates for now.

playbook-website/app/javascript/components/Website/src/components/LiveExamples/LiveExampleRails.tsx
kangaree: https://github.com/powerhome/playbook/pull/6359/changes/6f5e4d3836503efd2c57daef809c691372eb0bd8

ElisaShapiro (dismissed):
The only thing that pops out to me vs. React side is the code snippet not having its syntax highlighted (doc code snippets + react playground use react-syntax-highlighter) but that's small potatoes compared to the rest (and definitely not a blocker for this, can be explored at any time in the playground POC/beta life-cycle). Really awesome job!

playbook-website/app/services/playground/trusted_html.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Playground
+  # Helpers for playground preview HTML safety.
+  #
+  # Plain children from the props panel are user input — escape them.
+  # Output from pb_rails / structured children renderers is trusted kit HTML.
+  module TrustedHtml
```
nidaqg, line 12: This is great! TrustedHtml.plain_text is the right idea, but the trust boundary still feels a bit uneven perhaos
ChildrenRenderer already returns escaped text without html_safe, which is good. The weaker spots are where we escape (or assume kit templates escape), then later mark a joined buffer trusted.. like  JsxChildrenRenderer#playground_html_content, ErbChildrenRenderer’s outer from_renderer/join, and the compound dialog body path via RailsRenderer.

Could we keep a clearer split: raw user/plain text always goes through the escaped helper, and only actual pb_rails renderer output gets marked trusted (ideally via safe_join for mixed segments)? Also worth escaping the fallback in JsxChildrenRenderer#render_inner_content when nested JSX parse fails, instead of relying on kit templates alone

nidaqg: I may be overthinking all of this for sure, but we dont want a repeat of the issues we had with the OG playground that we shifted to Huddle

kangaree: Agreed- we don't want to run into the same issues with the OG Playground. 

I've made improvements in sanitization / escaping:
- raw plain text always goes through `TrustedHtml.plain_text`
- replace `from_renderer` with `safe_join`: mix kit + plain text instead of blanket accepting it with `from_renderer`
- escape JSX parse fallbacks in `render_inner_content`.

kangaree:
Merging master to get the new Icon fixes - https://github.com/powerhome/playbook/pull/6401.

kangaree:
I addressed feedback:

- Cap preview input size and redact error details
- Only allow schema props; block html_options / data / aria / style
- Reject URL/path/.svg-like icon values so the playground can’t poke Icon’s file/URL paths
- Tighten ERB children parsing (\A anchor) and only allow known child kit paths

PR #6379, opened 2026-07-09
[PLAY-3088] Figma Code Connect Trial

edwardsaleeby:
[https://runway.powerhrg.com/backlog_items/PLAY-3088?from=active_sprint ](
https://runway.powerhrg.com/backlog_items/PLAY-3088?from=active_sprint )

- Adds Figma Code Connect infrastructure for dual React + Rails Dev Mode snippets
- Pilots the pattern on **Title**, with mapping files co-located in `pb_title/docs/`
- Uses two shared parser configs (React label + Rails label)

1. Open Figma and place Title component; configure it's props.
2. Use Dev Mode and check for generated Rails and React Code

### DIFF
--- a/playbook/app/pb_kits/playbook/pb_title/docs/Title.figma.ts
+++ b/playbook/app/pb_kits/playbook/pb_title/docs/Title.figma.ts
@@ -1,0 +1,44 @@
+// url=https://www.figma.com/design/Zj4FIeaa9fyMn2LcuL15hG/Playbook-Design-System-Library?node-id=98-815
+import figma from "figma"
+
+function rubyProp(name: string, value: unknown): string {
+  if (value === undefined || value === null) return ""
+  if (typeof value === "boolean") return `${name}: ${value}`
+  if (typeof value === "number") return `${name}: ${value}`
+  return `${name}: "${value}"`
+}
+
+const instance = figma.currentLayer
+const textLayer = instance.findText("@text")
+const text = textLayer && textLayer.type === "TEXT" ? textLayer.textContent : ""
+const size = instance.getEnum("Size", {
+  "1": 1,
+  "2": 2,
+  "3": 3,
+  "4": 4,
+})
+const color = instance.getEnum("Color", {
+  default: "default",
+  light: "light",
+  lighter: "lighter",
+  link: "link",
+  success: "success",
+  error: "error",
+})
+const bold = instance.getEnum("Bold", {
+  true: true,
+  false: false,
+})
+
+const props = [
+  rubyProp("text", text),
+  rubyProp("size", size),
+  rubyProp("color", color),
+  rubyProp("bold", bold),
+].filter(Boolean).join(", ")
+
+export default {
+  id: "pb-title-rails",
+  example: figma.code`<%= pb_rails("title", props: { ${props} }) %>`,
+  metadata: { nestable: true },
+}
--- a/playbook/app/pb_kits/playbook/pb_title/docs/Title.figma.tsx
+++ b/playbook/app/pb_kits/playbook/pb_title/docs/Title.figma.tsx
@@ -1,0 +1,39 @@
+import React from "react"
+import figma from "@figma/code-connect/react"
+import Title from "../_title"
+
+figma.connect(
+  Title,
+  "https://www.figma.com/design/Zj4FIeaa9fyMn2LcuL15hG/Playbook-Design-System-Library?node-id=98-815",
+  {
+    props: {
+      text: figma.textContent("@text"),
+      size: figma.enum("Size", {
+        "1": 1,
+        "2": 2,
+        "3": 3,
+        "4": 4,
+      }),
+      color: figma.enum("Color", {
+        default: "default",
+        light: "light",
+        lighter: "lighter",
+        link: "link",
+        success: "success",
+        error: "error",
+      }),
+      bold: figma.enum("Bold", {
+        true: true,
+        false: false,
+      }),
+    },
+    example: ({ bold, color, size, text }) => (
+      <Title
+          bold={bold}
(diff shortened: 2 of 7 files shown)


playbook/app/pb_kits/playbook/pb_title/docs/Title.figma.ts
```diff
@@ -0,0 +1,44 @@
+// url=https://www.figma.com/design/Zj4FIeaa9fyMn2LcuL15hG/Playbook-Design-System-Library?node-id=98-815
+import figma from "figma"
+
+function rubyProp(name: string, value: unknown): string {
+  if (value === undefined || value === null) return ""
+  if (typeof value === "boolean") return `${name}: ${value}`
+  if (typeof value === "number") return `${name}: ${value}`
+  return `${name}: "${value}"`
```
nidaqg, line 8: This here inserts Figma-controlled strings directly into a Ruby string literal. Text containing quotes, backslashes, newlines, or #{...} can possibly create invalid or executable Ruby when the generated snippet is copied.  This can be unsafe for arbitrary text since the Title text comes directly from a Figma text layer and an accidentally crafted figma val;ue could generate Ruby interpolation.
Could we serialize values as safe Ruby literals rather than interpolating them directly?

playbook/package.json
```diff
@@ -67,6 +72,7 @@
     "intl-tel-input": "^24.7.0"
   },
   "devDependencies": {
+    "@figma/code-connect": "^1.4.9",
```
nidaqg, line 75: We should pin it to an exact version as opposed to adding the caret. The Code Connect project has previously shipped dependency security fixes and we dont want to accidentally install something malicious

yarn.lock
```diff
@@ -48,6 +48,17 @@
     "@jridgewell/gen-mapping" "^0.3.0"
```
nidaqg, line 48: There seems to be a pretty large footprint in the dependency tree for this package. In theory this should not be bundling in our dist but it does look like our package size goes up with this PR.

playbook/figma.config.react.json
```diff
@@ -0,0 +1,16 @@
+{
+  "codeConnect": {
+    "include": [
+      "app/pb_kits/playbook/**/docs/*.figma.tsx",
+      "app/pb_kits/playbook/pb_*/_*.tsx"
```
nidaqg, line 5: This looks like its essentially scanning all kit files, not just Title. Is that the intent? This will effect the parse time, publish, time, CI time, memory usage etc, not to mention that the failure blast radius is much larger if unrelated kits contain syntax thew parser cannot handle. The config should use the narrowest source glob Code Connect needs. If component files are necessary for resolving imports, this may be unavoidable, but it should be measured now with one kit and estimated at full scale. Is this the requirement?

playbook/app/pb_kits/playbook/pb_title/docs/Title.figma.tsx
```diff
@@ -0,0 +1,39 @@
...
+    example: ({ bold, color, size, text }) => (
+      <Title
+          bold={bold}
+          color={color}
+          size={size}
+          text={text}
+      />
+    ),
+  }
+)
```
nidaqg, line 39: Given that PB already generates structured metadata, I would investigate generating a meaningful portion of Code Connect mappings from that metadata rather than building another manually maintained prop-definition layer. Is this file auto generated like the metadata or manually maintained? if manually maintained thats too much to keep in sync when you lok at 100+ kits and hoe many changes we ships daily

playbook/package.json
```diff
@@ -39,7 +39,12 @@
     "generate:global-props-metadata": "node scripts/generate-global-props-metadata.mjs",
     "generate:all-ai-metadata": "node scripts/generate-ai-metadata.mjs && node scripts/generate-global-props-metadata.mjs",
     "build:ai": "node scripts/build-ai-dist.mjs",
-    "build:ai:full": "yarn generate:all-ai-metadata && yarn build:ai"
+    "build:ai:full": "yarn generate:all-ai-metadata && yarn build:ai",
+    "figma:connect:parse:react": "figma connect parse --config figma.config.react.json",
+    "figma:connect:parse:rails": "figma connect parse --config figma.config.rails.json",
+    "figma:connect:publish:react": "figma connect publish --config figma.config.react.json",
+    "figma:connect:publish:rails": "figma connect publish --config figma.config.rails.json",
+    "figma:connect:publish": "yarn figma:connect:publish:react && yarn figma:connect:publish:rails"
```
nidaqg, line 47: All these publish commands need a figma token. So every time a figma.ts or figma.tsx mapping changes, that change only exists in the repo. It does not automatically update  Figma till someone runs this command. This does not seem maintainable with the speed of PB changes. Either we all need access tokens or (a better option) we need a figma token at at powerhome level that is set in github secrets similar to FA. We would than include these scripts in our build commands similar to the metadata scripts so its automated. 

Ideally we would need:

- A team-controlled Figma credential
- Protected GitHub secrets
- Publishing restricted to the main branch
- No secret access from forked pull requests
- Parse validation before merge
- A workflow that only publishes when Code Connect files change
- Clear ownership when publishing fails
- A way to review generated snippets before they become visible in Figma

I would also need to look into safety meansures in Figma like we did for npm and rubygems given the amount of supply chain attacks that happen these days.

edwardsaleeby:
I closed this PR because it was to test out the Figma Code Connect.

It proved that it is possible to generate Rails and React code based on the templates, but we are very limited in capabilities as the template is not able to work with Figma's API. This meaning, we cannot simulate or apply Playbook's global props unless we make and add to our Figma component variant matrices. That would compound how things work and managing and working with components, and drastically affect Figma's overall performance.

PR #6423, opened 2026-07-22
[PLAY-3070] Timestamp Kit (React): Support Timezone Formats

nickamantia:
[PLAY-3070](https://runway.powerhrg.com/backlog_items/PLAY-3070?from=active_sprint)

This PR adds ActiveSupport timezone to the React side of the kit. You can pass in ether IANA or ActiveSupport timezones into the the `timezone` prop.

4. See addition/change

#### Checklist:


playbook/app/pb_kits/playbook/pb_kit/dateTime.ts
```diff
@@ -1,3 +1,5 @@
+import { ACTIVE_SUPPORT_TIME_ZONE_MAPPING } from './activeSupportTimeZones'
```
nidaqg: So since dateTime.ts is shared by multiple kits and utilities beyond Timestamp, I'm a little wary of adding this here. Feels likw  statically importing the full ActiveSupport mapping into dateTime.ts could potentially add it to consumers that only use unrelated helpers such as fromNow, toIso, or DatePicker quick-pick functions (the horror). 

Can we just perhaps make it so we scope the conversion to Timestamp rather than changing the shared dateTime.ts formatters?

We could place the mapping and resolveTimeZone helper inside the Timestamp kit folder itself, resolve the timezone prop once in the Timestamp component, and pass the resulting IANA value into the existing toTime, toDay, toMonth, toYear, and other utilities.

That would preserve the existing contract and bundle behavior of dateTime.ts, avoid changing every consumer of those shared utilities, and keep the implementation and test coverage aligned with what we want at present

Lastly, we are manually copying Rails timezone mapping here, which is fine but we should document that in the file like: 
```
// Mirrored from ActiveSupport::TimeZone::MAPPING
// Rails 8.0.1
// Last synced: July 2026
```

So future us knows what is happening in case we need to update this when we update Rails (potentially)

ElisaShapiro (approved):

nidaqg (approved):

PR #6424, opened 2026-07-22
[PLAY-3104] Typeahead Reset to Default Preparation - `data-default-value` and :set

kangaree:
https://runway.powerhrg.com/backlog_items/PLAY-3104

Add typeahead :set event to restore selections after clear. 

Set up for smart reset defaults in nitro_search_playbook_filter:

- IndicatorsContainer.tsx — listens for :set and calls react-select setValue (lives here because ClearIndicator unmounts when empty)
- _typeahead.tsx — stamps data-default-value on the kit root from initial defaultValue / default_options

4. See addition/change

#### Checklist:

### DIFF
--- a/playbook/app/pb_kits/playbook/pb_typeahead/_typeahead.tsx
+++ b/playbook/app/pb_kits/playbook/pb_typeahead/_typeahead.tsx
@@ -216,12 +216,24 @@ const Typeahead = forwardRef<HTMLInputElement, TypeaheadProps>(
     // Create a ref to access React Select instance
     const selectRef = useRef<any>(null)
 
-    // Mount-time stamp for smart Defaults (:clear → read → :set). Not updated on selection change.
+    // Mount-time data-default-value for smart Defaults (:clear → read → :set).
+    // Only label/value, with try/catch so non-serializable extras never block mount.
     const [dataDefaultValue] = useState(() => {
       const defaultValue = props.defaultValue
       if (defaultValue == null) return undefined
       const options = Array.isArray(defaultValue) ? defaultValue : [defaultValue]
-      return options.length ? JSON.stringify(options) : undefined
+      if (!options.length) return undefined
+
+      try {
+        return JSON.stringify(
+          options.map((option: SelectValueType) => ({
+            label: option?.label,
+            value: option?.value,
+          })),
+        )
+      } catch {
+        return undefined
+      }
     })
 
     const dialogCtx = useContext(DialogContext)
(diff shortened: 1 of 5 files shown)


playbook/app/pb_kits/playbook/pb_typeahead/_typeahead.tsx
```diff
@@ -216,6 +216,14 @@ const Typeahead = forwardRef<HTMLInputElement, TypeaheadProps>(
     // Create a ref to access React Select instance
     const selectRef = useRef<any>(null)
 
+    // Mount-time stamp for smart Defaults (:clear → read → :set). Not updated on selection change.
+    const [dataDefaultValue] = useState(() => {
+      const defaultValue = props.defaultValue
+      if (defaultValue == null) return undefined
+      const options = Array.isArray(defaultValue) ? defaultValue : [defaultValue]
+      return options.length ? JSON.stringify(options) : undefined
```
nidaqg: so we're serializing full option object into a DOM attribute via JSON.stringify here. My concern is that Typeahead options can contain other data as well and if it contains something non-serializable, JSON.stringify will throw during render and prevent the Typeahead from mounting even though react-select would not have an issue consuming that option.

Would it be safer to serialize only the fields we actually need for restoring state (label and value?) or otherwise guard against serialization failures in some way?

playbook/app/pb_kits/playbook/pb_typeahead/components/IndicatorsContainer.tsx
```diff
@@ -1,16 +1,32 @@
...
+
+  useEffect(() => {
+    if (!selectProps?.id) return
 
+    const handleSet = (event: Event) => {
+      const detail = (event as CustomEvent).detail
+      const value = selectProps.isMulti
+        ? (Array.isArray(detail) ? detail : [detail]).filter(Boolean)
+        : (Array.isArray(detail) ? detail[0] : detail) ?? null
+      setValue(value, 'select-option')
```
nidaqg: calling setValue(value, 'select-option') sends the programmatic `:set` operation through the real-selection route in our handleOnChange. 

This could potentially clear validation state, mark the field as user-selected, invoke onChange, dispatch the click-driven result-option-select event, etc. Basically it could trigger other selection related behaviour.

Also, for multi-select restores, react-select supplies no individual option, so that event's detail becomes the entire restored array, changing the payload shape expected by listeners. I think we need to track that a `:set` is in progress and suppress user-only side effects, or explicitly define and test a separate programmatic event contract

playbook/app/pb_kits/playbook/pb_typeahead/_typeahead.tsx
kangaree: I made it so it only uses label and value and also wrapped everything in a `try catch`!

playbook/app/pb_kits/playbook/pb_typeahead/components/IndicatorsContainer.tsx
kangaree: Addressed by gating :set with a programmatic flag. Value updates and `onChange` still run (for form sync), but user-only side effects are skipped — no result-option-select, and we don’t mark the field as user-selected / clear validation as if it were a click.

nidaqg (approved):

PR #6469, opened 2026-07-31
[PLAY-3072] Rails TextInput with Add On - 2 Bugs

ElisaShapiro:
[PLAY-3072](https://runway.powerhrg.com/backlog_items/PLAY-3072) addresses two bugs with the Rails Text Input kit when the `add-on` prop is used.
- 🐛 Bug1:  The add_on prop caused the internal validation error message from the `required: true` prop to appear to the right of the input instead of below it. This was happening for both versions of the Rails Text Input kit (pb_rails("text_input") kit and the form.text_field aka form builder)
- 🐞 Bug2: The add_on prop was reported as incompatible with the currency mask - additional investigation revealed it was not compatible with any of the text input mask types. This was happening for both versions of the Rails Text Input kit.

This PR adds a comprehensive yet temporary doc example on the Form kit page with two Form demos. 
I also checked React TextInputsand confirmed masks + addOns do not have any incompatibility.

demo 2
demo 1 one
demo 1 two
demo 1 three
console

1. Go to the temporary doc kits/form/rails#form_form_with_validate_addon_currency.
2. For both Demo 1 and 2 - submit the forms without filling anything out - see the error text appear below each field. 
3. Demo 2 displays how correct behavior now appears for both pb_rails and form builder text inputs with add ons, with and without masks.
4. Demo 1 shows each permutation of text input masks with add on, as pb_rails or form builder. To address any hidden input/form submission issues, fill out the form entirely and submit to see the submitted data in the box below Demo 1 and also logged in the console.

#### Checklist:
~~- [ ] **PLAYGROUND** I have added and tested Playground metadata and overrides for all kits and props updated in my code.~~
- [x] **RC** I have added an `inactive RC` label if not an active RC.


nidaqg (dismissed):
Looks great! Do you think any of these examples should live in the docs long term?

ElisaShapiro:
> Looks great! Do you think any of these examples should live in the docs long term?

Probably not in its current form (don't need every permutation/iteration) but could clean it up some up or integrate into existing docs. There may be value in adding a doc or adding the script to the default doc to show hidden input submission capture in the doc itself + in the console for all the form builder kits. For add on + mask specifically, not really (could throw the two props into a "Text Field 2" entry in default or default + validation docs but that would just be for our eye test testing). Happy to do here or in a minor docs only follow up.

nidaqg:
> > Looks great! Do you think any of these examples should live in the docs long term?
> 
> Probably not in its current form (don't need every permutation/iteration) but could clean it up some up or integrate into existing docs. There may be value in adding a doc or adding the script to the default doc to show hidden input submission capture in the doc itself + in the console for all the form builder kits. For add on + mask specifically, not really (could throw the two props into a "Text Field 2" entry in default or default + validation docs but that would just be for our eye test testing). Happy to do here or in a minor docs only follow up.

the form page needs a much larger overhaul lol.. lets leave docs for later and let this just be the bug fix it is

nidaqg (approved):

PR #6475, opened 2026-08-04
[PLAY-3109] investigation: Disabled State Not Getting to Typeahead in Form Group

nickamantia:
[PLAY-3109](https://runway.powerhrg.com/backlog_items/PLAY-3109)

This PR ensures that the Typeahead disabled state works within a form group.

4. See addition/change

#### Checklist:


playbook/app/pb_kits/playbook/pb_typeahead/index.ts
```diff
@@ -16,6 +16,8 @@ export default class PbTypeahead extends PbEnhancedElement {
   }
 
   connect() {
+    if (this.searchInput?.disabled) return
```
nidaqg: I think this might create a regression for dynamically enabled Typeaheads. If the kit is disabled when connect() runs, we return before attaching any of the listeners. If the input is later enabled, it visually becomes usable but focus/input/search/result/context behavior will never initialize. (this may need to be tested but I'm reasonably sure this may be the case)

Can we avoid skipping connect() entirely? The native disabled input should already prevent user interaction while disabled, while keeping the kit ready if its disabled state changes later.

nickamantia: yeah let me see!

nidaqg (dismissed):

nidaqg (approved):

PR #6538, opened 2026-08-21
[PLAY-3149] Dropdown: Rails Dynamic Options and Event Integration + React Hook Form Support for React

nickamantia:
[PLAY-3149](https://runway.powerhrg.com/backlog_items/PLAY-3149)

This PR also contains all code from [PLAY-3150](https://github.com/powerhome/playbook/pull/6544), cherry picked over for safer testing and merging. 

Screenshot 2026-08-21 at 2 59 35 PM

This PR brings over the dynamic options from the typeahead kit to the dropdown kit.
New props include:
`options_by_context`
`context_selector`
`clear_on_context_change`
`options_event_type`

They are slightly re-named from the typeahead kit just to fit in better with the dropdown.

As stated above, this PR also contains all code from PLAY-3150 which adds react hook form support to the React Dropdown kit.

4. See addition/change

#### Checklist:


playbook/app/pb_kits/playbook/pb_dropdown/docs/_dropdown_dynamic_options.html.erb
```diff
@@ -0,0 +1,83 @@
...
+}) %>
+
+<%= pb_rails("flex", props: { wrap: true, gap: "sm", align: "center", margin_bottom: "md" }) do %>
+  <%= pb_rails("button", props: { id: "update-cities-btn", text: "Update via event", variant: "secondary" }) %>
+  <%= pb_rails("button", props: { id: "select-city-btn", text: "Select Boston", variant: "secondary" }) %>
+  <%= pb_rails("button", props: { id: "clear-city-btn", text: "Clear", variant: "primary" }) %>
+<% end %>
+
+<script>
+  (function() {
```
ElisaShapiro: Currently, the "Update via event" button will change the Pick a City Dropdown to go to the West cities without changing the Select to match (so display shows Northeast or South with Western cities), and the Boston button will only select Boston if Northeast is selected in the Select (does nothing if on south or West). That one is restricted by the Select and the other defies it is a little confusing. Would changing the Update via Event button name to something West-specific add clarity, or using two Dropdowns instead of Select + Dropdown allow for more dynamic changes with button presses? And/or directly reference each example path in the markdown (Boston button demos this, first button demos that).

playbook/app/pb_kits/playbook/pb_dropdown/index.js
```diff
@@ -630,6 +658,214 @@ export default class PbDropdown extends PbEnhancedElement {
...
+    } else {
+      this.reconcileSelectionWithOptions();
+    }
+
+    if (this.target?.classList.contains("open")) {
+      this.adjustDropdownHeight();
+    }
+  }
+
+  reconcileSelectionWithOptions() {
```
nidaqg, line 800: so reconcileSelectionWithOptions looks like it can preserve stale option data when clearSelection is false. We currently reconcile based only on ID, but selectedOptions stores the full serialized option object. If an updated option keeps the same ID but changes its label/value, multi-select pills/hidden inputs continue using the old object. single-select also keeps the old displayed label/autocomplete value. Could we replace retained selections with the corresponding new option payload and refresh the trigger/pills/hidden inputs?

playbook/app/pb_kits/playbook/pb_dropdown/index.js
```diff
@@ -630,6 +658,214 @@ export default class PbDropdown extends PbEnhancedElement {
...
+      optionEls.map((opt) => {
+        try {
+          return JSON.parse(opt.dataset.dropdownOptionLabel).id;
+        } catch {
+          return null;
+        }
+      }),
+    );
+
+    if (this.isMultiSelect) {
```
nidaqg, line 814: I think this is still leaving selected rows visible. Multi-select + clearSelection: false only updates pills/hidden inputs when some selected ids are dropped. If every selected id still exists, this returns early and the newly built option nodes stay visible, selected values then show both as pills and in the list. 

lokks like selected rows are also stored as the old JSON.stringify payload so rebuilt nodes get a new string, so later selectedOptions.has(opt.dataset.dropdownOptionLabel) checks in search/click miss.

playbook/app/pb_kits/playbook/pb_dropdown/index.js
```diff
@@ -630,6 +658,214 @@ export default class PbDropdown extends PbEnhancedElement {
...
+      this.element.closest(`#${CSS.escape(selector)}`) ||
+      document.getElementById(selector)
+    );
+  }
+
+  get clearOnContextChange() {
+    return this.element.dataset.pbDropdownClearOnContextChange !== "false";
+  }
+
+  bindContextSelector() {
```
nidaqg, line 683: This only listens for later change. Until the first change, this keeps the SSR options list. A default or restored context that doesn’t match options shows the wrong list. The docs seem to work around it by duplicating the first context into options.

Perhaps on connect, apply optionsByContext[contextElement.value] the same way Typeahead applies current context up front?

playbook/app/pb_kits/playbook/pb_dropdown/index.js
```diff
@@ -630,6 +658,214 @@ export default class PbDropdown extends PbEnhancedElement {
...
+      normalized.label != null ? String(normalized.label) : "";
+
+    wrapper.appendChild(body);
+    listItem.appendChild(wrapper);
+    optionEl.appendChild(listItem);
+
+    return optionEl;
+  }
+
+  replaceOptions(options, { clearSelection = true } = {}) {
```
nidaqg, line 759: This does not re-run handleSearch or reset keyboardHandler.focusedOptionIndex. So if the autocomplete version still has a typed value, new options will appear unfiltered, and if menu is open the keyboard handler could give undefined on some actions. I'd say when options are changed via context, maybe we just clear typed values in autocomplete, reset keyboard piece etc to avoid issues.

playbook/app/pb_kits/playbook/pb_dropdown/docs/_dropdown_dynamic_options.html.erb
```diff
@@ -0,0 +1,83 @@
...
+
+    document.getElementById("update-cities-btn").addEventListener("click", function() {
+      document.dispatchEvent(new CustomEvent("pb:dropdown:updateOptions", {
+        detail: {
+          dropdownId: dropdownId,
+          options: [
+            { id: "por", label: "Portland", value: "por" },
+            { id: "den", label: "Denver", value: "den" },
+            { id: "aus", label: "Austin", value: "aus" },
+          ],
```
nidaqg: this should not be needed here. The duplicated options: [...] for the first region is not something Typeahead requires. Typeahead applies current context on its own, this list is only needed here because Dropdown doesn’t apply context on connect. Lets fix that connect behavior and drop the duplicate options

playbook/app/pb_kits/playbook/pb_dropdown/docs/_dropdown_dynamic_options.html.erb
```diff
@@ -0,0 +1,83 @@
...
+      }));
+    });
+
+    document.addEventListener("pb:dropdown:selected", function(e) {
+      if (e.target && e.target.id === dropdownId) {
+        console.log("City selected:", e.detail);
+      }
+    });
+  })();
+</script>
```
nidaqg: Lets also slim this script tag to match the typeahead so its clear that nothing extra is needed at time of implementation, otherwise it'll look like all this is required and people will copy/paste it. Extra things like updateOptions / options_event_type should have its own doc if needed

nidaqg: I think we should have separate docs for 'Dynamic Options', 'Dynamic Options with Autocomplete' and 'Dynamic Options with multiselect'

playbook/app/pb_kits/playbook/pb_dropdown/docs/_dropdown_dynamic_options.html.erb
nickamantia: yeah good call, let me try to make this more clear! thanks!

nidaqg (dismissed):

nidaqg (approved):

PR #6561, opened 2026-08-26
[PLAY-3199] Optimize Playbook Production Bundle

nidaqg:
What does this PR do?

This PR introduces Playbook bundle optimizations aimed at reducing shipped JavaScript and removing unnecessary duplicate code.

Changes include:

- Enables Terser compress and mangle during production builds for better minification.
- Adds sideEffects configuration for CSS so bundlers can more effectively tree shake unused JavaScript without removing required styles.
- Updates Advanced Table's VirtualizedTableView to use Playbook's existing shared debounce utility instead of defining its own implementation.

How to test?

- Build Playbook and confirm the production build completes successfully.
- Test Playbook in a consuming application to confirm kits and styles continue to load correctly.
- Regression test Advanced Table virtualization, including column resizing and sorting.
- Compare bundle output/size before and after the changes to confirm the expected optimization.

#### Checklist:

### DIFF
--- a/playbook/app/pb_kits/playbook/pb_advanced_table/Components/VirtualizedTableView.tsx
+++ b/playbook/app/pb_kits/playbook/pb_advanced_table/Components/VirtualizedTableView.tsx
@@ -4,6 +4,7 @@ import { flexRender, Cell, Row } from "@tanstack/react-table"
 import { VirtualItem } from "@tanstack/react-virtual"
 
 import { GenericObject } from "../../types"
+import { debounce } from "../../utilities/object"
 
 import { isChrome } from "../Utilities/BrowserCheck"
 import { getVirtualizedRowStyle } from "../Utilities/TableContainerStyles"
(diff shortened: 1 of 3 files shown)


nidaqg:
Status: Not worth pursuing

What was tried: Phase 1 of a JS/CSS bundle-size optimization pass for playbook-ui. Three changes:
1) enable Terser's mangle/compress passes in vite.config.ts (previously disabled)

2) add a sideEffects field to package.json for better tree-shaking,

3) dedupe a locally-reimplemented debounce in VirtualizedTableView.tsx.

What broke: Enabling Terser's mangle/compress caused runtime errors in downstream consumers (Nitro): TypeError: (0, r.createElement) is not a function inside the built _typeahead chunk, breaking many component renders.

Root cause: full Terser optimization interacts badly with how the externalized react/react-dom import bindings resolve across our code-split chunks. Local vite build and jest runs didn't catch it because our test suite exercises source via ts-jest, not the actual minified dist/chunks/*.js output. Only a real consumer like Nitro loading the built bundle hits it.

Decision: Reverted terserOptions: { mangle: false, compress: false } back to its original state. Kept the two harmless pieces (the debounce dedup and the sideEffects field) since they carry no runtime risk, but neither delivers a measurable size/perf win on its own. The sideEffects field is inert until the package has per-kit/subpath exports to prune (it doesn't yet), and the dedup is a ~10-line cleanup, not a bundle-size change.

Why not worth continuing on this path: Even a clean Terser win would only save ~170KB on the JS vendor chunk (~438KB → ~265KB), so honestly, not worth the lift

PR #6577, opened 2026-08-28
[PLAY-3202] Link kit: Add Configurable rel Prop

nidaqg:

Adds a rel prop to the Link kit (React + Rails) so consumers can override the default rel="noreferrer" that's applied to any target other than 'child'.

That default (noreferrer, which implies noopener) is a good secure-by-default choice, but it has a side effect: it prevents the browser from reusing a previously opened named tab, since a browsing context opened with noopener isn't discoverable by name on later navigations. A <Link target="myNamedTab"> clicked repeatedly would open a brand-new tab every time instead of reusing the one already open , target="child" was already exempted from this, but no other named target had a way to opt out.

rel is optional and additive :  omitting it renders the exact same rel attribute as before, so this is fully backwards compatible. Passing rel="" drops noreferrer/noopener and restores reusable-named-tab behavior, matching what target="child" has always done. This should only be used for targets pointing at trusted, same-app destinations — untrusted/external destinations should keep the default.

4. See addition/change

#### Checklist:


ElisaShapiro (dismissed):
Only one note for playground: does not seem possible to mimic ref="" like the docs (since the encapsulating quotation marks are automatic, you get four when you want two)
Image

nidaqg:
> Only one note for playground: does not seem possible to mimic ref="" like the docs (since the encapsulating quotation marks are automatic, you get four when you want two) Image

fixed!

ElisaShapiro (dismissed):

nickamantia (approved):

PR #6595, opened 2026-09-01
[PLAY-3158] Playbook Website: Form Docs v2

kangaree:

Rehaul the form docs. Add pills to choose which inputs you want to see. Show "submitted data". 

Screenshot 2026-09-02 at 10 19 34 AM

4. See addition/change

#### Checklist:


kangaree:
Oops, I ended up cutting out half the inputs that we had in the original docs! I'm adding those back.

nidaqg:
This looks awesome @kangaree  ! Minor things: 

1. For the code snippets, can those have a horizontal scroll? Especially the ones which are 2 side by side, the entire snippet is not visible and is cut off.
2. For the pills for selection in 'default' and 'validation' sections, the Dropdown multiselect is just labeled as 'MultiSelect' which can be confusing since we also have the MLS kit. Could we also make some of the rest clearer? for instance, url, email, etc are all TextInput under the hood, should that be called out same as how Collection Select says Select as well? Are intl_telephone and phone_number both the phone number input kit? Plus should we include a single select variant of the typeahead as well?

playbook/app/pb_kits/playbook/pb_form/docs/_form_validation.md
```diff
@@ -0,0 +1,3 @@
+Toggle pills to compose a validated form. Required fields and custom messages apply to the selected inputs. Submit an empty form to see error states.
```
ElisaShapiro: Not every form builder kit accepts a custom validation message - maybe outlining which ones do like the [current markdown](https://github.com/powerhome/playbook/blob/master/playbook/app/pb_kits/playbook/pb_form/docs/_form_form_with_validation_msg.md) does would make sense in this markdown (esp. since two different props)?
Since we are also removing the required indicator doc example (which makes sense, as that was pretty redundant) maybe it's [markdown](https://github.com/powerhome/playbook/blob/master/playbook/app/pb_kits/playbook/pb_form/docs/_form_with_required_indicator.md) could live here too (so people know about it, or could be added as default for these examples)?

playbook/app/pb_kits/playbook/pb_form/docs/_form_multi_column.html.erb
```diff
@@ -0,0 +1,58 @@
+<%
+  load Playbook.kit_path("form", "docs", "_sample_data.ruby").to_s
```
ElisaShapiro, line 2: Would a use-case for multi column be a Filter Popover that won't necessarily need the screen size adjustment? Could potentially call out something like, "if creating a multi column set up for an inflexible area like a Filter Popover, you do not need [x level flex] but the gap/wrap props are set to what you would need"? Might be too much feel free to disregard.

ElisaShapiro (commented):
Overall this is incredibly cool and so much more dynamic than the current page which is bulky and clunky!

Agree with Nida's comment about the Pills being a little ambiguous. For kits with multiple form_builder types, maybe we could prefix the Pills with the base kit (but can leave the input names as-is if that is easier :form_field wise?). That could also group like-with-like within the Alphabetization. e.g.
Dropdown: Default
Dropdown: Multi Select
Text Field: Default
Text Field: Email
Text Field: Number
etc...

Also the [loading](https://playbook.powerapp.cloud/kits/form/rails#form_form_with_loading) set up is not represented currently - if we would like to include that could probably be a more basic doc (don't love the script tag but feels helpful to pass along the UX rule of disabled state + spinner).

nidaqg (approved):

ElisaShapiro (approved):

PR #6606, opened 2026-09-02
[PLAY-3203] Route Playgrounds to Staging Behind VPN

nickamantia:
[PLAY-3203](https://runway.powerhrg.com/backlog_items/PLAY-3203)

This PR redirects all playground URLS to Staging. When on staging, non-playground links will redirect back to prod.

4. See addition/change

#### Checklist:


playbook-website/config/routes.rb
```diff
@@ -1,6 +1,29 @@
...
+
+    root to: redirect("https://playbook.powerapp.cloud/"), as: nil
+
+    get "(*path)", to: redirect { |_params, request|
+      qs = request.query_string.present? ? "?#{request.query_string}" : ""
+      "https://playbook.powerapp.cloud#{request.path}#{qs}"
+    }, as: nil
+  end
+
+  # Everywhere else (prod): Playground is served by the SPA so we can show a VPN
```
nidaqg: Could we limit this redirect to the production host only? As written, it applies everywhere except staging, including localhost and review environments. That means we can no longer open or test our local Playground changes at /playground because we’ll always be redirected to the deployed staging site.. seeing that already on the review env.

nickamantia: OOOHH yes for sure thanks

playbook-website/app/javascript/components/Website/src/pages/KitShow/index.tsx
```diff
@@ -105,8 +140,23 @@ const KitShow = () => {
   }, [currentKit, name]);
 
   const handleTabChange = (tab: string) => {
+    if (isLocalHost()) {
+      if (tab === "playground") setDarkMode(false);
+      setActiveTab(tab);
```
nidaqg, line 177: Small  inconsistency: if you move from the staging Playground to Props tab, the production URL will contain '?tab=props' right? but it looks like if you click 'Docs' tab after that it'll update the displayed tab but leave that query parameter in place, so refreshing or sharing the URL opens Props again? Should selecting Docs remove tab from the URL, and selecting Props keep it synchronized? Or am I wrong in what i think this will do?

playbook-website/config/routes.rb
```diff
@@ -1,6 +1,30 @@
...
+    get "(*path)", to: redirect { |_params, request|
+      qs = request.query_string.present? ? "?#{request.query_string}" : ""
+      "https://playbook.powerapp.cloud#{request.path}#{qs}"
+    }, as: nil
+  end
+
+  # Production host only: Playground SPA loads here first so we can show a VPN
+  # message when staging is unreachable; the React app then sends users to staging.
+  # Localhost / review apps serve Playground on the current host (no staging bounce).
+  get "playground", to: "pages#application"
```
nidaqg, line 26: I believe this does not fully protect the Playground data. Production still serves /playground.json, and React Router fetches that payload before we run the staging/VPN redirect. So, in theory, someone can also request the JSON endpoint directly without VPN access.

Since the intent is to protect the Playground data and not only hide the UI, production must not generate or return the Playground JSON payload. We probsbly ned to handle this in the controller or loader. ts .. The current  [PlaygroundLoader always fetches /playground.json](https://github.com/powerhome/playbook/blob/master/playbook-website/app/javascript/components/Website/src/hooks/loaders.ts#L119) before rendering the page, a cleaner implementation would avoid loading Playground data on production entirely

playbook-website/app/javascript/components/Website/src/components/PlaygroundVpnRequiredDialog.tsx
```diff
@@ -0,0 +1,44 @@
...
+    const open = () => setOpened(true)
+    window.addEventListener(PLAYGROUND_VPN_REQUIRED_EVENT, open)
+    return () => window.removeEventListener(PLAYGROUND_VPN_REQUIRED_EVENT, open)
+  }, [])
+
+  const close = () => setOpened(false)
+
+  const retry = () => {
+    close()
+    void goToStaging(`${STAGING_ORIGIN}/playground`)
```
nidaqg: Could the failed destination be retained and used here? The retry always opens the main /playground page. If the user originally selected a kit’s Playground tab, or opened a shared Playground URL containing query parameters, connecting to VPN and clicking “Try again” loses that destination and any shared config

nidaqg (approved):

PR #6545, opened 2026-08-24
[PLAY-3155] POC Form Kit - Auto-bind Model Values and Errors in Fields

kangaree:
**What does this PR do?** A clear and concise description with your runway ticket url.
https://runway.powerhrg.com/backlog_items/PLAY-3155

Improve pb_form_with DX so model-bound fields pre-populate and show validation errors without manual default_value or error props.
- Add AttributeDefaults helpers for model attribute + error lookup
- Wire auto-bind for date_picker, time_picker, dropdown_field, star_rating_field, multi_level_select, phone/intl fields, typeahead, and text/select fields
- Add Form kit docs example: Model Values + Errors
- Add builder_attribute_defaults_spec (12 examples)

**Screenshots:** Screenshots to visualize your addition/change
Screenshot 2026-08-24 at 2 09 29 PM

**How to test?** Steps to confirm the desired behavior:
4. See addition/change

#### Checklist:
- [ ] **LABELS** Add a label: `enhancement`, `bug`, `improvement`, `new kit`, `deprecated`, or `breaking`. See [Changelog & Labels](https://github.com/powerhome/playbook/wiki/Changelog-&-Labels) for details.
- [ ] **DEPLOY** I have added the `milano` label to show I'm ready for a review.
- [ ] **TESTS** I have added test coverage to my code.
- [ ] **PLAYGROUND** I have added and tested Playground metadata and overrides for all kits and props updated in my code.
- [ ] **SEMVER** I have added a `minor`, `major`, or `patch` label for release.
- [ ] **RC** I have added an `inactive RC` label if not an active RC.

PR #6608, opened 2026-09-02
[PLAY-3196] Form Kit - Auto-bind Model Values and Errors in Fields

kangaree:
**What does this PR do?**
https://runway.powerhrg.com/backlog_items/PLAY-3196

Previous POC: https://github.com/powerhome/playbook/pull/6545

Improve pb_form_with DX so model-bound fields pre-populate and show validation errors without manual default_value or error props.

- Add AttributeDefaults helpers for model attribute + error lookup
- Wire auto-bind for date_picker, time_picker, dropdown_field, star_rating_field, multi_level_select, phone/intl fields, typeahead, and text/select fields
- Add Form kit docs example: Model Values + Errors
- Add builder_attribute_defaults_spec (12 examples)

Screenshot 2026-09-02 at 3 47 40 PM

4. See addition/change

#### Checklist:


playbook/lib/playbook/forms/builder/attribute_defaults.rb
```diff
@@ -0,0 +1,88 @@
...
+        def serialize_date_picker_default(value)
+          case value
+          when Time, DateTime, ActiveSupport::TimeWithZone
+            value.utc.iso8601
+          when Date
+            value.iso8601
+          else
+            value.presence&.to_s
+          end
+        end
```
nidaqg: This may cause date picker values to shift to the wring calendar date because they are converted to UTC.. Calling value.utc.iso8601 changes the model’s wall-clock date, so flatpickr might could select the wrong date, did we confirm that this won't be an issue?

playbook/lib/playbook/forms/builder/attribute_defaults.rb
```diff
@@ -0,0 +1,88 @@
...
+            find_dropdown_option(value, options)
+          end
+        end
+
+        def resolve_dropdown_array_default(values, options)
+          return values if values.all? { |entry| entry.is_a?(Hash) }
+
+          matched = values.filter_map { |entry| find_dropdown_option(entry, options) }
+          matched.presence
+        end
```
nidaqg: Unsure on this one but, it does feel like an array-values model attribute will crash a single select Dropdown? Every array is resolved into an array of option hashes, regardless of multi_select. The Dropdown kit’s single-select path then calls transform_keys directly on default_value, which raises on an Array.. This can happen if an array-backed attribute is accidentally or intentionally used without multi_select: true. Maybe we pass multi_select into the resolver and only return an array for multi-select dropdowns? For single-select, either select the first match or leave the default unset

playbook/lib/playbook/forms/builder/attribute_defaults.rb
kangaree: This is valid. I'll fix so there's no conversion to UTC.

playbook/lib/playbook/forms/builder/attribute_defaults.rb
kangaree: Thanks- I went in this direction: pass multi_select for multi-select dropdowns and for single-select, choose first match

kangaree:
I made two updates to harden against regressions: 

- Date picker model defaults now use wall-clock date/time strings so timezone parsing can't shift the selected day
- MLS selected-id matching compares strings and numbers (i.e. `"101"` is equal to `101`).

playbook/lib/playbook/forms/builder/attribute_defaults.rb
```diff
@@ -0,0 +1,98 @@
...
+      # explicitly passed those props.
+      module AttributeDefaults
+      private
+
+        def form_attribute_value(name)
+          return nil unless @object
+          return nil unless @object.respond_to?(name)
+
+          @object.public_send(name)
+        end
```
nidaqg, line 30: form_attribute_value here reads the type-cast attribute using @object.public_send(name). so if there is a failed validation, Rails might have converted invalid submitted input to nil or another value. This could cause fields such as DatePicker to render empty instead of preserving what the user entered so they can correct it. Could this be an issue? Or should we prefer the before-type-cast value when available?

kangaree: Great point! I've made it so that fields like date picker will preserve the user input with `before_type_cast`.

playbook/lib/playbook/forms/builder/date_picker_field.rb
```diff
@@ -20,19 +31,23 @@ def date_picker(name, props: {})
...
+        # does not re-run apply_form_error! and undo the opt-out.
+        input_props = {
+          label: nil,
+          placeholder: props[:placeholder],
+          required: props[:required],
+          validation: props[:validation_message].present? ? { message: props[:validation_message] } : {},
+        }
+        input_props[:error] = props[:error] if props.key?(:error)
+
         input = text_field(
```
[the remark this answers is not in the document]
kangaree: Leaving as-is for now. Flaptickr prefers `defaultDate` over input value when both are set. Not worth it!

playbook/lib/playbook/forms/builder/attribute_defaults.rb
```diff
@@ -0,0 +1,113 @@
...
+        # bare id/string — that raises in Dropdown#input_default_value.
+        def resolve_dropdown_default(value, options, multi_select: false)
+          case value
+          when Hash
+            value
+          when Array
+            resolve_dropdown_array_default(value, options, multi_select: multi_select)
+          else
+            find_dropdown_option(value, options)
+          end
```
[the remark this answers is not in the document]
kangaree: Fixed.

nidaqg (approved):

PR #6613, opened 2026-09-03
Playbook Website: Fixes for Dialog popup for VPN

nidaqg:
**What does this PR do?**

Fixes the "VPN required" Playground dialog, which was unreliable in two ways after [PLAY-3203] shipped:

- **False positive (Firefox-specific, off VPN):** the reachability check used `fetch(..., { mode: "no-cors" })`, which resolves on *any* HTTP response — including a block/error page served for off-VPN traffic — so it couldn't tell a real block from success. That was also browser-dependent (Firefox's DNS handling could reach a different response than Chrome's), so the dialog showed correctly in Chrome but not Firefox, which just hit a raw 404 instead.
- **False negative (all browsers, even on VPN):** switching the check to an `` load probe (onload/onerror instead of fetch) fixed the first bug, but the probe target, `/favicon.ico`, was never added to staging's route allow-list in `routes.rb`. Any unlisted path on staging falls through to its own catch-all, which redirects back to prod — so the probe was silently checking prod's favicon the whole time, always "succeeding" as a redirect, which made the dialog show up even while on VPN with no way to get to staging.

**Fix:**
- `isStagingReachable()` (`siteNavigation.ts`) now probes via an `` element instead of a no-cors `fetch` — a non-image response (e.g. an HTML block page) correctly fails to decode and fires `onerror`, independent of browser-specific network/DNS behavior.
- Added `GET /vpn_check.gif` to staging's host-constrained routes, ahead of the catch-all, serving a tiny inline 1×1 GIF. Being explicitly allow-listed, it can never be swept into the redirect-to-prod fallback, so the probe now genuinely reflects staging's own reachability.

**How to test?**
1. Off VPN, in both Chrome and Firefox: open a kit's Playground tab and confirm the "VPN required" dialog shows consistently in both (no raw 404 in either).
2. On VPN, in any browser: open a kit's Playground tab and confirm it redirects straight to staging with no dialog.
3. `curl -sI https://staging.playbook.powerapp.cloud/vpn_check.gif` while on VPN — expect `200`, `Content-Type: image/gif`, no redirect.

#### Checklist:
- [ ] **LABELS** `bug`
- [ ] **DEPLOY** milano label
- [ ] **TESTS**
- [ ] **PLAYGROUND** n/a (no kit/prop changes)
- [ ] **SEMVER** `patch`
- [ ] **RC** n/a

### DIFF
--- a/playbook-website/app/javascript/components/Website/index.tsx
+++ b/playbook-website/app/javascript/components/Website/index.tsx
@@ -23,6 +23,7 @@ import {
   syncStoredPlatformFromLocation,
   writeStoredPlatform,
 } from "./src/helpers/platform";
+import { navigateSite } from "./src/utils/siteNavigation";
 
 function WebsiteContent() {
   const {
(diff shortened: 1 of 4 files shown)


nidaqg:
not doing this for now, going a different route

